PR #15, opened 2019-11-22
Set recommended prefetch on default smartgateway

csibbitt:


leifmadsen:
I'll make the same corresponding change in the SAO.

leifmadsen:
> I'll make the same corresponding change in the SAO.

Oops, didn't notice which repo I was in :)

PR #16, opened 2019-11-25
Clean up some missing functionality

### LINKED ISSUE #2: Pass operator-sdk scorecard fully
Will need to address somethings so that the scorecard passes fully. There is no reason we don't pass everything since the Operator effectively does nothing :)

```
operator-sdk scorecard
INFO[0000] Using config file: /home/sysop/src/github.com/redhat-service-assurance/service-assurance-operator/.osdk-scorecard.yaml 
basic:
	Spec Block Exists                   : pass
	Labels: 
		"necessity":"required"
		"suite":"basic"

	Status Block Exists                 : pass
	Labels: 
		"necessity":"required"
		"suite":"basic"

	Writing into CRs has an effect      : pass
	Labels: 
		"necessity":"required"
		"suite":"basic"

olm:
	Status fields with descriptors      : fail
	Labels: 
		"necessity":"required"
		"suite":"olm"
	Suggestions: 
		Add a status descriptor for conditions

	Provided APIs have validation       : fail
	Labels: 
		"necessity":"required"
		"suite":"olm"
	Suggestions: 
		Add CRD validation for ServiceAssurance/

	Owned CRDs have resources listed    : pass
	Labels: 
		"suite":"olm"
		"necessity":"required"
	Suggestions: 
		If it would be helpful to an end-user to understand or troubleshoot your CR, consider adding resources [serviceassurances/v1alpha1 replicasets/v1 deployments/v1 configmaps/v1] to the resources section for owned CRD ServiceAssurance

	CRs have at least 1 example         : pass
	Labels: 
		"necessity":"required"
		"suite":"olm"

	Spec fields with descriptors        : pass
	Labels: 
		"necessity":"required"
		"suite":"olm"

```

leifmadsen:
Go through the Operator and clean up some missing functionality that we previously
had in the tech preview deployment. Primary items that have been added are:

* request PVC for Prometheus instance (default 20G)
* create the sslProfile and credentials required for clients to connect
  to the AMQ Interconnect route
* reduce the default router number from 2 to 1
* adjust default metrics/events enable flags to better align to
  how the Ansible Operator prefers parameters to be passed (less error prone)
* update CRD to match the new parameters available

One additional note about the CRD update. Changes in the CSV have also
been updated to reflect the new CRD changes. While doing that, the operator-sdk
scorecard was run and with better knowledge about descriptors, all changes
were made to allow for the scorecard test to pass fully.

Closes #2

### DIFF
--- a/roles/serviceassurance/tasks/component_qdr.yml
+++ b/roles/serviceassurance/tasks/component_qdr.yml
@@ -9,7 +9,7 @@
         namespace: '{{ meta.namespace }}'
       spec:
         deploymentPlan:
-          size: 2
+          size: 1
           role: interior
           livenessPort: 8888
           placement: Any
@@ -45,6 +45,13 @@
           - expose: true
             http: true
             port: 8672
+        sslProfiles:
+          - caCert: saf-demo-interconnect-openstack-ca
+            credentials: saf-demo-interconnect-openstack-credentials
+            generateCaCert: true
+            generateCredentials: true
+            name: openstack
+
   when: interconnect_manifest is not defined
 
 - name: Create QDR instance
--- a/roles/serviceassurance/tasks/component_qdr.yml
+++ b/roles/serviceassurance/tasks/component_qdr.yml
@@ -13,7 +13,6 @@
           role: interior
           livenessPort: 8888
           placement: Any
-          resources: {}
         addresses:
           - distribution: closest
             prefix: closest
@@ -58,7 +57,6 @@
             generateCredentials: true
             mutualAuth: true
             name: inter-router
-        users: {{ meta.name }}-interconnect-users
   when: interconnect_manifest is not defined
 
 - name: Create QDR instance
--- a/roles/serviceassurance/tasks/component_qdr.yml
+++ b/roles/serviceassurance/tasks/component_qdr.yml
@@ -13,6 +13,7 @@
           role: interior
           livenessPort: 8888
           placement: Any
+          resources: {}
         addresses:
           - distribution: closest
             prefix: closest
@@ -51,7 +52,6 @@
             generateCaCert: true
             generateCredentials: true
             name: openstack
-
   when: interconnect_manifest is not defined
 
 - name: Create QDR instance
--- a/roles/serviceassurance/tasks/component_qdr.yml
+++ b/roles/serviceassurance/tasks/component_qdr.yml
@@ -47,11 +47,18 @@
             http: true
             port: 8672
         sslProfiles:
-          - caCert: saf-demo-interconnect-openstack-ca
-            credentials: saf-demo-interconnect-openstack-credentials
+          - caCert: {{ meta.name }}-interconnect-openstack-ca
+            credentials: {{ meta.name }}-interconnect-openstack-credentials
             generateCaCert: true
             generateCredentials: true
             name: openstack
+          - caCert: {{ meta.name }}-interconnect-inter-router-ca
+            credentials: {{ meta.name }}-interconnect-inter-router-credentials
+            generateCaCert: true
+            generateCredentials: true
+            mutualAuth: true
+            name: inter-router
+        users: {{ meta.name }}-interconnect-users
   when: interconnect_manifest is not defined
 
 - name: Create QDR instance
--- a/roles/serviceassurance/tasks/component_qdr.yml
+++ b/roles/serviceassurance/tasks/component_qdr.yml
@@ -32,7 +32,7 @@
           - expose: true
             host: 0.0.0.0
             port: 5671
-            saslMechanisms: EXTERNAL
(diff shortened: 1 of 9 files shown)


leifmadsen:
```
operator-sdk scorecard 
INFO[0000] Using config file: /home/sysop/src/github.com/redhat-service-assurance/service-assurance-operator/.osdk-scorecard.yaml 
basic:
	Spec Block Exists                   : pass
	Labels: 
		"necessity":"required"
		"suite":"basic"

	Status Block Exists                 : pass
	Labels: 
		"necessity":"required"
		"suite":"basic"

	Writing into CRs has an effect      : pass
	Labels: 
		"suite":"basic"
		"necessity":"required"

olm:
	Provided APIs have validation       : pass
	Labels: 
		"necessity":"required"
		"suite":"olm"

	Owned CRDs have resources listed    : pass
	Labels: 
		"suite":"olm"
		"necessity":"required"

	CRs have at least 1 example         : pass
	Labels: 
		"necessity":"required"
		"suite":"olm"

	Spec fields with descriptors        : pass
	Labels: 
		"necessity":"required"
		"suite":"olm"

	Status fields with descriptors      : pass
	Labels: 
		"necessity":"required"
		"suite":"olm"
```

leifmadsen:
[Minor deployment clean up things](https://trello.com/c/ONRTitGp/815-minor-deployment-clean-up-things)

roles/serviceassurance/tasks/main.yml
```diff
@@ -1,8 +1,8 @@
 ---
 - name: Execute default tasks when metrics are requested to be enabled
   include_tasks: sa_metrics.yml
-  when: metrics.enabled
+  when: metrics_enabled
```
leifmadsen: If you're wondering why the mix of snake_case and camelCase:

> Defining the spec for an Ansible-based Operator can be done entirely in Ansible. The Ansible Operator passes all 
> key-value pairs listed in the CR spec field along to Ansible as variables. The names of all variables in the spec 
> field are converted to snake case (lowercase with an underscore) by the Operator before running Ansible. For 
> example, serviceAccount in the spec becomes service_account in Ansible.

[source: https://docs.openshift.com/container-platform/4.1/applications/operator_sdk/osdk-ansible.html#osdk-building-ansible-operator_osdk-ansible] bullet 3

roles/serviceassurance/tasks/component_qdr.yml
```diff
@@ -9,10 +9,11 @@
         namespace: '{{ meta.namespace }}'
       spec:
         deploymentPlan:
-          size: 2
+          size: 1
           role: interior
           livenessPort: 8888
           placement: Any
+          resources: {}
```
csibbitt: If this just to pass the scorecard? If you're going to define a resources block is should really define some requests, at least.

roles/serviceassurance/tasks/component_qdr.yml
```diff
@@ -45,6 +46,19 @@
...
+          - caCert: {{ meta.name }}-interconnect-openstack-ca
+            credentials: {{ meta.name }}-interconnect-openstack-credentials
+            generateCaCert: true
+            generateCredentials: true
+            name: openstack
+          - caCert: {{ meta.name }}-interconnect-inter-router-ca
+            credentials: {{ meta.name }}-interconnect-inter-router-credentials
+            generateCaCert: true
+            generateCredentials: true
+            mutualAuth: true
```
csibbitt: Oooh neat

roles/serviceassurance/tasks/component_qdr.yml
leifmadsen: I was having issues with the Operator looping against the creation of the `component_qdr.yaml` where a change was being applied, so the Operator would never stop updating the k8s resource, leading to the `generation` of the object increasing indefinitely (and the Operator never taking a break from running Ansible).

However, I think there might have been some other issues going on in the test environment that were due to the use of the quickstart script you provided. I haven't investigated yet to figure out why that would cause any sort of issue (maybe some sort of race condition, or just weird out of order stuff).

I'm going to pull this out along with a couple other things that I think should be unnecessary and see if the patching issue returns. If not, then I will pull this out of the PR. If so, then we'll need to leave it in.

(The QDR Operator adds this automatically when undefined. I was just trying to make the spec match fully.)

roles/serviceassurance/tasks/component_qdr.yml
leifmadsen: Yea once I figured that part out, I was like, "cooooool!"

roles/serviceassurance/tasks/component_qdr.yml
leifmadsen: I have resolved this now. The `resource: {}` was unnecessary and has been dropped.

PR #19, opened 2019-11-27
Update statusDescriptors so it is valid again

### LINKED ISSUE #17: Latest CSV incompatible React object
When looking at a ServiceAssurance object in the OpenShift UI it will crash out due to an invalid React component.

I believe this has to do with PR #16 where I added a `statusDescriptor`. I would guess that I've used the wrong x-descriptor for the `condition` status.

leifmadsen:
Unfortunately I might have found a bug or usecase that is not yet documented. Our status.conditions
output is a list, and so far I have not found many examples of the status output being a list.

There does exist an option in the watches.yaml for manageStatus: false that would allow us to
manage the status ourselves. That might ultimately be the solution here, however is a task
for another day.

This change is to give us some information about what happened in the run, and to allow for
the status to show up. Unforunately at this point it will result in the operator-sdk
scorecard from passing fully, but that's more a nice-to-have option at this point.

Closes #17

### DIFF
--- a/deploy/olm-catalog/service-assurance-operator/0.1.0/service-assurance-operator.v0.1.0.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/service-assurance-operator/0.1.0/service-assurance-operator.v0.1.0.clusterserviceversion.yaml
@@ -73,11 +73,36 @@ spec:
         x-descriptors:
         - urn:alm:descriptor:tectonic.ui:text
       statusDescriptors:
-      - description: Status conditions of the ServiceAssurance object creation
-        displayName: Status
-        path: conditions
-        x-descriptors:
-        - urn:alm:descriptor:tectonic.ui:statusConditions
+        - description: Tasks changed
+          displayName: Changed
+          path: conditions[0].ansibleResult.changed
+          x-descriptors:
+            - urn:alm:descriptor:com.tectonic.ui:arrayFieldGroup:ansibleResults
+            - urn:alm:descriptor:com.tectonic.ui:number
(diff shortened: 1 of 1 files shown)


leifmadsen:
Figured this out earlier and it was sitting in my changes for events work, so just want to get this out of my local repo.

leifmadsen:
@csibbitt for reference -- https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md

* https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md#array-field-group
* https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md#text-input
* https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md#number-input

These are just used for the UI output when you click on the `saf-default` instance. Purpose is to take the data output by the `status` field and expose it to the UI so that you can display status of the ServiceAssurance object.

leifmadsen:
All I've done here is basically "group" the values as a single field. Unfortunately the UI is not yet capable of showing grouped objects in the _display view_ so it's actually kind of silly to group them, but if the UI ever catches up, it'll look nicer going forward :)

The ideal would have been to us the `urn:alm:descriptor:io.kubernetes.conditions` descriptor, but because of the way the Ansible Operator controller exposes additional data (and as a list) it doesn't appear as if the React component is compatible.

PR #21, opened 2019-12-11
Fixing events URL

csibbitt:
I noticed this mismatch while trying to get events working in my HA testing. The URL here is wrong.

Here are the some places that suggest it should be "notify" instead:
* https://redhat-service-assurance.github.io/saf-documentation/#configuring-red-hat-openstack-platform-overcloud-for-saf-events_completing-the-saf-configuration
* https://github.com/redhat-service-assurance/telemetry-framework/blob/master/tests/performance-test/deploy/config/minimal-collectd.conf#L16
* https://github.com/redhat-service-assurance/telemetry-framework/blob/master/tests/smoketest/minimal-collectd.conf.template#L21

### DIFF
--- a/roles/serviceassurance/tasks/component_events_smartgateway.yml
+++ b/roles/serviceassurance/tasks/component_events_smartgateway.yml
@@ -8,7 +8,7 @@
         namespace: '{{ meta.namespace }}'
       spec:
         size: 1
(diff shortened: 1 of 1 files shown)


csibbitt:
@mrunge and @elfiesmelfie , if you are still having trouble getting events into ES, this might be one layer of the onion.

pleimer:
Absolutely should be `notify`. Source: all the events work I did in the performance test

PR #23, opened 2019-12-18
Allow deploying ES in HA mod

elfiesmelfie:
WIP

### DIFF
--- a/deploy/crds/infra.watch_serviceassurances_crd.yaml
+++ b/deploy/crds/infra.watch_serviceassurances_crd.yaml
@@ -42,12 +42,9 @@ spec:
             eventsEnabled:
               description: Whether the Service Assurance Operator should enable components related to events collection and storage.
               type: boolean
-            haEnabled:
+            highAvailabilityEnabled:
               description: Whether to deploy the services in HA mode.
               type: boolean
-            esNodeCount:
-              description: Number of elastic search instances to deploy in HA mode (ignored unless haEnabled == True)
-              type: string
             prometheusStorageClass:
               description: Storage class name used for Prometheus PVC
               type: string
--- a/deploy/crds/infra.watch_serviceassurances_crd.yaml
+++ b/deploy/crds/infra.watch_serviceassurances_crd.yaml
@@ -42,6 +42,12 @@ spec:
             eventsEnabled:
               description: Whether the Service Assurance Operator should enable components related to events collection and storage.
               type: boolean
+            haEnabled:
+              description: Whether to deploy the services in HA mode.
+              type: boolean
+            esNodeCount:
+              description: Number of elastic search instances to deploy in HA mode (ignored unless haEnabled == True)
+              type: string
             prometheusStorageClass:
               description: Storage class name used for Prometheus PVC
               type: string
--- a/deploy/olm-catalog/service-assurance-operator/0.1.0/infra.watch_serviceassurances_crd.yaml
+++ b/deploy/olm-catalog/service-assurance-operator/0.1.0/infra.watch_serviceassurances_crd.yaml
@@ -20,62 +20,18 @@ spec:
   validation:
     openAPIV3Schema:
       properties:
-        apiVersion:
-          description: 'APIVersion defines the versioned schema of this representation
-            of an object. Servers should convert recognized schemas to the latest
-            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources'
-          type: string
-        kind:
-          description: 'Kind is a string value representing the REST resource this
-            object represents. Servers may infer this from the endpoint the client
-            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds'
-          type: string
-        metadata:
-          description: Metadata definition for the ServiceAssurance object
-          type: object
         spec:
           description: Specification of the desired behavior of the Service Assurance Operator.
           properties:
-            metricsEnabled:
-              description: Whether the Service Assurance Operator should enable components related to metrics collection and storage.
-              type: boolean
-            eventsEnabled:
-              description: Whether the Service Assurance Operator should enable components related to events collection and storage.
-              type: boolean
-            haEnabled:
-              description: Whether to deploy the services in HA mode.
-              type: boolean
-            esNodeCount:
-              description: Number of elastic search instances to deploy in HA mode (ignored unless haEnabled == True)
-              type: string
-            prometheusStorageClass:
-              description: Storage class name used for Prometheus PVC
-              type: string
-            prometheusStorageResources:
-              description: Storage resource definition for Prometheus
-              type: string
-            prometheusStorageSelector:
-              description: Storage selector definition for Prometheus
-              type: string
-            prometheusPvcStorageRequest:
-              description: PVC storage requested size for Prometheus
-        status:
-          description: Status results of an instance of Service Assurance
-          properties:
-            conditions:
-              description: The resulting conditions when a Service Assurance is instantiated
-              items:
-                properties:
-                  status:
-                    type: string
-                  type:
-                    type: string
-                  reason:
-                    type: string
-                  lastTransitionTime:
-                    type: string
-                type: object
-              type: array
-          type: object
-
-# vim: set tabstop=2 shiftwidth=2 expandtab ft=yaml:
+            metrics:
+              description: Options related to metrics collection and storage.
+              properties:
+                enabled:
+                  description: Whether the Service Assurance Operator should enable components related to metrics collection and storage.
+                  type: boolean
+            events:
+              description: Options related to events collection and storage.
+              properties:
+                enabled:
+                  description: Whether the Service Assurance Operator should enable components related to events collection and storage.
+                  type: boolean
--- a/deploy/olm-catalog/service-assurance-operator/0.1.0/infra.watch_serviceassurances_crd.yaml
+++ b/deploy/olm-catalog/service-assurance-operator/0.1.0/infra.watch_serviceassurances_crd.yaml
@@ -42,6 +42,12 @@ spec:
             eventsEnabled:
               description: Whether the Service Assurance Operator should enable components related to events collection and storage.
               type: boolean
+            haEnabled:
+              description: Whether to deploy the services in HA mode.
+              type: boolean
+            esNodeCount:
+              description: Number of elastic search instances to deploy in HA mode (ignored unless haEnabled == True)
+              type: string
             prometheusStorageClass:
               description: Storage class name used for Prometheus PVC
               type: string
--- a/deploy/olm-catalog/service-assurance-operator/0.1.0/infra.watch_serviceassurances_crd.yaml
+++ b/deploy/olm-catalog/service-assurance-operator/0.1.0/infra.watch_serviceassurances_crd.yaml
@@ -20,18 +20,56 @@ spec:
   validation:
     openAPIV3Schema:
       properties:
+        apiVersion:
+          description: 'APIVersion defines the versioned schema of this representation
+            of an object. Servers should convert recognized schemas to the latest
+            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources'
+          type: string
+        kind:
+          description: 'Kind is a string value representing the REST resource this
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          description: Metadata definition for the ServiceAssurance object
+          type: object
         spec:
           description: Specification of the desired behavior of the Service Assurance Operator.
           properties:
-            metrics:
-              description: Options related to metrics collection and storage.
-              properties:
-                enabled:
-                  description: Whether the Service Assurance Operator should enable components related to metrics collection and storage.
-                  type: boolean
-            events:
-              description: Options related to events collection and storage.
-              properties:
-                enabled:
-                  description: Whether the Service Assurance Operator should enable components related to events collection and storage.
-                  type: boolean
+            metricsEnabled:
+              description: Whether the Service Assurance Operator should enable components related to metrics collection and storage.
+              type: boolean
+            eventsEnabled:
+              description: Whether the Service Assurance Operator should enable components related to events collection and storage.
+              type: boolean
+            prometheusStorageClass:
+              description: Storage class name used for Prometheus PVC
+              type: string
+            prometheusStorageResources:
+              description: Storage resource definition for Prometheus
+              type: string
+            prometheusStorageSelector:
+              description: Storage selector definition for Prometheus
+              type: string
+            prometheusPvcStorageRequest:
+              description: PVC storage requested size for Prometheus
+        status:
+          description: Status results of an instance of Service Assurance
+          properties:
+            conditions:
+              description: The resulting conditions when a Service Assurance is instantiated
+              items:
+                properties:
+                  status:
+                    type: string
+                  type:
+                    type: string
+                  reason:
+                    type: string
+                  lastTransitionTime:
+                    type: string
+                type: object
+              type: array
+          type: object
+
+# vim: set tabstop=2 shiftwidth=2 expandtab ft=yaml:
--- a/deploy/olm-catalog/service-assurance-operator/0.1.0/service-assurance-operator.v0.1.0.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/service-assurance-operator/0.1.0/service-assurance-operator.v0.1.0.clusterserviceversion.yaml
@@ -8,12 +8,15 @@ metadata:
           "apiVersion": "infra.watch/v1alpha1",
           "kind": "ServiceAssurance",
           "metadata": {
-            "name": "saf-default"
+            "name": "cloud1"
           },
           "spec": {
-            "eventsEnabled": false,
-            "metricsEnabled": true,
-            "haEnabled": false,
+            "events": {
+              "enabled": false
+            },
+            "metrics": {
+              "enabled": true
+            }
           }
         }
       ]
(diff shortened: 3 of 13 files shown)


deploy/crds/infra.watch_serviceassurances_crd.yaml
```diff
@@ -42,6 +42,12 @@ spec:
             eventsEnabled:
               description: Whether the Service Assurance Operator should enable components related to events collection and storage.
               type: boolean
+            haEnabled:
+              description: Whether to deploy the services in HA mode.
+              type: boolean
+            esNodeCount:
+              description: Number of elastic search instances to deploy in HA mode (ignored unless haEnabled == True)
+              type: string
```
csibbitt: Isn't this an integer?

deploy/crds/infra.watch_serviceassurances_crd.yaml
```diff
@@ -42,6 +42,12 @@ spec:
             eventsEnabled:
               description: Whether the Service Assurance Operator should enable components related to events collection and storage.
               type: boolean
+            haEnabled:
```
csibbitt: Is this file still in use now that we use OLM? I think maybe it should be deleted from the tree, but not 100% certain

deploy/olm-catalog/service-assurance-operator/0.1.0/infra.watch_serviceassurances_crd.yaml
```diff
@@ -42,6 +42,12 @@ spec:
             eventsEnabled:
               description: Whether the Service Assurance Operator should enable components related to events collection and storage.
               type: boolean
+            haEnabled:
+              description: Whether to deploy the services in HA mode.
+              type: boolean
+            esNodeCount:
+              description: Number of elastic search instances to deploy in HA mode (ignored unless haEnabled == True)
+              type: string
```
csibbitt: integer?

roles/serviceassurance/tasks/component_elasticsearch.yml
```diff
@@ -15,9 +15,9 @@
               memory: 4Gi
             requests:
               memory: 1Gi
-        redundancyPolicy: ZeroRedundancy
+        redundancyPolicy: {{ "FullRedundancy" if haEnabled else "ZeroRedundancy" }}
```
csibbitt: Given the current state, I think the haEnabled is not necessary; simply testing esNodeCount > 1 would have the same effect. Perhaps this is meant for future use? I'm trying to consider whether any of my work would benefit from such a flag, or whether similarly setting a "size" or "nodecount" property to >1 is all I really need. We can feel this out as we make progress.

elfiesmelfie: I was considering whether this was needed as well. Potentially, we can provide reasonable defaults for HA (esNodeCount, etc), and have them set all the time, but ignored unless haEnabled==True. This could be useful.

deploy/crds/infra.watch_serviceassurances_crd.yaml
elfiesmelfie: There is some duplication, but I don't know what is required or not. Perhaps we should get @leifmadsen to weight in on this

deploy/olm-catalog/service-assurance-operator/0.1.0/infra.watch_serviceassurances_crd.yaml
elfiesmelfie: This was my thought also, however, I think it has to correspond with deploy/olm-catalog/service-assurance-operator/0.1.0/service-assurance-operator.v0.1.0.clusterserviceversion.yaml, and I'm not sure where those types come from, when I find that out, this should be updated to and integer field

deploy/crds/infra.watch_serviceassurances_crd.yaml
leifmadsen: This file is 100% required. When you build the `olm-catalog` with the `operator-courier` tool, it uses the files in `deploy/crds/` and `deploy/cr` to construct the CSV.

deploy/crds/infra.watch_serviceassurances_crd.yaml
```diff
@@ -42,6 +42,12 @@ spec:
...
+              description: Whether to deploy the services in HA mode.
+              type: boolean
+            esNodeCount:
```
leifmadsen: I would call this out at `elasticsearchNodeCount`

deploy/crds/infra.watch_serviceassurances_crd.yaml
leifmadsen: I would call this out to `highAvailabilityEnabled` or similar.

deploy/olm-catalog/service-assurance-operator/0.1.0/infra.watch_serviceassurances_crd.yaml
leifmadsen: As noted earlier, this is a compiled file from the contents of `deploy/`. You would compile a new CSV with `operator-sdk olm-catalog gen-csv --update-crds --csv-version 0.1.0`

deploy/olm-catalog/service-assurance-operator/0.1.0/service-assurance-operator.v0.1.0.clusterserviceversion.yaml
```diff
@@ -12,7 +12,8 @@ metadata:
           },
           "spec": {
             "eventsEnabled": false,
-            "metricsEnabled": true
+            "metricsEnabled": true,
```
leifmadsen: Don't modify this manually. You use the `operator-sdk` to build this out using `gen-csv`

roles/serviceassurance/tasks/component_elasticsearch.yml
leifmadsen: You actually want to check on this not just via a single flag either. You need > 2 nodes for FullRedundancy I'm pretty sure. If you have `elasticsearchNodeCount: 2` then you want this to be `SingleRedundancy` I think.

deploy/olm-catalog/service-assurance-operator/0.1.0/infra.watch_serviceassurances_crd.yaml
csibbitt: Should we gitignore the compiled files, then?

roles/serviceassurance/tasks/component_elasticsearch.yml
csibbitt: Ah yes, if we need 3 nodes that's important to know. I'm not sure if there has been an testing on failed ES states yet or not.

deploy/olm-catalog/service-assurance-operator/0.1.0/infra.watch_serviceassurances_crd.yaml
leifmadsen: No because they are only partially compiled :)

leifmadsen: So basically when you run the `gen-csv` it'll update the `alm-example` section with the values in the `cr` file within `deploy/crds/`. It'll also copy over the `_crd` file from the `deploy/crds/` directory into the `olm-catalog/service-assurance-operator/0.1.0` directory.

The rest of the CSV is basically then maintained by hand. If this were a Go Operator, then more things could be handled in the code and the `gen-csv` command would have more hints to generating more of the CSV file itself.

When you're doing testing, the `deploy/crds` directory is used, like when we run the `operator-sdk scorecard`. It's expected that the `deploy/crds` directory exists with your CRD and CR files for handling a deployment. The `olm-catalog` directory is effectively a package manifest for the deployment of the Operator.

Some things are duplicated, but those that are only exist in a single location, and the duplicated data is generated by `operator-sdk`. Another example is if you change the `deploy/role.yaml` to modify the roles available for the Operator, that will be updated in the `CSV` file when you run `gen-csv` as well.

csibbitt: Thanks!

roles/serviceassurance/tasks/component_elasticsearch.yml
elfiesmelfie: These docs [1] say nothing about limitations for nodeCount vs FullRedundancy. For nodeCount==2, SingleRedundancy is equivalent to SingleRedundancy AFAICT. 
The only limitation I saw noted was SingleRedundancy doesn't work on single node deployment.

I tested nodeCount=2 with FullRedundancy, and the smoketests ran. 

[1] https://docs.openshift.com/container-platform/4.1/logging/config/efk-logging-elasticsearch.html

deploy/crds/infra.watch_serviceassurances_crd.yaml
elfiesmelfie: Makes sense

elfiesmelfie: @csibbitt @leifmadsen Rather than letting the use specify a number of nodes for each component in HA mode, would it make more sense to actually toggle the HA mode with a pre-defined set of configs for HA and noHA? Or do we want to user to be able to turn these knobs?

leifmadsen: I'm kind of leaning towards the opinion that we should limit the number of knobs made available. I like the idea of `highAvailabilityEnabled` with a value of just `true` or `false` and then strictly defining what that means. It will limit our test surface as well.

We already have the ability for manifest overrides in the Operator, so I'm totally fine hiding a lot of these tunables and just giving `yes / no` setup options and defining strictly what was expect "HA" to mean.

deploy/olm-catalog/service-assurance-operator/0.1.0/service-assurance-operator.v0.1.0.clusterserviceversion.yaml
leifmadsen: Also note that you want to use operator-sdk 0.12.0 or 0.13.1 if it is available. The operator-sdk 0.13.0 has a bug in it which destroys the CSV values under the `customresourcedefinition.owned` section when using a non-Go Operator (like we're doing here).

When you're ready for the `olm-catalog` directory to be updated let me know, and I'm happy to generate the updated package manifest as a commit here. I can also run the scorecard and all that as well. Happy to give you a demo when you are ready.

deploy/crds/infra.watch_serviceassurances_crd.yaml
csibbitt: I think for now I tend to agree with leif that there should be exactly 1 knob; actually a switch. HA on or off. We will set the correct number and config of everything from there. As customer needs expand we might need to add more, but that seems like the best place to start.

roles/serviceassurance/tasks/component_elasticsearch.yml
csibbitt: Did you test what happens to the stream of events when you kill a node?

deploy/crds/infra.watch_serviceassurances_crd.yaml
leifmadsen: The only other consideration I could think of would be to allow it per deployment type, e.g.

* `metricsHighAvailabilityEnabled: {true|false}`
* `eventsHighAvailabilityEnabled: {true|false}`

Otherwise, our unstated goal has been to make the configuration as simple as possible, and to provide a biased configuration using information from testing. Providing a well defined set of simple configuration options with the ability to override that was the approach Chris originally recommended when discussing SAO, and I am a believer.

elfiesmelfie: I'd lean towards all or nothing in terms of HA i.e. no extra knobs per deployment type. If that is required later, we can add it later, but for now, metricsEnabled, eventsEnabled and highAvailabilityEnabled would seem sufficient.

Does the esNodeCount value need to remain in this file? Or should I use some hardcoded integers in the ansible files?
Indeed, what does this variable do here, and where is it set/manipulated?

roles/serviceassurance/tasks/component_elasticsearch.yml
elfiesmelfie: No, no I did not. I just see the smoke tests passing and the pods recovering.

deploy/olm-catalog/service-assurance-operator/0.1.0/service-assurance-operator.v0.1.0.clusterserviceversion.yaml
elfiesmelfie: I was seeing some weird behaviour with operator-sdk, wondering why the contents of that file were obliterated; it was version 0.13.0! I'll re-install a different version and prod it. 

Would appreciate a demo tomorrow!

deploy/crds/infra.watch_serviceassurances_crd.yaml
leifmadsen: I don't think it needs to remain in the file. I'd just code it into the `defaults/main.yml` as a variable within the role, and then reference that from the default template at the top of the elasticsearch playbook.

deploy/olm-catalog/service-assurance-operator/0.1.0/service-assurance-operator.v0.1.0.clusterserviceversion.yaml
leifmadsen: No problem, my calendar is pretty open, so just ping me when you're around on chat and I can show you what I remember from before the break :) Remind me to record it and I'll ship it off to @csibbitt who is off tomorrow.

deploy/olm-catalog/service-assurance-operator/0.1.0/infra.watch_serviceassurances_crd.yaml
```diff
@@ -20,62 +20,18 @@ spec:
   validation:
     openAPIV3Schema:
       properties:
-        apiVersion:
```
leifmadsen: This is stripped probably because you generated with olm-catalog 0.13.0. I'd suggest going back a version so that this isn't stripped out.

deploy/olm-catalog/service-assurance-operator/0.1.0/infra.watch_serviceassurances_crd.yaml
```diff
@@ -20,62 +20,18 @@ spec:
...
-                  reason:
-                    type: string
-                  lastTransitionTime:
-                    type: string
-                type: object
-              type: array
-          type: object
-
-# vim: set tabstop=2 shiftwidth=2 expandtab ft=yaml:
+            metrics:
```
leifmadsen: This should all show up from the `operator-sdk olm-catalog gen-csv` command I'm pretty sure.

deploy/olm-catalog/service-assurance-operator/0.1.0/service-assurance-operator.v0.1.0.clusterserviceversion.yaml
```diff
@@ -8,12 +8,15 @@ metadata:
           "apiVersion": "infra.watch/v1alpha1",
           "kind": "ServiceAssurance",
           "metadata": {
-            "name": "saf-default"
+            "name": "cloud1"
```
leifmadsen: We've been using `saf-default` not `cloud1` as the default name for an SAF instance. I think @mrunge might be depending on that name for the smoke tests as well.

deploy/olm-catalog/service-assurance-operator/0.1.0/service-assurance-operator.v0.1.0.clusterserviceversion.yaml
```diff
@@ -8,12 +8,15 @@ metadata:
...
           },
           "spec": {
-            "eventsEnabled": false,
-            "metricsEnabled": true,
-            "haEnabled": false,
+            "events": {
```
leifmadsen: Again this comes from the `deploy/crd` directory and should be automatically generated with the correct version of `operator-sdk` >= 13.1 and < 13.0

deploy/olm-catalog/service-assurance-operator/0.1.0/infra.watch_serviceassurances_crd.yaml
```diff
@@ -20,62 +20,18 @@ spec:
...
+              description: Options related to metrics collection and storage.
+              properties:
+                enabled:
+                  description: Whether the Service Assurance Operator should enable components related to metrics collection and storage.
+                  type: boolean
+            events:
```
leifmadsen: Why is this showing up as a nested value? Everything needs to be flat with the Ansible operator. The configuration you're referencing in the CR is `eventsEnabled` and `metricsEnabled`, so that's what needs to exist here. I'm going to call it out further down as well.

deploy/olm-catalog/service-assurance-operator/0.1.0/service-assurance-operator.v0.1.0.clusterserviceversion.yaml
```diff
@@ -8,12 +8,15 @@ metadata:
...
+              "enabled": false
```
leifmadsen: You can't make this a map in Ansible Operators (well technically you can, but let me explain why it's a bad idea).

When you used mapped values with Ansible Operators instead of keeping everything flat (single plane), then if you ever add anything to the `metrics` or `events` mapping, then you need to re-define the entire map if you want to change just one value. The entire map is overridden, so let let's say you have `metrics.enabled` and `metrics.highAvail` and you just want to override one of those values, then you _must_ define both. If you don't then it'll just be a non-existent parameter when the Ansible runs.

The easiest way to avoid this confusion is to define everything flat like we've been doing.

deploy/olm-catalog/service-assurance-operator/0.1.0/service-assurance-operator.v0.1.0.clusterserviceversion.yaml
```diff
@@ -57,63 +60,26 @@ spec:
         name: ""
         version: v1
       specDescriptors:
+      - description: Metrics related configuration options
+        displayName: Metrics Configuration Options
+        path: metrics
+        x-descriptors:
+        - 'urn:alm:descriptor:com.tectonic.ui:fieldGroup:'
```
leifmadsen: Yea I did all of this before and ripped it out :) As explained above, we need to keep all parameters flat and not a mapping.

roles/serviceassurance/defaults/main.yml
```diff
@@ -2,11 +2,15 @@
 # set default enablement
 metrics_enabled: true
 events_enabled: false
+highAvailabilityEnabled: false
```
leifmadsen: So with Ansible Operators, a camelCase will be automatically converted into snake_case. That's why everything is defined here in the defaults as the snake_case in order to just have a backup conversion in case the camelCase is defined incorrectly in the CR.

Change this to `high_availability_enabled: false`

roles/serviceassurance/defaults/main.yml
```diff
@@ -2,11 +2,15 @@
...
 
 # default values
 prometheus_storage_class: ""
 prometheus_storage_resources: {}
 prometheus_storage_selector: {}
 prometheus_pvc_storage_request: 20G
 
+# defaults for when high availbilty is enabled
+elasticSearchNodeCount: 3
```
leifmadsen: `elasticsearch_node_count` -- we would expect the inbound camelCase to be `elasticsearchNodeCount`

roles/serviceassurance/tasks/component_elasticsearch.yml
```diff
@@ -15,9 +15,9 @@
               memory: 4Gi
             requests:
               memory: 1Gi
-        redundancyPolicy: {{ "FullRedundancy" if haEnabled else "ZeroRedundancy" }}
+        redundancyPolicy: {{ "FullRedundancy" if highAvailabilityEnabled else "ZeroRedundancy" }}
         nodes:
-          - nodeCount: {{ esNodeCount if haEnabled else "1" }}
+          - nodeCount: {{ elasticSearchNodeCount if highAvailabilityEnabled else "1" }}
```
leifmadsen: `elasticsearch_node_count` and `high_availability_enabled`

roles/serviceassurance/tasks/component_elasticsearch.yml
```diff
@@ -15,9 +15,9 @@
               memory: 4Gi
             requests:
               memory: 1Gi
-        redundancyPolicy: {{ "FullRedundancy" if haEnabled else "ZeroRedundancy" }}
+        redundancyPolicy: {{ "FullRedundancy" if highAvailabilityEnabled else "ZeroRedundancy" }}
```
leifmadsen: `high_availability_enabled`

deploy/olm-catalog/service-assurance-operator/0.1.0/service-assurance-operator.v0.1.0.clusterserviceversion.yaml
csibbitt: Thanks for injecting this thinking early. Asking the user to keep the entire config in sync with whatever is in the default map is truly unfair and causes lots of trouble. I've suffered through this in Chef, Puppet, and Helm, and the solution (staying flat) is SO SIMPLE, you just need to stick to it.

roles/serviceassurance/tasks/component_elasticsearch.yml
csibbitt: Instead of peppering all of our files with ` if high_availability_enabled else <default> ` is it possible to move this to one central place? I haven't looked into how to do this in our ansible yet, but pseudo-code I'm basically looking for this:
```
# Defaults for noHA
elasticsearch_node_count = 1
qdr_node_count = 1

# Defaults for HA
if high_availability_enabled then {
 elasticsearch_node_count = 3
 qdr_node_count = 2
}
```

leifmadsen: Sounds like you want to have a `set_fact` in the `roles/serviceassurance/tasks/main.yml` that sets the variables conditionally. Basically same thing we're doing here in the manifest configuration, but just have that manifest reference the variables, and move their dynamic parameter sets in a single location.

csibbitt: Something like this, then: https://github.com/redhat-service-assurance/service-assurance-operator/pull/26

PR #28, opened 2020-01-15
Instantiate alertmanager

mrunge:
The issue with this is currently:

MountVolume.SetUp failed for volume "config-volume" : secret "alertmanager-saf-default" not found

I wonder where prometheus gets the config generated from....

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ ServiceAssurance) via the YAML configuration to override the default manifests
 loaded for you.
 
 * prometheus_manifest
+* alertmanager_manifest
 * elasticsearch_secret_manifest
 * interconnect_manifest
 * elasticsearch_manifest
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ ServiceAssurance) via the YAML configuration to override the default manifests
 loaded for you.
 
 * prometheus_manifest
+* alertmanager_config_manifest
 * alertmanager_manifest
 * elasticsearch_secret_manifest
 * interconnect_manifest
@@ -37,12 +38,3 @@ loaded for you.
 * smartgateway_metrics_manifest
 * smartgateway_events_manifest
 * servicemonitor_manifest
-
-
-## Local testing ##
-
-Dependencies:
-
-* python3-ansible-runner
-* python3-kubernetes
-* python3-openshift
--- a/README.md
+++ b/README.md
@@ -37,3 +37,12 @@ loaded for you.
 * smartgateway_metrics_manifest
 * smartgateway_events_manifest
 * servicemonitor_manifest
+
+
+## Local testing ##
+
+Dependencies:
+
+* python3-ansible-runner
+* python3-kubernetes
+* python3-openshift
--- a/roles/serviceassurance/tasks/component_alertmanager.yml
+++ b/roles/serviceassurance/tasks/component_alertmanager.yml
@@ -10,8 +10,6 @@
         namespace: '{{ meta.namespace }}'
       spec:
         replicas: 1
-        ruleSelector: {}
-        securityContext: {}
         serviceAccountName: prometheus-k8s
         serviceMonitorSelector:
           matchLabels:
--- a/roles/serviceassurance/tasks/component_alertmanager.yml
+++ b/roles/serviceassurance/tasks/component_alertmanager.yml
@@ -48,9 +48,13 @@
           - name: 'null'
   when: alertmanager_config_manifest is not defined
 
-- name: Create an instance of Alertmanager
+- name: Create an Alertmanager configuration secret
   k8s:
     definition:
       '{{ alertmanager_config_manifest }}'
 
+- name: Create an instance of Alertmanager
+  k8s:
+    definition:
+      '{{ alertmanager_manifest }}'
 # TODO: Create route to access Alertmanager (conditional, default off)
--- a/roles/serviceassurance/tasks/component_alertmanager.yml
+++ b/roles/serviceassurance/tasks/component_alertmanager.yml
@@ -12,7 +12,7 @@
         replicas: 1
         ruleSelector: {}
         securityContext: {}
-        serviceAccountName: alertmanager-k8s
+        serviceAccountName: prometheus-k8s
         serviceMonitorSelector:
           matchLabels:
             app: smart-gateway
--- a/roles/serviceassurance/tasks/component_alertmanager.yml
+++ b/roles/serviceassurance/tasks/component_alertmanager.yml
@@ -44,10 +44,6 @@
             group_interval: 5m
             repeat_interval: 12h
             receiver: 'null'
-            routes:
-            - match:
-                alertname: DeadMansSwitch
-              receiver: 'null'
           receivers:
           - name: 'null'
   when: alertmanager_config_manifest is not defined
--- a/roles/serviceassurance/tasks/component_alertmanager.yml
+++ b/roles/serviceassurance/tasks/component_alertmanager.yml
@@ -1,0 +1,35 @@
+- name: Set default Alertmanager manifest
+  set_fact:
+    alertmanager_manifest: |
+      apiVersion: monitoring.coreos.com/v1
+      kind: Alertmanager
+      metadata:
+        labels:
+          alertmanager: '{{ meta.name }}'
+        name: '{{ meta.name }}'
+        namespace: '{{ meta.namespace }}'
+      spec:
+        replicas: 1
+        ruleSelector: {}
+        securityContext: {}
+        serviceAccountName: alertmanager-k8s
+        serviceMonitorSelector:
+          matchLabels:
+            app: smart-gateway
+        storage:
+          class: {{ alertmanager_storage_class }}
+          resources: {{ alertmanager_storage_resources }}
+          selector: {{ alertmanager_storage_selector }}
+          volumeClaimTemplate:
+            spec:
+              resources:
+                requests:
+                  storage: {{ alertmanager_pvc_storage_request }}
+  when: alertmanager_manifest is not defined
+
+- name: Create an instance of Alertmanager
+  k8s:
+    definition:
+      '{{ alertmanager_manifest }}'
+
+# TODO: Create route to access Alertmanager (conditional, default off)
--- a/roles/serviceassurance/tasks/component_alertmanager.yml
+++ b/roles/serviceassurance/tasks/component_alertmanager.yml
@@ -27,9 +27,36 @@
                   storage: {{ alertmanager_pvc_storage_request }}
   when: alertmanager_manifest is not defined
 
+- name: Set default Alertmanager configuration manifest
+  set_fact:
+    alertmanager_config_manifest: |
+      apiVersion: v1
+      kind: Secret
+      metadata:
+        name: alertmanager-'{{ meta.name }}'
+        namespace: '{{ meta.namespace }}'
+      type: Opaque
+      stringData:
+        alertmanager.yaml: |-
+          global:
+            resolve_timeout: 5m
+          route:
+            group_by: ['job']
+            group_wait: 30s
+            group_interval: 5m
+            repeat_interval: 12h
+            receiver: 'null'
+            routes:
+            - match:
+                alertname: DeadMansSwitch
+              receiver: 'null'
+          receivers:
+          - name: 'null'
+  when: alertmanager_config_manifest is not defined
+
 - name: Create an instance of Alertmanager
   k8s:
     definition:
-      '{{ alertmanager_manifest }}'
+      '{{ alertmanager_config_manifest }}'
 
 # TODO: Create route to access Alertmanager (conditional, default off)
--- a/deploy/crds/infra.watch_serviceassurances_crd.yaml
+++ b/deploy/crds/infra.watch_serviceassurances_crd.yaml
@@ -53,6 +53,17 @@ spec:
               type: string
             prometheusPvcStorageRequest:
               description: PVC storage requested size for Prometheus
+            alertmanagerStorageClass:
+              description: Storage class name used for Alertmanager PVC
+              type: string
+            alertmanagerStorageResources:
+              description: Storage resource definition for Alertmanager
+              type: string
+            alertmanagerStorageSelector:
+              description: Storage selector definition for Alertmanager
+              type: string
+            alertmanagerPvcStorageRequest:
+              description: PVC storage requested size for Alertmanager
         status:
           description: Status results of an instance of Service Assurance
           properties:
--- a/deploy/olm-catalog/service-assurance-operator/0.1.0/infra.watch_serviceassurances_crd.yaml
+++ b/deploy/olm-catalog/service-assurance-operator/0.1.0/infra.watch_serviceassurances_crd.yaml
@@ -42,6 +42,9 @@ spec:
             eventsEnabled:
               description: Whether the Service Assurance Operator should enable components related to events collection and storage.
               type: boolean
+            highAvailabilityEnabled:
+              description: Whether to deploy the services in HA mode.
+              type: boolean
             prometheusStorageClass:
               description: Storage class name used for Prometheus PVC
               type: string
--- a/deploy/olm-catalog/service-assurance-operator/0.1.0/infra.watch_serviceassurances_crd.yaml
+++ b/deploy/olm-catalog/service-assurance-operator/0.1.0/infra.watch_serviceassurances_crd.yaml
@@ -53,6 +53,17 @@ spec:
               type: string
             prometheusPvcStorageRequest:
               description: PVC storage requested size for Prometheus
+            alertmanagerStorageClass:
+              description: Storage class name used for Alertmanager PVC
+              type: string
+            alertmanagerStorageResources:
+              description: Storage resource definition for Alertmanager
+              type: string
+            alertmanagerStorageSelector:
+              description: Storage selector definition for Alertmanager
+              type: string
+            alertmanagerPvcStorageRequest:
+              description: PVC storage requested size for Alertmanager
         status:
           description: Status results of an instance of Service Assurance
           properties:
--- a/deploy/olm-catalog/service-assurance-operator/0.1.0/service-assurance-operator.v0.1.0.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/service-assurance-operator/0.1.0/service-assurance-operator.v0.1.0.clusterserviceversion.yaml
@@ -125,6 +125,11 @@ spec:
       kind: Prometheus
       name: prometheuses.monitoring.coreos.com
       version: v1
+    - description: Creation of Alertmanager instances
+      displayName: Alertmanager
+      kind: Alertmanager
+      name: alertmanager.monitoring.coreos.com
+      version: v1
     - description: Creation of Smart Gateways
       displayName: Smart Gateway
       kind: SmartGateway
--- a/deploy/olm-catalog/service-assurance-operator/0.1.0/service-assurance-operator.v0.1.0.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/service-assurance-operator/0.1.0/service-assurance-operator.v0.1.0.clusterserviceversion.yaml
@@ -12,6 +12,7 @@ metadata:
           },
           "spec": {
             "eventsEnabled": false,
+            "highAvailabilityEnabled": false,
             "metricsEnabled": true
           }
         }
--- a/deploy/olm-catalog/service-assurance-operator/0.1.0/service-assurance-operator.v0.1.0.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/service-assurance-operator/0.1.0/service-assurance-operator.v0.1.0.clusterserviceversion.yaml
@@ -125,11 +125,6 @@ spec:
       kind: Prometheus
       name: prometheuses.monitoring.coreos.com
       version: v1
-    - description: Creation of Alertmanager instances
-      displayName: Alertmanager
-      kind: Alertmanager
-      name: alertmanager.monitoring.coreos.com
-      version: v1
     - description: Creation of Smart Gateways
       displayName: Smart Gateway
       kind: SmartGateway
--- a/roles/serviceassurance/defaults/main.yml
+++ b/roles/serviceassurance/defaults/main.yml
@@ -9,4 +9,8 @@ prometheus_storage_resources: {}
 prometheus_storage_selector: {}
 prometheus_pvc_storage_request: 20G
 
+alertmanager_storage_class: ""
+alertmanager_storage_resources: {}
+alertmanager_storage_selector: {}
+alertmanager_pvc_storage_request: 20G
 # vim: set ft=yaml shiftwidth=2 tabstop=2 expandtab:
--- a/roles/serviceassurance/tasks/sa_metrics.yml
+++ b/roles/serviceassurance/tasks/sa_metrics.yml
@@ -3,6 +3,10 @@
 - name: Create Prometheus instance
   include_tasks: component_prometheus.yml
 
+# create Alertmanager instance
+- name: Create Alertmanager instance
+  include_tasks: component_alertmanager.yml
+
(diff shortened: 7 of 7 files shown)


leifmadsen:
In the old setup there was a `secret.yaml` that had a single basic rule that was loaded by Alertmanager. You might need to create that object via another CRD like prometheusrules or something.

leifmadsen:
```
cat service-assurance/alertmanager/secret.yaml 
apiVersion: v1
kind: Secret
metadata:
  name: alertmanager-sa
type: Opaque
stringData:
  alertmanager.yaml: |-
    global:
      resolve_timeout: 5m
    route:
      group_by: ['job']
      group_wait: 30s
      group_interval: 5m
      repeat_interval: 12h
      receiver: 'null'
      routes:
      - match:
          alertname: DeadMansSwitch
        receiver: 'null'
    receivers:
    - name: 'null'
```

leifmadsen:
Ah looking at the docs, it doesn't look like the Prometheus Operator provides an API interface for setting up the Alertmanager configuration (I guess because it'd really just be writing it to a Secret for you).

https://github.com/coreos/prometheus-operator/blob/master/Documentation/user-guides/alerting.md

Configuration of Alertmanager is done as it was in SAF TP, where you'll need to provide an `alertmanager_config_manifest` and write that as a `Secret` object to k8s. Then if someone wants to change the configuration, they would just pass in the new `alertmanager_config_manifest` to the SAO.

mrunge:
name of the secret needs to be alertmanager-saf-default

roles/serviceassurance/tasks/component_alertmanager.yml
```diff
@@ -0,0 +1,35 @@
...
+      kind: Alertmanager
+      metadata:
+        labels:
+          alertmanager: '{{ meta.name }}'
+        name: '{{ meta.name }}'
+        namespace: '{{ meta.namespace }}'
+      spec:
+        replicas: 1
+        ruleSelector: {}
+        securityContext: {}
```
leifmadsen: I would just drop these two lines

leifmadsen:
Name of the secret should be generated based on the name of the SAF instance. The default instance is called `saf-default`. The Alertmanager object is being created with `name: '{{ meta.name }}'` so I think the `alertmanager` part must be getting prepended.

The name of the secret should then probably be `name: alertmanager-{{ meta.name }}`

roles/serviceassurance/tasks/component_alertmanager.yml
```diff
@@ -0,0 +1,35 @@
...
+    alertmanager_manifest: |
+      apiVersion: monitoring.coreos.com/v1
+      kind: Alertmanager
+      metadata:
+        labels:
+          alertmanager: '{{ meta.name }}'
+        name: '{{ meta.name }}'
+        namespace: '{{ meta.namespace }}'
+      spec:
+        replicas: 1
```
csibbitt: I've added a card (#861) to our triage backlog about HA Alertmanager. It may be as simple as putting a 2 in this field, but there may also be adjustments needed on the prometheus side, or a proper way to do it via the operator.

https://prometheus.io/docs/alerting/alertmanager/#high-availability

leifmadsen: From what I've been reading if you just set the replicas to > 1 all the Alertmanagers should coordinate with each other to notice who has sent a notification. When configuring Prometheus, alerts go to all available alertmanagers. The coordination among the AMs is what results in only a single fired event.

Looks like the default docs around using Prometheus Operator suggests using 3 AMs. The work will be around the testing, but the configuration change likely just follows the same pattern as ElasticSearch, where the `main.yml` will get a `set_fact` with `alertmanager_replicas: 3` when `high_available_enabled` is `true`. I think that'll likely be the only change required.

roles/serviceassurance/tasks/component_alertmanager.yml
```diff
@@ -0,0 +1,62 @@
...
+            resolve_timeout: 5m
+          route:
+            group_by: ['job']
+            group_wait: 30s
+            group_interval: 5m
+            repeat_interval: 12h
+            receiver: 'null'
+            routes:
+            - match:
+                alertname: DeadMansSwitch
```
csibbitt: Does this Alert exist in our Prometheus config? I only see rules about ES that I think come from the ECL.

mrunge: I copy/pasted from tech-preview.

Alertmanager is only about sending alerts.

mrunge:
In my understanding, an alert is a rule in prometheusrules, which can be configured through the operator gui, for example, as tab in the prometheus operator.

roles/serviceassurance/tasks/component_alertmanager.yml
csibbitt: > It was copy/pasted from tech-preview.

Okay, but AFAICT it's a routing rule for an alert that doesn't exist. Is that what we want here? We should not copy/paste things if they don't make sense. Does this make sense, or is it an oops?

> Alertmanager is only about sending alerts.

I don't know what you're trying to say?

csibbitt:
> In my understanding, an alert is a rule in prometheusrules, which can be configured through the operator gui, for example, as tab in the prometheus operator.

That sounds right to me, though I haven't tried it yet. As we see above, though, alerts defined in prometheus are sometimes named in the alertmanager config. Does the prometheusrule object take care of that somehow, or is the user expected to use some other process to define the relevant alertmanager config? That's the part I'm asking about. 

Ex. I want the ElasticsearchClusterNotHealthy alert (which actually exists in my SAF Prom right now) to go to my email. What do I do?

mrunge:
> > In my understanding, an alert is a rule in prometheusrules, which can be configured through the operator gui, for example, as tab in the prometheus operator.
> 
> That sounds right to me, though I haven't tried it yet. As we see above, though, alerts defined in prometheus are sometimes named in the alertmanager config. Does the prometheusrule object take care of that somehow, or is the user expected to use some other process to define the relevant alertmanager config? That's the part I'm asking about.
> 
> Ex. I want the ElasticsearchClusterNotHealthy alert (which actually exists in my SAF Prom right now) to go to my email. What do I do?

In short: you are defining a rule in prometheus, and prometheus will send out a notification to alertmanager, which will trigger sending out the notification.

We already have a card for this planned to be worked on next.
https://trello.com/c/CvqExzrx/850-document-creating-rules-in-prometheus-for-raising-alerts

roles/serviceassurance/tasks/component_alertmanager.yml
leifmadsen: @csibbitt note that with the change to ECK that no rules are being populated at all by default. Can you capture the rules you're seeing from ECL? Maybe they will be useful to implement.

leifmadsen: For reference, here is a prometheus configuration for a semi-recent SAF deployment with no AM or PR:

```
global:
  scrape_interval: 30s
  scrape_timeout: 10s
  evaluation_interval: 30s
  external_labels:
    prometheus: sa-telemetry/saf-default
    prometheus_replica: prometheus-saf-default-0
alerting:
  alert_relabel_configs:
  - separator: ;
    regex: prometheus_replica
    replacement: $1
    action: labeldrop
rule_files:
- /etc/prometheus/rules/prometheus-saf-default-rulefiles-0/*.yaml
scrape_configs:
- job_name: sa-telemetry/saf-default/0
  scrape_interval: 1s
  scrape_timeout: 1s
  metrics_path: /metrics
  scheme: http
  kubernetes_sd_configs:
  - role: endpoints
    namespaces:
      names:
      - sa-telemetry
  relabel_configs:
  - source_labels: [__meta_kubernetes_service_label_app]
    separator: ;
    regex: smart-gateway
    replacement: $1
    action: keep
  - source_labels: [__meta_kubernetes_endpoint_port_name]
    separator: ;
    regex: prom-http
    replacement: $1
    action: keep
  - source_labels: [__meta_kubernetes_endpoint_address_target_kind, __meta_kubernetes_endpoint_address_target_name]
    separator: ;
    regex: Node;(.*)
    target_label: node
    replacement: ${1}
    action: replace
  - source_labels: [__meta_kubernetes_endpoint_address_target_kind, __meta_kubernetes_endpoint_address_target_name]
    separator: ;
    regex: Pod;(.*)
    target_label: pod
    replacement: ${1}
    action: replace
  - source_labels: [__meta_kubernetes_namespace]
    separator: ;
    regex: (.*)
    target_label: namespace
    replacement: $1
    action: replace
  - source_labels: [__meta_kubernetes_service_name]
    separator: ;
    regex: (.*)
    target_label: service
    replacement: $1
    action: replace
  - source_labels: [__meta_kubernetes_pod_name]
    separator: ;
    regex: (.*)
    target_label: pod
    replacement: $1
    action: replace
  - source_labels: [__meta_kubernetes_service_name]
    separator: ;
    regex: (.*)
    target_label: job
    replacement: ${1}
    action: replace
  - separator: ;
    regex: (.*)
    target_label: endpoint
    replacement: prom-http
    action: replace
  metric_relabel_configs:
  - separator: ;
    regex: pod
    replacement: $1
    action: labeldrop
  - separator: ;
    regex: namespace
    replacement: $1
    action: labeldrop
  - separator: ;
    regex: instance
    replacement: $1
    action: labeldrop
  - separator: ;
    regex: job
    replacement: $1
    action: labeldrop
```

leifmadsen: Also this route is typically part of the example configurations for Alertmanager since it basically proves that alerts are working. It basically is always firing so that if your remote doesn't see it firing any more they can alert that alerts aren't working :)

I agree that we should really make sure we keep things as clean and intentional as possible here.

leifmadsen: Here are 3 basic objects for a single Alertmanager + Secret + PrometheusRules:

**Secret**
```
apiVersion: v1
kind: Secret
metadata:
  name: alertmanager-saf-default-test
  namespace: sa-telemetry
type: Opaque
stringData:
  alertmanager.yaml: |-
    global:
      resolve_timeout: "5m"
    receivers:
    - name: "null"
    route:
      group_by:
      - namespace
      group_interval: "5m"
      group_wait: "30s"
      receiver: "null"
      repeat_interval: "12h"
      routes:
      - match:
          alertname: "Watchdog"
        receiver: "null"
```

**Alertmanager**
```
apiVersion: monitoring.coreos.com/v1
kind: Alertmanager
metadata:
  creationTimestamp: null
  name: saf-default
  namespace: sa-telemetry
spec:
  replicas: 1
  securityContext: {}
```

**PrometheusRules**
```
apiVersion: monitoring.coreos.com/v1
kind: PrometheusRule
metadata:
  labels:
    prometheus: saf-default
    role: alert-rules
  name: saf-default
  namespace: sa-telemetry
spec:
  groups:
    - name: general.rules
      rules:
        - alert: Watchdog
          annotations:
            message: >
              This is an alert meant to ensure that the entire alerting pipeline
              is functional.

              This alert is always firing, therefore it should always be firing
              in Alertmanager

              and always fire against a receiver. There are integrations with
              various notification

              mechanisms that send a notification when this alert is not firing.
              For example the

              "DeadMansSnitch" integration in PagerDuty.
          expr: vector(1)
          labels:
            severity: none
```

leifmadsen: I think the bare minimum here would be to to set the Secret to:

```
apiVersion: v1
kind: Secret
metadata:
  name: alertmanager-saf-default
  namespace: sa-telemetry
type: Opaque
stringData:
  alertmanager.yaml: |-
    route:
      receiver: dummy
    receivers:
      - name: dummy
```

This results in a configuration built in Alertmanager that looks like this (from the alertmanager webUI):

```
global:
  resolve_timeout: 5m
  http_config: {}
  smtp_hello: localhost
  smtp_require_tls: true
  pagerduty_url: https://events.pagerduty.com/v2/enqueue
  hipchat_api_url: https://api.hipchat.com/
  opsgenie_api_url: https://api.opsgenie.com/
  wechat_api_url: https://qyapi.weixin.qq.com/cgi-bin/
  victorops_api_url: https://alert.victorops.com/integrations/generic/20131114/alert/
route:
  receiver: dummy
receivers:
- name: dummy
templates: []
```

And then set the `Alertmanager` object as above, and you'll have a loaded Alertmanager. Don't load any PrometheusRules at this point as there will be no receiver for them.

leifmadsen: Also in the templates above, `s/saf-default/{{ meta.name }}/g`

leifmadsen:
> How is a user expected to add alerts to the configuration?

A user will first need to create a receiver and add it to the route. This is done in the `alertmanager-{{ meta.name }}` secret object. Whatever receiver name is defined in the route should be the default (based on docs I've read).

From there, you'd just need to add your rules to `PrometheusRules` object, and Prometheus Operator will load them for you.

If you want to start grouping your rules in `PrometheusRules` then you can send those groups of rules can be filtered by the receiver, etc. That would require changes to the `alertmanager.yaml` in the Secret object.

leifmadsen:
https://daenney.github.io/2018/04/21/setting-up-alertmanager

leifmadsen:
> > In my understanding, an alert is a rule in prometheusrules, which can be configured through the operator gui, for example, as tab in the prometheus operator.
> 
> That sounds right to me, though I haven't tried it yet. As we see above, though, alerts defined in prometheus are sometimes named in the alertmanager config. Does the prometheusrule object take care of that somehow, or is the user expected to use some other process to define the relevant alertmanager config? That's the part I'm asking about.
> 
> Ex. I want the ElasticsearchClusterNotHealthy alert (which actually exists in my SAF Prom right now) to go to my email. What do I do? (Assume I have read the fine manual and know all about configuring receivers and routes in alertmanager, just not how to do it in an openshift operator world)

The simple answer here is basically create a default receiver and route in your Secret (holding the `alertmanager.yaml` and that's it. Make sure your rule is then loaded via `PrometheusRules`.

If you want to get fancy, you can start grouping your rules, in which case you can start having your routes filter on just the groups you want that receiver to handle. So if you wanted ES alerts to go to one email, and Metrics alerts go to another email, you could handle that in your `alertmanager.yaml` configuration.

https://prometheus.io/docs/alerting/configuration/#route

csibbitt:
@leifmadsen Is the short version of your answer "The user will supply their own alertmanager_config_manifest via the SAO property"

leifmadsen:
> @leifmadsen Is the short version of your answer "The user will supply their own alertmanager_config_manifest via the SAO property"

Yes. I'm not sure we can provide any sort of default receiver or route for rules, and so the user will need to supply that configuration.

mrunge:
What's missing here?

csibbitt:
@mrunge You never answered my question here: https://github.com/redhat-service-assurance/service-assurance-operator/pull/28/files#r367005132 

> Okay, but AFAICT it's a routing rule for an alert that doesn't exist. Is that what we want here? We should not copy/paste things if they don't make sense. Does this make sense, or is it an oops?

Leif gave us a lot of info on the topic, and if I read correctly, he seems to agree that we should not ship a config that names an alert that doesn't exist. It seems there is a common configuration that uses this config to detect if your monitoring system has broken - that sounds desirable; do we have that though?

I'm happy to give a :+1: if that rule were removed or if I could have a clear indication that it's intentional and/or useful and/or necessary and that works as intended in our environment. If it's just dead code that was copy/pasted, please remove it.

mrunge:
Okay, thank you Chris. I'll push an update asap.
Meanwhile, I drafted some docs https://github.com/mrunge/saf-documentation/blob/promrules/doc-Service-Assurance-Framework/modules/proc_creating-an-alert.adoc

leifmadsen:
Awesome start to the docs. Thanks! Now that you have the rules, I guess next sections would be configuring the `alertmanager.yaml` and receivers.

roles/serviceassurance/tasks/component_alertmanager.yml
```diff
@@ -0,0 +1,56 @@
...
+            repeat_interval: 12h
+            receiver: 'null'
+          receivers:
+          - name: 'null'
+  when: alertmanager_config_manifest is not defined
+
+- name: Create an instance of Alertmanager
+  k8s:
+    definition:
+      '{{ alertmanager_config_manifest }}'
```
leifmadsen: You need to create the Alertmanager instance as well via `alertmanager_manifest`

README.md
```diff
@@ -30,9 +30,19 @@ ServiceAssurance) via the YAML configuration to override the default manifests
 loaded for you.
 
 * prometheus_manifest
+* alertmanager_manifest
```
leifmadsen: Also note that you can load `alertmanager_config_manifest`

README.md
```diff
@@ -30,9 +30,19 @@ ServiceAssurance) via the YAML configuration to override the default manifests
...
 * elasticsearch_secret_manifest
 * interconnect_manifest
 * elasticsearch_manifest
 * smartgateway_metrics_manifest
 * smartgateway_events_manifest
 * servicemonitor_manifest
+
+
+## Local testing ##
```
leifmadsen: Is there a local testing procedure or just the deps?

mrunge: Ah, the deps were first step for documenting local testing. It's not finished, I can remove it, if you prefer.

leifmadsen: I was more just curious. I think if there aren't testing steps then I'd just drop this section for now.

roles/serviceassurance/tasks/component_alertmanager.yml
```diff
@@ -0,0 +1,60 @@
...
+                  storage: {{ alertmanager_pvc_storage_request }}
+  when: alertmanager_manifest is not defined
+
+- name: Set default Alertmanager configuration manifest
+  set_fact:
+    alertmanager_config_manifest: |
+      apiVersion: v1
+      kind: Secret
+      metadata:
+        name: alertmanager-'{{ meta.name }}'
```
leifmadsen: Sorry small nit. Can you just drop the `alertmanager` from this `name` line? When creating an Alertmanager the operator already tacks on `-alertmanager` to the value here, resulting in `alertmanager-saf-default-alertmanager`.

leifmadsen:
Other than the small night I noted this looks good to go. Please provide some test output proving this works and then I'll sign off on it. Thanks!

roles/serviceassurance/tasks/component_alertmanager.yml
mrunge: This is a secret, without alertmanager in the name, it'll become saf-default, I tested this again.

mrunge:
```
oc describe secrets saf-default
Name:         saf-default
Namespace:    sa-telemetry
Labels:       <none>
Annotations:  <none>

Type:  Opaque

Data
====
alertmanager.yaml:  163 bytes
```

mrunge:
```
oc get pods
NAME                                         READY   STATUS              RESTARTS   AGE
elasticsearch-operator-6d6795556c-x9lhs      0/1     ContainerCreating   0          63s
interconnect-operator-55b5f8867b-wr2lk       0/1     ContainerCreating   0          40s
prometheus-operator-68f97b8f79-xz556         1/1     Running             0          52s
service-assurance-operator-7b5946489-6qdmp   2/2     Running             0          9m48s
smart-gateway-operator-6d7544bf95-bqsn5      2/2     Running             0          57s

export SAF_PROJECT=sa-telemetry
 mrunge  euler  ☎  ~  …  service-assurance-operator  master  8⬆  1✎  4+  $    oc create -f - <<EOF
> apiVersion: infra.watch/v1alpha1
> kind: ServiceAssurance
> metadata:
>   name: saf-default
>   namespace: ${SAF_PROJECT}
> spec:
>   metricsEnabled: true
>   eventsEnabled: false
> EOF
serviceassurance.infra.watch/saf-default created

oc get pods
NAME                                          READY   STATUS              RESTARTS   AGE
alertmanager-saf-default-0                    0/2     ContainerCreating   0          3m4s
elasticsearch-operator-6d6795556c-x9lhs       1/1     Running             0          5m29s
interconnect-operator-55b5f8867b-wr2lk        1/1     Running             0          5m6s
prometheus-operator-68f97b8f79-xz556          1/1     Running             0          5m18s
prometheus-saf-default-0                      3/3     Running             1          3m5s
saf-default-interconnect-57bcfc6969-zrppw     1/1     Running             0          3m9s
saf-default-telemetry-smartgateway-1-deploy   0/1     Completed           0          2m53s
saf-default-telemetry-smartgateway-1-k2czd    1/1     Running             0          2m38s
service-assurance-operator-7b5946489-6qdmp    2/2     Running             0          14m
smart-gateway-operator-6d7544bf95-bqsn5       2/2     Running             0          5m23s


oc describe pod alertmanager-saf-default-0 
Name:               alertmanager-saf-default-0
Namespace:          sa-telemetry
....
 Warning  FailedMount  118s (x9 over 4m6s)  kubelet, crc-rk2fc-master-0  MountVolume.SetUp failed for volume "config-volume" : secret "alertmanager-saf-default" not found
```

This worked last time when I used the right secret name.

README.md
mrunge: I've added content here: https://github.com/redhat-service-assurance/service-assurance-operator/issues/30

roles/serviceassurance/tasks/component_alertmanager.yml
leifmadsen: Oops misread. Sorry. Thanks!

roles/serviceassurance/tasks/component_alertmanager.yml
```diff
@@ -0,0 +1,60 @@
+- name: Set default Alertmanager manifest
+  set_fact:
+    alertmanager_manifest: |
+      apiVersion: monitoring.coreos.com/v1
+      kind: Alertmanager
+      metadata:
+        labels:
+          alertmanager: '{{ meta.name }}'
+        name: '{{ meta.name }}'
```
leifmadsen: :+1:

leifmadsen:
One last thing to confirm. You ran the `operator-sdk olm-catalog gen-csv` command to update the `deploy/olm-catalog/service-assurance-operator/0.1.0` contents right? (version 0.12.0 of operator-sdk)

mrunge:
I did, but that removed in deploy/olm-catalog/service-assurance-operator/0.1.0/service-assurance-operator.v0.1.0.clusterserviceversion.yaml everything here: https://github.com/redhat-service-assurance/service-assurance-operator/blob/master/deploy/olm-catalog/service-assurance-operator/0.1.0/service-assurance-operator.v0.1.0.clusterserviceversion.yaml#L34-L106

So the question is, what that causes, especially since I did not change anything there.

leifmadsen:
> I did that just now, but that removed in deploy/olm-catalog/service-assurance-operator/0.1.0/service-assurance-operator.v0.1.0.clusterserviceversion.yaml everything here: https://github.com/redhat-service-assurance/service-assurance-operator/blob/master/deploy/olm-catalog/service-assurance-operator/0.1.0/service-assurance-operator.v0.1.0.clusterserviceversion.yaml#L34-L106
> 
> So the question is, what that causes, especially since I did not change anything there.
> so the change caused by runing build/update_csv.sh is not included in this PR.

Show me the output of `operator-sdk version`. I bet it's > 0.12.0. You can't use the latest `operator-sdk` or that's the output you get.

We need to make sure `olm-catalog` is created via the `operator-sdk` generator, and not by manual changes (unless they are in areas not managed by the `operator-sdk`).

mrunge:
you are right, I have operator-sdk version: "v0.13.0", commit: "1af9c95bb51420c55a7f7f2b7fabebda24451276", go version: "go1.13.3 linux/amd64"

mrunge:
hah! found an earlier version. It seems I did not change anything there, but the HA change was missing.

mrunge:
Anything else or are we good to merge this now?

leifmadsen:
> hah! found an earlier version. It seems I did not change anything there, but the HA change was missing.

And that's why I was adamant about not manually adjusting the `olm-catalog` :)

Looks good to me!

PR #29, opened 2020-01-15
Set Interconnect default config for OSP

leifmadsen:
Set the default configuration for Interconnect to be compatible with OSP13 and 16
connections.

**BLOCKED** _do not merge_

Working on external end to end testing to validate exactly what kind of server side connection we are going to expose for a default `enable-saf.yaml` environment in OSP.

### DIFF
--- a/roles/serviceassurance/tasks/component_qdr.yml
+++ b/roles/serviceassurance/tasks/component_qdr.yml
@@ -26,18 +26,17 @@
             prefix: broadcast
           - distribution: multicast
             prefix: collectd
-        edgeListeners:
-          - expose: true
-            host: 0.0.0.0
-            port: 5671
-            saslMechanisms: ANONYMOUS
-            sslProfile: openstack
         interRouterListeners:
           - authenticatePeer: true
             expose: true
             port: 55671
             saslMechanisms: EXTERNAL
             sslProfile: inter-router
+          - expose: true
+            host: 0.0.0.0
+            port: 5671
+            saslMechanisms: ANONYMOUS
+            sslProfile: openstack
         listeners:
           - port: 5672
           - expose: true


roles/serviceassurance/tasks/component_qdr.yml
```diff
@@ -33,14 +33,13 @@
             saslMechanisms: ANONYMOUS
             sslProfile: openstack
         interRouterListeners:
-          - authenticatePeer: true
-            expose: true
+          - expose: true
             port: 55671
-            saslMechanisms: EXTERNAL
-            sslProfile: inter-router
+            saslMechanisms: ANONYMOUS
```
csibbitt: At first glance I'd expect this to break the inter-router connections that automatically get generated when you ask the operator for >1 QDR. Did you test that?

csibbitt:
Can you please add a bit of text describing exactly what led to this change? I think I can guess, but having a clear idea of what you're attempting to change and why would be helpful to make sure we get it right and don't break anything else.

roles/serviceassurance/tasks/component_qdr.yml
```diff
@@ -33,14 +33,13 @@
             saslMechanisms: ANONYMOUS
             sslProfile: openstack
         interRouterListeners:
-          - authenticatePeer: true
-            expose: true
+          - expose: true
```
mrunge: what does expose mean here? I couldn't find it in the qdrouter documentation?

leifmadsen: https://access.redhat.com/documentation/en-us/red_hat_amq/7.5/html/deploying_amq_interconnect_on_openshift/connecting-clients-router-network-router-ocp#exposing-ports-router-ocp

roles/serviceassurance/tasks/component_qdr.yml
leifmadsen: I didn't test that actually. I'm thinking I might need to actually create a separate inbound connection because I hadn't thought about the multi-QDR scenario. I will review today and do some additional testing.

leifmadsen:
> Can you please add a bit of text describing exactly what led to this change? I think I can guess, but having a clear idea of what you're attempting to change and why would be helpful to make sure we get it right and don't break anything else.

So basically OSP13 connects all the nodes as edge connections, so the existing exposed port 5671 is what we were using. During testing with OSP16 our controllers are supposed to connect as inter-routers (with the non-controllers nodes connecting to the controllers as edge routers).

Since we don't have methods developed for authentication or certificate exchange etc we need to use the same type of connection as previously, but setup as inter-router (not edge).

Going to review again today and see if I can come up with a better solution. I'm likely going to have to put a bunch of it back to how it was, and then just expose a third port for connections from OSP16 environments.

Do you think it would be worth adding some switches around this, such as `interRouterConnectionExposed: true` etc? I'm not sure I really want to add a bunch of configuration options here (based on our previous discussions) but I also feel like maybe exposing 2 unauthenticated AMQP ports by default is also bad? Thoughts?

leifmadsen:
@csibbitt I'm going to actually block this for now until after I finish the end-to-end testing with `enable-saf.yaml`. Having done some discussions last week with Andy and Aaron, I'm not convinced we're going to end up changing QDR mesh by default.

leifmadsen:
Closing this since no change is required for edge mode.

PR #31, opened 2020-01-17
[HOTFIX] Update invalid syntax for alertmanager manifest

leifmadsen:


leifmadsen:
I have tested this and confirmed that it fixes the issue I was seeing in my environment after building and deploying the Service Assurance Operator.

leifmadsen:
Fixes issue introduced in #28

mrunge:
doh! oops. Thank you for the quick fix. I can confirm, it works now.

PR #33, opened 2020-01-24
Set reset_index to false so SG doesn't delete all events

csibbitt:
This is an **untested** change (by me)

Closes https://github.com/redhat-service-assurance/service-assurance-operator/issues/32


leifmadsen:
Should be something that can be confirmed with the SAF development instantiation procedure.

csibbitt:
Merging based on https://github.com/redhat-service-assurance/service-assurance-operator/issues/32#issuecomment-578282609

PR #36, opened 2020-01-27
Adding remove_saf.sh

csibbitt:
* Expunges everything that gets installed with SAF
* Includes the cluster-wide certmanager stuff
* Does not include ECK stuff (since we don't install it automatically)

### DIFF
--- a/deploy/remove_saf.sh
+++ b/deploy/remove_saf.sh
@@ -1,0 +1,25 @@
+#!/bin/sh
+#
+# Removes SAF and amq7 cert manager from your cluster
+#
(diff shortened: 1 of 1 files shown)


csibbitt:
Am I missing anything? I went through a few cycles of `./remove_saf.sh; ./quickstart.sh;` and it looks like it's reliable; no errors from the quickstart.

csibbitt:
[Write an SAF uninstall/cleanup script](https://trello.com/c/bounHyKV/899-write-an-saf-uninstall-cleanup-script)

PR #42, opened 2020-02-10
CI running smoketest

pleimer:
`ci.yml` specifies the build, smoketest, and tear down of the SAO image based on updates in whatever branch contains this file. For the CI Bot to run tests on a branch, this file must be included in the same format as in this PR. Results of this process are posted to a gist that can be accessed by selecting the `details` link when hovering over the commit status:

image

The CI Bot is currently running on Pradeep's local machine in a CRC. As such, this script must unfortunately have machine specific login commands for CRC for the time being.

Commit statuses are based on the exit status of scripts run in the `script` section of `ci.yml`. The `after_scrip` section allows for clean up scripts to execute without interfering with test statuses (it's important to cleanup everything created by the script sections because the CI Bot is running on a persistent CRC with limited resources).

### DIFF
--- a/ci.yml
+++ b/ci.yml
@@ -1,6 +1,4 @@
 script:
-        - cd build
-        - echo $PATH
         - export HOME=$(dirname "$PWD")
         - oc login -u kubeadmin -p BdTY8-Vra64-gD3cI-dAPS6 --insecure-skip-tls-verify=true https://api.crc.testing:6443
         - IMAGE_BUILD_ARGS="--storage-driver vfs" ./build_ci.sh
--- a/ci.yml
+++ b/ci.yml
@@ -1,0 +1,12 @@
+script:
+        - cd build
+        - echo $PATH
+        - export HOME=$(dirname "$PWD")
+        - oc login -u kubeadmin -p BdTY8-Vra64-gD3cI-dAPS6 --insecure-skip-tls-verify=true https://api.crc.testing:6443
+        - IMAGE_BUILD_ARGS="--storage-driver vfs" ./build_ci.sh
+        - cd ../tests/smoketest 
+        - ./smoketest.sh
+after_script:
+        - cd deploy/ 
+        - export HOME=$(dirname "$PWD")
+        - ./remove_saf.sh
--- a/ci.yml
+++ b/ci.yml
@@ -1,8 +1,8 @@
 script:
         - export HOME=$(dirname "$PWD")
         - oc login -u kubeadmin -p BdTY8-Vra64-gD3cI-dAPS6 --insecure-skip-tls-verify=true https://api.crc.testing:6443
-        - IMAGE_BUILD_ARGS="--storage-driver vfs" ./build_ci.sh
-        - cd ../tests/smoketest 
+        - IMAGE_BUILD_ARGS="--storage-driver vfs" ./build/build_ci.sh
+        - cd ./tests/smoketest 
         - ./smoketest.sh
 after_script:
         - cd deploy/ 
--- a/ci.sh
+++ b/ci.sh
@@ -1,3 +1,0 @@
-#!/usr/bin/env sh
-echo "Passing CI with this thorough test"
-


ci.yml
```diff
@@ -0,0 +1,12 @@
+script:
+        - cd build
```
csibbitt: This shouldn't be needed, we tried to make the build scripts path agnostic

ci.yml
```diff
@@ -0,0 +1,12 @@
...
+        - echo $PATH
```
csibbitt: Debugging output can go away when you're done with it

ci.yml
```diff
@@ -0,0 +1,12 @@
...
+        - export HOME=$(dirname "$PWD")
```
csibbitt: The CI system itself (your node.js code in this case) should be setting HOME to be the private workdir of the job rather than requiring every ci.yml to implement something like this.

ci.yml
```diff
@@ -0,0 +1,12 @@
...
+        - oc login -u kubeadmin -p BdTY8-Vra64-gD3cI-dAPS6 --insecure-skip-tls-verify=true https://api.crc.testing:6443
```
csibbitt: Technically this password is actually config owned by the CI server, right? Can we have it passed in through the provided environment (like HOME) instead of hard coding into every ci.yml that wants to use the system?

ci.yml
```diff
@@ -0,0 +1,12 @@
...
+        - IMAGE_BUILD_ARGS="--storage-driver vfs" ./build_ci.sh
```
csibbitt: It's not merged yet, but I'm hoping to eliminate the need for these args. See https://github.com/redhat-service-assurance/service-assurance-operator/pull/43

ci.yml
```diff
@@ -0,0 +1,12 @@
...
+        - cd ../tests/smoketest
```
csibbitt: Should not need, as above

ci.yml
```diff
@@ -0,0 +1,12 @@
+script:
+        - cd build
+        - echo $PATH
+        - export HOME=$(dirname "$PWD")
+        - oc login -u kubeadmin -p BdTY8-Vra64-gD3cI-dAPS6 --insecure-skip-tls-verify=true https://api.crc.testing:6443
+        - IMAGE_BUILD_ARGS="--storage-driver vfs" ./build_ci.sh
+        - cd ../tests/smoketest 
+        - ./smoketest.sh
+after_script:
+        - cd deploy/
```
csibbitt: as above

ci.yml
```diff
@@ -0,0 +1,12 @@
...
+        - cd build
+        - echo $PATH
+        - export HOME=$(dirname "$PWD")
+        - oc login -u kubeadmin -p BdTY8-Vra64-gD3cI-dAPS6 --insecure-skip-tls-verify=true https://api.crc.testing:6443
+        - IMAGE_BUILD_ARGS="--storage-driver vfs" ./build_ci.sh
+        - cd ../tests/smoketest 
+        - ./smoketest.sh
+after_script:
+        - cd deploy/ 
+        - export HOME=$(dirname "$PWD")
```
csibbitt: as above

ci.yml
mrunge: the password is included in the file ~/.crc/cache/<version>/kube.....

csibbitt: Just to help clarify; the code in this ci.yml is running from a ci pod in openshift that does not have ready access to that file. The suggestion is for the for the ci server itself to provide it via the environment so it can be made use of in ci scripts without hard-coding it in each.

To achieve this we should document in the README how to generate an openshift secret from the file Matthias mentions (from the context of the machine running CRC), and the ci server should consume this secret into it's environment and expose it in the environment in passes down to the ci scripts.

Not suggesting this needs to happen before this PR merges, I'd just prefer to see it happen before the 2nd or 3rd instance of this password gets copy/pasted around.

PS. This password is, as far as we can tell, actually shared among all CRC instances from the same build and is in no way secret, so it's appearance in this repo should not be a major concern; though it's not a practice we should like to see in general.

pleimer: I see a kubeconfig at ~/.crc/machines/crc/kubeconfig, I'm not certain if this is a permanent one associated with crc? If so, that can be used to create the secret used by deploy-on-k8s.yml and then we wouldn't need that login command in CI.yml anymore. I'll check into it for a future PR

PR #48, opened 2020-02-13
Phase 2 renaming

### LINKED ISSUE #47: Update usage of `saf` to `stf` in both code and filename
We currently reference `saf` in several of the internal paths. These instances should be updated to `stf.

leifmadsen:
Rename usage of saf into stf, and adjust default namespace

Closes #47


README.md
```diff
@@ -58,10 +58,10 @@ oc delete po -l name=service-telemetry-operator
 ## Tech Preview
 
 See the [official
-documentation](https://redhat-service-assurance.github.io/saf-documentation)
+documentation](https://redhat-service-assurance.github.io/stf-documentation)
 for more information about installing for production-style use cases on OCP3.
 
-Please use [the legacy saf-ocp3 branch](https://github.com/redhat-service-assurance/telemetry-framework/tree/saf-ocp3) for all such installations.
+Please use [the legacy stf-ocp3 branch](https://github.com/redhat-service-assurance/telemetry-framework/tree/stf-ocp3) for all such installations.
```
csibbitt: :-1:  URL does not exist. You could copy that branch in the now-defunct repo, but you could also just leave this URL intact.

Our directive is to avoid shipping a product with "assurance" in the name, not erase all history of us ever having called it that pre-release.

leifmadsen: I caught that last night! I think the fix is local, but I was scared to push while the existing run was going. I'll push it now. Nice eyes :)

leifmadsen: Ha! Commit even shows 17 hours ago! :)

PR #55, opened 2020-02-19
Auto push application artifacts to quay.io

### LINKED ISSUE #3: Merges to master branch should result in new builds to quay.io
Update .travis.yml to automatically build the latest CSV and container image, then push them to the appropriate repositories on quay.io when changes are merged to master branch.

leifmadsen:
Provide a method to automatically push application artifacts on merge to
master.

Closes #3


.travis.yml
```diff
@@ -1,16 +1,16 @@
...
+  - secure: ojZXqjvCIqIbXJN13fma6rMH9A0y4b0B1Z9g/hG5ll3Dw76C6GnoxaH5tzAQuUEa3PgPG6oLsgmKr0drMZ4x6YESSVzLq2JIDUfct/qrYQ4CsUB9IjqO/8S2hxxMYsNbfLars34s6MTzNvReqMIMrxDPTEwc84a6jnmvIaFTTaqN4jQHrspZXcLD479muf1++YFnUgdobBKeS0XofZ/umRtW4thHv/yLSX2dcZ92Hz4Ad//YOromLunCfG1O6F7ggMEmEI4cuUp0J+TjBV4TdRnJz+T3ljAibgudSjUWVXSftxCei8CbwGuCK3f0UsQHS8qTBpylbt5iCFJEQf1MdFHi+yZY26jZqYZNboUmH7swPsA38je3Nd6siIXzPllf1B34O5xBubJ+CCqIBpNUe9cfJOqJ5EoibGYTuQmLdEwwY9OYd2dFvRtflSrKCHbz/8DgBmW+K8L1qGja2J069p0Ef7nuZpnhUqYB/mhExTb7hCKZeIKsVCIlmXzSTYqV8+HcyiQjol1eqehDjNrFaumVO5LX48APe1QpHClB/6uM74ZfCc4xa9UO0soBEYp8dOPsI3fbrRfVBI4gogv9/OLQFWTFZ1SKnjCYMcotFosoxp8x5ETki/VwGNCH0zobrFefnB4U4uX7NsrO4Vrx2zbur1/6NrhX1MChfGjZ1ok=
```
csibbitt: Are you sure that `Lars34s6MT` shouldn't be `Lars34sGMT`?

.travis.yml
```diff
@@ -1,16 +1,16 @@
...
-
+- pip install operator-courier
+- pip install ansible-lint
 script:
-  - operator-courier verify --ui_validate_io deploy/olm-catalog/service-telemetry-operator
-  - ansible-lint roles/servicetelemetry/
+- operator-courier verify --ui_validate_io deploy/olm-catalog/service-telemetry-operator
+- ansible-lint roles/servicetelemetry/
+after_success:
+- ./build/travis.sh
```
csibbitt: Eventually your script block will get too big and the natural name for IT will be `./build/travis.sh` (or `travis_script.sh` which will be confusing), so you might want to call this `travis_after_success.sh` or something.

build/travis.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# Add everything, get ready for commit. But only do it if we're on
+# master. If you want to deploy on different branches, you can change
+# this.
+if [[ "$BRANCH" =~ ^master$|^[0-9]+\.[0-9]+\.X$ ]]; then
+    echo "Branch is master, so push new application to Quay registry"
+    export PATH=$HOME/bin:$PATH
+    curl -L https://github.com/operator-framework/operator-sdk/releases/download/v0.12.0/operator-sdk-v0.12.0-x86_64-linux-gnu -o $HOME/bin/operator-sdk-v0.12.0
+    chmod +x $HOME/bin/operator-sdk-v0.12.0
+    operator-courier verify --ui_validate_io deploy/olm-catalog/service-telemetry-operator
```
csibbitt: This command and the one below it may(?) require `${REL}/../` in the paths in order to execute from anywhere outside the project root. I introduced this concept in this repo, but I'm not too strict on requiring that EVERYTHING work that way as long as we have some sort of consistent rule for which do and which don't.

build/travis.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# Fail on error
+set -e
+
+REL=$(dirname "$0")
+
+# Add everything, get ready for commit. But only do it if we're on
```
csibbitt: Comment bug, it isn't only on master. It's also on `<#>.<#>.X` branches I think

.travis.yml
leifmadsen: I just have whatever was generated/added by the `travis` binary (installed via `gem`).

So no... I'm not sure :)

build/travis.sh
leifmadsen: Since these are run from the location relative to the `.travis.yml` it's not necessary since it was working prior to these changes :)

I could probably move these installs to separate `build/travis/[...].sh` scripts but that's starting to seem like overkill for this particular patch set. What do you think?

build/travis.sh
leifmadsen: Ah ha! Nice catch. I just copy/pasta'd this from the `documentation` repository since it uses a similar trick.

I had actually thought about this wondering about versioned branches and how that would all work. It's actually nice that it's `master` or `number.number.X` as well. Will need to keep that in mind going forward. Might even want to adjust this to something like `v<#>.<#>.<#>`.

Frankly we're going to need to do some testing around releases and upgrades and how to adjust the CSV and package files to allow for Operator upgrades.

Will fix the comment but leave the logic alone for now. Thanks!

leifmadsen:
Addressed some comments left by @csibbitt. I think this is good to go. Unfortunately only a merge to master _truly_ tests this from end to end, but I think it should work since it's the same approach I used for the `documentation` repository to publish our documentation to github pages.

build/travis.sh
csibbitt: I think that's overkill, yeah. I only commented because you used REL inconsistently (on some lines but not others)

PR #70, opened 2020-03-03
Include Tripleo templates for placing ceilometer events on interonnect bus

pleimer:
I am not sure this is what should be merged into the main branch since it uses the enable-legacy-telemetry.yaml, but wanted to add this here to track what I used.


tests/infrared/16/metrics-qdr-connectors.yaml.template
```diff
@@ -2,6 +2,8 @@
 tripleo_heat_templates:
     - /usr/share/openstack-tripleo-heat-templates/environments/metrics/collectd-write-qdr.yaml
     - /usr/share/openstack-tripleo-heat-templates/environments/metrics/qdr-edge-only.yaml
+    - /usr/share/openstack-tripleo-heat-templates/environments/enable-legacy-telemetry.yaml
```
leifmadsen: I guess this is here for the redis setup? This is a dependency for ceilometer setup currently?

PR #74, opened 2020-03-05
Add predefined alerts

mrunge:


deploy/alerts/alerts.yaml
```diff
@@ -0,0 +1,136 @@
+
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  creationTimestamp: null
+  labels:
+    prometheus: example
```
leifmadsen: Our default Prometheus uses `prometheus: '{{ meta.name }}'` which means this should be `prometheus: stf-default` by default for compatibility.

deploy/alerts/alerts.yaml
```diff
@@ -0,0 +1,136 @@
...
+    role: alert-rules
+  name: prometheus-example-rules
```
leifmadsen: `name: stf-default-prometheus-openstack-rules` would be my suggestion.

deploy/alerts/alerts.yaml
```diff
@@ -0,0 +1,136 @@
...
+  namespace: service-telemetry
+spec:
+  groups:
+    - name: ./example.rules
```
leifmadsen: `openstack.rules`

deploy/alerts/simple-comparison.yaml
```diff
@@ -0,0 +1,124 @@
+apiVersion: monitoring.coreos.com/v1
```
leifmadsen: Sorry it's not obvious to me why there are two files?

deploy/alerts/simple-comparison.yaml
```diff
@@ -0,0 +1,124 @@
...
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: fixed-compares
```
leifmadsen: I think you'd set the same `prometheus: stf-default` as above.

deploy/alerts/simple-comparison.yaml
```diff
@@ -0,0 +1,124 @@
...
+    role: alert-rules
+  name: prometheus-alarms-part-two
```
leifmadsen: `name: stf-default-prometheus-openstack-rules-simple`? Not sure what to suggest as I don't know why these can't be merged together.

deploy/alerts/simple-comparison.yaml
```diff
@@ -0,0 +1,124 @@
...
+  namespace: service-telemetry
+spec:
+  groups:
+    - name: ./example.rules
```
leifmadsen: Does this get merged automatically when you use the same file name? See suggested file name from earlier.

leifmadsen:
Please move this to the infrawatch org under a branch there rather than from a fork of the repo.

The testing is failing because it thinks this is a change to master and would have actually resulted in a push to quay.io (this is a bug, probably a security issue).

When it's on a fork, it's impossible to collaborate on the code. Just create a branch like `mrunge-predefined-alerts` or whatever you want to name it. If you don't have permissions on the infrawatch org just ask, because you should.

deploy/alerts/simple-comparison.yaml
mrunge: I have a script to create one alerts.yaml, since there is much repetition. The other file is much simpler.

deploy/alerts/simple-comparison.yaml
mrunge: That doesn't matter.

deploy/alerts/simple-comparison.yaml
mrunge: See above, it's mostly how they became generated.

deploy/alerts/alerts.yaml
mrunge: It works this way as well. Not sure if that's a bug or a feature. I'm happy to change it to `foo-bar-doesn-t-matter`

leifmadsen: I'm just trying to keep naming consistent across a deployment. My suggestions are generally how they would look should they have been deployed by the Operator using a template. It doesn't matter from a "does it work" perspective, but it does matter from a consistency and user experience perspective.

deploy/alerts/simple-comparison.yaml
leifmadsen: Again, consistency with how this would look when deploying via templates and potentially having multiple clouds, etc.

deploy/alerts/simple-comparison.yaml
leifmadsen: The above comment still doesn't really explain why there actually needs to be two separate files other than that's how your script was created. When providing this to a customer it's not going to be obvious why there are two separate files, and what to adjust should they want to customize things based on this example.

deploy/alerts/alerts.yaml
mrunge: Okay, the prometheus operator docs says, it's "best practice" to name it after the used prometheus. That apparently means, it doesn't matter. I'll rename it to follow best practices.

deploy/alerts/simple-comparison.yaml
mrunge: There is no technical need to have it in separate files. It is more convenient to handle though. 
I've tried to add some more expanation in https://github.com/infrawatch/service-telemetry-operator/pull/76/files , deploy/alerts/README.md

deploy/alerts/simple-comparison.yaml
mrunge: it doesn't get merged. I haven't found any docs or reason why this tag is there at all.

PR #76, opened 2020-03-05
Add predefined alerts

mrunge:
As requested in the other PR I'm opening this again.


deploy/alerts/README.md
```diff
@@ -0,0 +1,23 @@
...
+## simple-comparison.yaml
+
+These alerts are much simpler, as there are basically only comparisons
+of two numbers or percentages.
+
+
+
+## Installation
+
+Log into OpenShift and
```
pkilambi: would it make sense to call out the login command?

deploy/alerts/README.md
```diff
@@ -0,0 +1,23 @@
+# Predefined alerts
+
+These alarming rules come in two files described in the following
+
+## alerts.yaml
+These alarms are triggered by a substantial change (over a certain
+amount of standard deviation). This requires a bit of preparation
```
pkilambi: "requires a bit of preparation" this statement indicates some pre work needs to be done but doesnt quite say what it is.  Should we just drop this line or make it a bit less ambiguous.

deploy/alerts/README.md
leifmadsen: Logging in can vary depending on the environment. If they have made it this far they should already know how to login.

deploy/alerts/README.md
mrunge: You're right, this needs a bit more explanation. In the way we are doing this here, it requires about 4 rules to get created per alarm for a single metric.

deploy/alerts/alerts.yaml
```diff
@@ -0,0 +1,136 @@
...
+      rules:
+        - expr: 'rate(collectd_disk_disk_io_time_io_time_total[5m])'
+          record: 'job:iotime:rate_5m'
+        - expr: 'stddev_over_time(job:iotime:rate_5m[1h])'
+          record: 'job:iotime:rate_5m:stddev_over_time_1h'
+        - expr: 'avg_over_time(job:iotime:rate_5m[1h])'
+          record: 'job:iotime:rate_5m:avg_over_time_1h'
+        - alert: iotime warn
+          expr: >-
+            (abs(job:iotime:rate_5m - job:iotime:rate_5m:avg_over_time_1h) / job:iotime:rate_5m:stddev_over_time_1h) > 3
```
csibbitt: Purely curiosity, but how did you decide on 3 StdDevs rather than 2 for a warning? I've actually never modeled this on a real workload to see what's best - is there a best practice you read somewhere, or is this an informed guess? Looking at the use of 3 & 6 StdDevs I'm going to guess "six sigma"?

deploy/alerts/simple-comparison.yaml
```diff
@@ -0,0 +1,124 @@
...
+          expr: >-
+            sum without (endpoint,service,type) (collectd_df_df_inodes{df="root",type="used"})/ (sum without (endpoint,service,type) (collectd_df_df_inodes{df="root",type=~"free|used"})) > 0.6
+          for: 10m
+        - alert: inode usage
+          labels:
+            severity: critical
+          annotations:
+            summary: Inodes usage (critical)
+          expr: >-
+            sum without (endpoint,service,type) (collectd_df_df_inodes{df="root",type="used"})/ (sum without (endpoint,service,type) (collectd_df_df_inodes{df="root",type=~"free|used"})) > 0.6
```
csibbitt: The critical alarm should have a higher threshold than the warning. Suggest 0.8 maybe?

deploy/alerts/alerts.yaml
mrunge: sorry, I didn't see this earlier.

I switched to 3 and 6 stddevs since I was under the impression, with 1 and 2 I was getting too many false alarms.

deploy/alerts/simple-comparison.yaml
mrunge: Oh yes, good catch. That is a typo and should be 0.8, indeed.

PR #78, opened 2020-03-09
Create stable channel

leifmadsen:
Create a stable channel to hold the new 1.0.0-beta1 release. Update the bundle metadata to match.


deploy/olm-catalog/service-telemetry-operator/service-telemetry-operator.package.yaml
```diff
@@ -1,5 +1,7 @@
 channels:
-- currentCSV: service-telemetry-operator.v0.1.1
+- currentCSV: service-telemetry-operator.v1.0.0-beta1
   name: beta
-defaultChannel: beta
+- currentCSV: service-telemetry-operator.v1.0.0-beta1
+  name: stable
```
csibbitt: I think I was slightly confused until I got to this line - This says the current CSV in the stable channel IS the beta CSV. That clarifies things.

roles/servicetelemetry/tasks/component_events_smartgateway.yml
```diff
@@ -29,7 +29,7 @@
         useBasicAuth: true
         elasticUser: '{{ elastic_user }}'
         elasticPass: '{{ elastic_pass }}'
-        containerImagePath: 'quay.io/infrawatch/smart-gateway:latest'
+        containerImagePath: 'quay.io/infrawatch/smart-gateway:v1.0.0-beta1'
```
csibbitt: Reconfirmed that this actually exists.

https://quay.io/repository/infrawatch/smart-gateway?tag=latest&tab=tags

deploy/olm-catalog/service-telemetry-operator/service-telemetry-operator.package.yaml
leifmadsen: Yea so I'm playing around with not needing to create any extra branches to deal with versioning and such, but also still providing us the ability to continue using the `latest` tag to be the deployment our development is done against (at least as a base start, so whatever is merged to master is what would be deployed from the `latest` channel).

Since you can't just set `operator-sdk generate csv --csv-version=latest` (I tried, it didn't like the non-semantic version string), I figured maybe whatever our _next_ version is could just hold `latest` and would be the foundation of the next release.

Release workflow I think will look something like this:

* create working branch
* update package file to set `currentCSV` to version containing `latest` (our next release) [1.0.0-beta2]
* run `operator-sdk generate csv --from-version=<<next_release>> --csv-version=<<future_release_now_latest>>` [1.0.0]
* update <<next_version>> CSV links to point at new container image version that matches release (s/`latest`/`this_version`/g)
* profit?

The issue going through this exercise now is let's say I decide I don't want 1.0.0 but want to create a 1.0.0-beta3. I suppose I could create the proper setup if I used a `--from-version=1.0.0` but then I'd have to go and manually adjust the `replaces` line in the CSV with the version I want to migrate from (1.0.0-beta2 in this case).

If that's the only additional change, then that's not too bad. I'll have to play around with this some more and see where that gets me.

roles/servicetelemetry/tasks/component_events_smartgateway.yml
leifmadsen: This is the last part I'm trying to figure out. I actually want to leave this at `latest` for development/release purposes and only adjust this when I go to do an actual release, with it being only adjusted based on the STF release you're going to deploy with.

So I think what this is going to mean is making this something that gets passed in by the default CR when deploying a `ServiceTelemetry` instance. Since the CSV contains the example CR, we could actually just provide a `smartgatewayContainerImageTag` value into the CR to override the default of `latest`.

Anyways, something else I'll be playing around with tomorrow as well.

leifmadsen:
Closing in favour of a squashed patch set.

PR #87, opened 2020-04-21
Tweaks to result in an OSP16 deployment

leifmadsen:
Changes to the infrared-openstack.sh script to result in the successful deployment of OSP16.0
with Ceph and Tempest added for fun and profit. Parameterizes several options that were static
before to allow for a more flexible script while assuming primarily defaults will be utilized.

Re-adds the internal knowledge because it's been discussed that leaking this doesn't seem to
be a huge issue(?). Can pull it out in a subsequent commit as it'll be killed in a squash.

### DIFF
--- a/tests/infrared/16/infrared-openstack.sh
+++ b/tests/infrared/16/infrared-openstack.sh
@@ -11,18 +11,21 @@ AMQP_PORT=${AMQP_PORT:-443}
 SSH_KEY="${SSH_KEY:-${HOME}/.ssh/id_rsa}"
 NTP_SERVER="${NTP_SERVER:-clock.redhat.com,10.5.27.10,10.11.160.238}"
 
-VM_IMAGE_URL_PATH="${VM_IMAGE_URL_PATH:-http://127.0.0.1/my_image_location/}"
-if [ "${VM_IMAGE_URL_PATH}" = "http://127.0.0.1/my_image_location/" -a -z "${VM_IMAGE}" ]; then
-    echo "Please provide a VM_IMAGE_URL_PATH or VM_IMAGE"
-    exit 1
-fi
+VM_IMAGE_URL_PATH="${VM_IMAGE_URL_PATH:-http://download.eng.bos.redhat.com/brewroot/packages/rhel-guest-image/8.1/333/images}"
 # Recommend these default to tested immutable dentifiers where possible, pass "latest" style ids via environment if you want them
-VM_IMAGE="${VM_IMAGE:-${VM_IMAGE_URL_PATH}/rhel-guest-image-8.1-413.x86_64.qcow2}"
-OSP_BUILD="${OSP_BUILD:-RHOS_TRUNK-16.0-RHEL-8-20200204.n.1}"
+VM_IMAGE="${VM_IMAGE:-rhel-guest-image-8.1-333.x86_64.qcow2}"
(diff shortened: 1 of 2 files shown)


leifmadsen:
Tested with the following command on internal lab server:

```
LIBVIRT_DISKPOOL="/home/libvirt/images" AMQP_HOST="stf-default-interconnect-5671-service-telemetry.apps.stfcloudops1.lab.upshift.rdu2.redhat.com" ./infrared-openstack.sh
```

Default RHEL8 installation doesn't result in a large root partition, so `LIBVIRT_DISKPOOL` pointed at `/home/libvirt/images/` after a `mkdir -p /home/libvirt/images/` results in success.

The output is quite long and not particularly exciting, so I'll ask you trust me that it worked :) (I can show the environment for a day or so before I re-deploy again as I don't believe the `enable-legacy-telemetry.yaml` is necessary(?)

tests/infrared/16/infrared-openstack.sh
```diff
@@ -11,29 +11,33 @@ AMQP_PORT=${AMQP_PORT:-443}
 SSH_KEY="${SSH_KEY:-${HOME}/.ssh/id_rsa}"
 NTP_SERVER="${NTP_SERVER:-clock.redhat.com,10.5.27.10,10.11.160.238}"
 
-VM_IMAGE_URL_PATH="${VM_IMAGE_URL_PATH:-http://127.0.0.1/my_image_location/}"
-if [ "${VM_IMAGE_URL_PATH}" = "http://127.0.0.1/my_image_location/" -a -z "${VM_IMAGE}" ]; then
-    echo "Please provide a VM_IMAGE_URL_PATH or VM_IMAGE"
-    exit 1
-fi
+VM_IMAGE_URL_PATH="${VM_IMAGE_URL_PATH:-http://download.eng.bos.redhat.com/brewroot/packages/rhel-guest-image/8.1/333/images}"
```
csibbitt, line 14: :+1:

tests/infrared/16/infrared-openstack.sh
```diff
@@ -11,29 +11,33 @@ AMQP_PORT=${AMQP_PORT:-443}
...
 # Recommend these default to tested immutable dentifiers where possible, pass "latest" style ids via environment if you want them
-VM_IMAGE="${VM_IMAGE:-${VM_IMAGE_URL_PATH}/rhel-guest-image-8.1-413.x86_64.qcow2}"
-OSP_BUILD="${OSP_BUILD:-RHOS_TRUNK-16.0-RHEL-8-20200204.n.1}"
+VM_IMAGE="${VM_IMAGE:-rhel-guest-image-8.1-333.x86_64.qcow2}"
```
csibbitt, line 16: Interesting that this appears to be going backwards. As long as you're sure. :)

leifmadsen: I only picked what magic combination worked for me based on what QE is testing in Jenkins. Since this is the base image in the Phase 1 that QE is testing with, I just aligned to it regardless of my thoughts :)

(It's possible the job is wrong, but I'm not sure how to determine that.)

leifmadsen:
I see a bunch of approvals and this worked for me, so I'm merging it. I expect this to be a living document based on our ongoing testing so others are encouraged to make additional PR changes.

Frankly I think we should really just move this internal and out of STO repo, but that's a separate discussion.

PR #88, opened 2020-04-21
Update recommended memory defaults in crc-stf.sh

leifmadsen:


leifmadsen:
@csibbitt I don't use this script. I just happened to notice before going to bed the values in this script were not what I would recommend for a local deployment so just wanted it to reflect reality.

It could probably be mostly dropped, but since it's here it should at least reflect best practices.

PR #90, opened 2020-04-23
Call python3 to align with container image in smoketest

leifmadsen:


leifmadsen:
This seems to work, so I'm :cowboy_hat_face: merging it.

PR #92, opened 2020-04-28
Changes based on current deployment testing

leifmadsen:


tests/infrared/16/infrared-openstack.sh
```diff
@@ -18,69 +18,94 @@ VM_IMAGE_LOCATION="${VM_IMAGE_URL_PATH}/${VM_IMAGE}"
 
 OSP_BUILD="${OSP_BUILD:-RHOS_TRUNK-16.0-RHEL-8-20200406.n.1}"
 OSP_VERSION="${OSP_VERSION:-16}"
-OSP_TOPOLOGY="${OSP_TOPOLOGY:-undercloud:1,controller:1,compute:1,ceph:1}"
```
leifmadsen, line 21: This might not actually be necessary any more since the issues with deploy I was seeing appear to be self-inflicted, and not related to the Ceph deployment components.

leifmadsen:
I should add some README.md updates to provide an example command that results in an Overcloud deployment being successful.

PR #98, opened 2020-06-26
Latest OSP16 test procedures

leifmadsen:


leifmadsen:
This is currently running in my environment. I will verify this once the cloud comes up.

PR #102, opened 2020-07-07
Changes to infrared-openstack.sh for OSP13

leifmadsen:
These changes result in an OSP13 deployment, but don't result in an STF enabled overcloud.


leifmadsen:
I have re-tested this, and everything comes up now with the simplified `stf-connectors.yaml`. It doesn't result in what we ultimately want an STF configuration to look like, but it does result in an end-to-end deployment finishing successfully.

CC: @csibbitt

PR #112, opened 2020-07-16
Support for Smart Gateway compile changes

leifmadsen:
Adds changes to the CI system to support the changes for Smart Gateway moving to
CentOS 8 and go mod. Was having some issues with building the Smart Gateway image
so this adds some additional logging output on the console. Also does some clean up
by removing the first build which is expected to fail. (Failure is expected because
on creation of the BuildConfig it will fire a Build, but it fails due to how the
Build is configured on auto-trigger. Subsequent builds pass.)

### DIFF
--- a/build/stf-run-ci/tasks/create_builds.yml
+++ b/build/stf-run-ci/tasks/create_builds.yml
@@ -2,10 +2,13 @@
 - name: Create BuildConfig and ImageStream
   shell: oc new-build --name {{ artifact.name }} --dockerfile - < {{ artifact.working_build_dir }}/{{ artifact.dockerfile_path }}
 
+- name: Kill first build since it will always fail (triggered on BuildConfig creation)
+  shell: sleep 10 ; oc delete build {{ artifact.name }}-1
+
 - name: Kick off build
   block:
     - name: Start local image build
-      command: oc start-build {{ artifact.name }} --wait --from-dir "{{ artifact.working_build_dir }}"
+      command: oc start-build {{ artifact.name }} --wait --follow --from-dir "{{ artifact.working_build_dir }}"
       register: build_name
   always:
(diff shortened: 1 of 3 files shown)


build/stf-run-ci/tasks/create_builds.yml
```diff
@@ -2,10 +2,13 @@
 - name: Create BuildConfig and ImageStream
   shell: oc new-build --name {{ artifact.name }} --dockerfile - < {{ artifact.working_build_dir }}/{{ artifact.dockerfile_path }}
 
+- name: Kill first build since it will always fail (triggered on BuildConfig creation)
```
csibbitt, line 5: Why does it always fail?

leifmadsen: Because when you do a BuildConfig it creates a Build automatically (and you can't squelch that unfortunately). It seems to fail on importing the contents of the repo as a binary and fails the Build. I could never figure out why it failed like that, but it seems like I wasn't the only one to have that issue. It has something to do with how the binary/contents of the source are copied in during `BuildConfig` creation.

Doing any subsequent builds with `oc start-build` seems to result in successful builds though.

leifmadsen:
I don't understand why it just seems to hang after the sg-core image finishes building. I don't have that issue in my CRC environment...

leifmadsen:
Paul asked me a really great question... "what version of CRC are you running?" I'm running 1.12 (OCP 4.4.8) and his reinstall of the system just uses `master` so it's OCP 4.5. Maybe there is an issue with the version? He's gonna downgrade CRC to 1.12 and we'll try again.

In the meantime, I re-ran locally and it worked without issue. Here is the massive log dump.

https://pastebin.com/0KMzX5B3

PR #114, opened 2020-08-20
Promxy POC in test directory

csibbitt:
Not much to say about this other than it works to fill in gaps when doing HA Prometheus. We're unlikely to support this downstream, preferring Loki/Observatorium as our target, but this may still be useful for upstream usage, or for various testing scenarios.

### DIFF
--- a/tests/promxy/README.md
+++ b/tests/promxy/README.md
@@ -1,0 +1,5 @@
+# Promxy POC
+
+This is a test deployment of promxy to help validate HA work.
+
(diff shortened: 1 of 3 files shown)


csibbitt:
[Experiment with promxy to solve Prom HA deficiencies](https://trello.com/c/euFg0OoW/796-experiment-with-promxy-to-solve-prom-ha-deficiencies)

leifmadsen:
Very safe :) Won't break anything. Ship it!

PR #115, opened 2020-08-26
Basic Graphing Support

leifmadsen:
Adds some basic graphing support to instantiate a Grafana instance and Prometheus datasource.


leifmadsen:
image

leifmadsen:
[Add Grafana Operator support to STO](https://trello.com/c/X4D7dI3o/1040-add-grafana-operator-support-to-sto)

roles/servicetelemetry/templates/manifest_grafana.j2
```diff
@@ -0,0 +1,25 @@
...
+kind: Grafana
+metadata:
+  name: {{ meta.name }}
+  namespace: {{ meta.namespace }}
+spec:
+  ingress:
+    enabled: {{ graphing_spec_ingress_enabled }}
+  config:
+    auth:
+      disable_signout_menu: true
```
csibbitt, line 11: Suggest false unless there is a good reason

Makefile
```diff
@@ -0,0 +1,12 @@
+.PHONY: all oc-create-build oc-start-build
```
csibbitt, line 1: Doesn't seem related to graphing but OK

roles/servicetelemetry/templates/manifest_grafana.j2
leifmadsen: Sorry, the GH output makes it confusing. Are you referring to `graphing_spec_ingress_enabled` or `disable_signout_menu`?

If the former, I only put it to what our current default already is, but I did look at this and thought about making this inverted to our current default? What do you think? I feel it's safe enough to revert it.

Maybe I should also add another parameter to make it easier to expose?

Makefile
leifmadsen: Heh, I can drop it, I was just using it for local builds, but you're right :)

csibbitt: Keep it, seems useful!

roles/servicetelemetry/templates/manifest_grafana.j2
csibbitt: Sorry, `disable_signout_menu` is what I was referring to. I think it should be enabled, not disabled.

Now that you mention the ingress, though, considering that the defaults are not very secure, I think it should be off by default with an option to enable it at the top level.

leifmadsen:
> This looks great!
> 
> Not sure if it's appropriate here or not, but this file has the other datasources we might want to connect to (STF elasticsearch and openshift specific monitoring prometheue): https://github.com/infrawatch/service-telemetry-operator/blob/b03f96da71570896b8d178c6fe63e8b68c405f11/tests/performance-test/grafana/datasource.yaml

I had actually intended to add other data sources here, especially around ElasticSearch. What I didn't know when looking was actually how to do that.

For example, I was looking at your performance dashboarding work and how I could "solve" for that by making it available somehow simply by importing a dashboard, and then configuring `ServiceTelemetry` object to support it. The issue I had primarily was around the `database` (index) parameter, and what should go there by default? So then I got thinking I need to provide a parameter where someone can specify a list of indices, and then loop through and create ES connections?

Do you have any thoughts on what adding other datasources might look like and how those would need to be configured? I think I'll open that separately, but maybe there's some obvious configs I'm not thinking of.

Additionally since none of our existing dashboards provide annotations via ES, and we don't have an aggregated logging dashboard yet, I decided to leave that for future scoping.

mrunge (approved):
looks good and sane to me.

I did not test this though.

Makefile
leifmadsen: I just moved it to a separate PR #116

csibbitt:
> The issue I had primarily was around the database (index) parameter, and what should go there by default? So then I got thinking I need to provide a parameter where someone can specify a list of indices, and then loop through and create ES connections?

Oh I just looked at this and learned that we create lots of different indexes. I wonder if there is a better way than a dedicated datasource for each? I would need some time to play with this and think through the implications before making a good suggestion.

> ...since none of our existing dashboards provide annotations via ES, and we don't have an aggregated logging dashboard yet, I decided to leave that for future scoping.

I agree that we should constrain the default datasources to those that are actually used by the dashboard we will provide; so these can be left for later.

roles/servicetelemetry/templates/manifest_grafana.j2
leifmadsen: @csibbitt ok I've created a new commit that adds the changes for defaults to both false. They are not exposed as a parameter to ServiceTelemetry (at least via API). We can rely on the auto-passthrough, but long term we should not do that.

So basically, as it stands now if you know the specific knobs to turn you can easily override from the `ServiceTelemetry` object, but they are not explicitly presented via the API (CRD, specDescriptors).

Are we ok with that?

csibbitt: Sounds ok to me.

leifmadsen:
OK I'm going to merge this down for now. Gives us some simpler methods forward to standing up Grafana even if it lacks ES datasources and default dashboards.

I've tested creating additional datasources and importing dashboards from the Operator, so everything still works effectively the same with a slightly easier approach to standing the environment up.

Future work will include ES datasources once we have dashboards that make use of them.

PR #118, opened 2020-08-27
Use valid parameter for PVC in ES

### LINKED ISSUE #117: volumeClaimTemplate size not used in ElasticSearch manifest
The generated manifest for the `nodeSets` contains:
```yaml
    volumeClaimTemplate:
    - metadata:
        name: elasticsearch-data
      spec:
        accessModes:
        - ReadWriteOnce
        resources:
          requests:
            storage: 20Gi
```

The created storage has only 1Gi volumes. It seems that's it's due to a simple syntax error on `volumeClaimTemplate[s]` as it works as expected with:
``` yaml
    volumeClaimTemplates:
    - metadata:
        name: elasticsearch-data
      spec:
        accessModes:
        - ReadWriteOnce
        resources:
          requests:
            storage: 20Gi
```

I'm not sure if it's the same with `Prometheus` and `Alertmanager` manifests.

leifmadsen:
Use the proper (plural) form for volumeClaimTemplates in the ElasticSearch manifest since the
elasticsearches.elasticsearch.k8s.elastic.co CRD uses a list (plural) vs a singular dictionary
(prometheuses.monitoring.coreos.com etc).

This change allows specifying a requested PersistentVolume size via a PersistentVolumeClaim.
Prior to this change the request would only be 1Gi not the value requested.

Closes #117

### DIFF
--- a/roles/servicetelemetry/templates/manifest_elasticsearch.j2
+++ b/roles/servicetelemetry/templates/manifest_elasticsearch.j2
@@ -36,7 +36,7 @@ spec:
             - emptyDir: {}
               name: elasticsearch-data
 {% else %}
-      volumeClaimTemplate:
+      volumeClaimTemplates:
         - metadata:
             name: elasticsearch-data
           spec:


leifmadsen:
Untested as I don't have an environment with dynamic PVC creation. Based on what I'm looking at in the CRDs this is the correct change though.

PR #119, opened 2020-08-28
Dictionary-based API and Multi-Cloud Support

leifmadsen:
I'm opening this issue as a discussion point around a future API interface (CRD) for the `ServiceTelemetry` object (`servicetelemetrys.infra.watch`). Most of the functionality to support this is implemented in this PR, but untested.

Additionally, I didn't the tasks to allow the Operator to support and map the existing API (flat) into the dictionary, but would, should we approve this approach.

The pattern here follows the same as used in the kiali-operator at https://github.com/kiali/kiali-operator and also uses a custom filter plugin from that repository.

By expanding this to a dictionary, it makes it cleaner when adding additional parameters and features. This would be in preparation for multi-cloud support (where a list of objects could now be passed) and allows us to override portions of the dictionary without having to specify the entire dictionary in the CustomResource.

The API pattern is to create a top-level "function" parameter (events, metrics, transport, storage, etc) and then create application-specific configuration changes within that (`metrics.prometheus`, `events.elasticsearch`, `transport.qdr`, `storage.ephemeral`, etc). If we had future endpoints we were going to support (Kafka, Observatorium, etc) then they would have a natural place to land without totally polluting the top-level API.

Not looking to merge this yet (needs more work, testing, backward compatibility added, smoketest updates, etc), but would like feedback on the general approach and API.


deploy/olm-catalog/service-telemetry-operator/1.1.0/service-telemetry-operator.v1.1.0.clusterserviceversion.yaml
```diff
@@ -60,36 +60,48 @@ spec:
         version: v1
       specDescriptors:
       - description: Whether to enable collection and storage components for metrics
-        displayName: Metrics Enabled
-        path: metricsEnabled
+        displayName: Metrics
```
leifmadsen, line 63: The pattern in the specDescriptors changes a bit (not much) which leverages the natural grouping control noted here: https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md#fieldgroup

It makes the UI look a bit nicer too, and allows us to group the top-level API abstractions a bit easier.

image

roles/servicetelemetry/tasks/main.yml
```diff
@@ -28,24 +28,33 @@
...
+    msg: "{{ msg.split('\n') }}"
+
 - name: Adjust defaults for HA
-  when: high_availability_enabled
+  when: high_availability.enabled
   set_fact:
-    elasticsearch_node_count: 3
-    elasticsearch_redundancy_policy: FullRedundancy
-    qdr_deployment_size: 2
+    servicetelemetry_vars.events.elasticsearch.node_count: 3
```
csibbitt, line 43: This is a lot more verbose, darn near java-like, but I think it's worth it.

roles/servicetelemetry/vars/main.yml
```diff
@@ -1,2 +1,45 @@
 ---
 # vars file for servicetelemetry
+servicetelemetry_vars:
+  metrics: |
+    {%- if metrics is defined and metrics is iterable -%}
+    {{ servicetelemetry_defaults.metrics | combine((metrics | stripnone), recursive=True) }}
```
csibbitt, line 6: This is the magic right here. Super useful!

roles/servicetelemetry/tasks/main.yml
leifmadsen: We could always change `servicetelemetry_vars` to `stv` or something if we hated the verbosity :)

roles/servicetelemetry/vars/main.yml
leifmadsen: I take no credit for it other than to copy and paste it :)

leifmadsen:
You'll also notice the CRD is not only more verbose, but actually starts to cover a lot more parameters that we were using, but were not actually covered in our existing CRD, so that'll be a big change even beyond this. Unfortunately, there is currently no current methods to generate the CRD from the Ansible role, unlike in a Golang operator where you can decorate the code, and the CRDs and such are auto-generated for you.

We'll need to be careful to make sure we keep the CRD up to date manually as we add / change functionality.

I suspect we want to bump the API version. I'm considering a v1beta1 bump from v1alpha1.

leifmadsen:
An example CR using the new formatting for reference. Note that the `servicetelemetry_vars` is an internal construct and not exposed to the API itself.

```
apiVersion: infra.watch/v1alpha1
kind: ServiceTelemetry
metadata:
  name: stf-default
  namespace: service-telemetry
spec:
  events:
    enabled: false
  graphing:
    enabled: true
  highAvailability:
    enabled: false
  metrics:
    enabled: true
  storage:
    ephemeral:
      enabled: true
```

roles/servicetelemetry/vars/main.yml
jmazzitelli: Do you know the blood, sweat, and toil I went through to finally come up with this? I still have the scars. :-D

leifmadsen: OMG I can only imagine. It's pretty epic though :)

I did some testing on a smaller sample before going through all this work, and it was awesome.

roles/servicetelemetry/defaults/main.yml
```diff
@@ -1,39 +1,57 @@
...
+  qdr:
+    deployment_size: 1
+
+
+graphing:
+  grafana:
+    ingress_enabled: false
+    admin_password: secret
+    admin_user: root
+    disable_signout_menu: false
```
jmazzitelli, line 48: I'm pretty sure this entire section needs to go under a top-level `servicetelemetry_defaults` dict - this is how the magic works in the vars/main.yaml. [See where your vars/main.yml line 6 uses `servicetelemetry_defaults`](https://github.com/infrawatch/service-telemetry-operator/pull/119/files#diff-6d0d7b6f88656c55602ecc09636b1337R6) ?? That var should be defined here in your defaults/main.yml

leifmadsen: Ah yes! I knew that actually, but it seems I made a mod somewhere and removed it. Nice catch! I haven't tested this yet, but I'm sure I'd have been confused for some time :) Will fix!

leifmadsen: Oh I know what I did. I had that on my _other_ branch that I was experimenting with, then I ported the logic to a separate branch, and I forgot to add that whe I did that.

roles/servicetelemetry/vars/main.yml
csibbitt: > Do you know the blood, sweat, and toil I went through to finally come up with this? I still have the scars. :-D

Thanks!

roles/servicetelemetry/templates/manifest_smartgateway_events.j2
```diff
@@ -11,8 +11,8 @@ spec:
   debug: false
   elasticUrl: 'https://elasticsearch-es-http.{{ meta.namespace }}.svc.cluster.local:9200'
   useBasicAuth: true
-  elasticUser: '{{ elastic_user }}'
-  elasticPass: '{{ elastic_pass }}'
+  elasticUser: '{{ elastic_user | default('elastic') }}'
```
leifmadsen, line 14: For note why this isn't exposed, it's because these are basically done via internal lookup to populate, so I thought exposing them are parameters was not appropriate. Keeps the logic a bit simpler too.

mrunge:
Thank you for the PR.

TBH, I am twofold here. 

First of all, the proposal makes most things much more readable. On the other side, I am having an issue with files like deploy/crds/infra.watch_servicetelemetrys_crd.yaml (the proposed changes). Level of indents are not good to identify (you can see clearly something is indented), but in most cases, and especially when code blocks are getting longer, it is hard to see the context, or to find when a piece is misaligned.
 
If we can find a way to work around that issue, the proposal would be perfect.

leifmadsen:
> Thank you for the PR.
> 
> TBH, I am twofold here.
> 
> First of all, the proposal makes most things much more readable. On the other side, I am having an issue with files like deploy/crds/infra.watch_servicetelemetrys_crd.yaml (the proposed changes). Level of indents are not good to identify (you can see clearly something is indented), but in most cases, and especially when code blocks are getting longer, it is hard to see the context, or to find when a piece is misaligned.
> 
> If we can find a way to work around that issue, the proposal would be perfect.

I'm not really sure I follow the issue, here can you elaborate?

The layout of the CRD is basically a pattern we need to follow, and there isn't anything I can do about that. Everything follows this pattern:

```
<parameter>:
  description: <desc>
  properties:
    <parameter>:
      description: <desc>
      type: <type>
```

If you look at other CRDs, it will follow the same configuration. You can't change how it is represented.

mrunge:
Yes, it is clear and a shortcoming of yaml.

leifmadsen:
> Yes, it is clear and a shortcoming of yaml.

Got it. I'll assume then no issues :)

csibbitt:
I sympathize with @mrunge 's comments about getting lost in the CRD YAML. I find them absolutely ridiculous to read, but I don't think that's a problem we can really solve here.

Using a development environment with good YAML support can really help. I've seen screenshots of even better setups (details unknown to me) but personally I use https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml  and it provides the context of the node your cursor is on in the header of the editor window like this:

image

It's still a bit brutal, but at least I'm not spending all my time scrolling up and down trying to figure out where I am.

mrunge (approved):

deploy/olm-catalog/service-telemetry-operator/1.1.0/service-telemetry-operator.v1.1.0.clusterserviceversion.yaml
leifmadsen: I've started down trying to figure out how to make this work. I have a feeling this isn't working as documented (but will in the future?)

leifmadsen:
Looking at the `collectors` section in the `clouds` dictionary, I wonder if we could replace the following?

* `collector_type` for `source`
* `subscription_address` for `address`

leifmadsen:
Going back to the CRD readability issues, I even more sympathize after going through the exercise of building the CRD file out for real.

A trick I did learn though is that you are going to always add a `type: object` after a `properties` header.

The pattern will basically go:

```
<my_object>:
  description: Something about this object
  properties:
    <something_nested>
  type: object
```

Every object pretty much follows this pattern. From there you can start nesting your other objects. If you have a flat nested dictionary, it will look like this:

```
metrics:
  description: Some sweet metrics configuration
  properties:
  type: object
```

I will then go back and add my nested dictionary items. Note that your new parameters go _between_ the `properties` and `type` parameters. If you look for this pattern it will make reading the CRD easier.

```
metrics:
  description: Some sweet metrics configuration
  properties:
    enabled:
      description: Enable those metrics
      type: boolean
  type: object
```

I can just add more parameters now:

```
metrics:
  description: Some sweet metrics configuration
  properties:
    enabled:
      description: Enable those metrics
      type: boolean
    storage:
      description: Storage stuff
      properties:
       further_nested_property: ...
      --> OOOOOO!!!! I saw a properties! Don't forget the type: object! See next line
      type: object
  type: object
```

The lists objects are a bit special but not too much harder.

We start with our basic object:

```
clouds:
  description: Some sweet metrics configuration
  properties:

  type: object
```

Except we're going to replace `properties` with `items`. And we're going to replace `object` with `array`.

```
clouds:
  description: Some sweet metrics configuration
  items:
    
  type: array
```

And now we're going to build out the structure allowing a list of objects.

```
clouds:
  description: Some sweet metrics configuration
  items:
    properties:

    type: object
  type: array
```

Now add your properties for the object.


```
clouds:
  description: Some sweet metrics configuration
  items:
    properties:
      name:
        description: Name of the cloud
        type: string
    type: object
  type: array
```

csibbitt (approved):

deploy/crds/infra.watch_servicetelemetrys_crd.yaml
```diff
@@ -5,91 +5,294 @@ metadata:
...
+            properties:
+              qdr:
+                description: QDR configuration for data transport
+                properties:
+                  enabled:
+                    description: Enable QDR data transort
+                    type: boolean
+                type: object
+            type: object
+          graphing:
```
pleimer, line 146: sweet!

pleimer (approved):
This looks great

leifmadsen:
> question Did you mean to remove this file? AFAICT it's still referenced here: https://github.com/infrawatch/service-telemetry-operator/blob/master/build/test-local/start.sh#L2 but it's been awhile since I did local testing of this code, I'm not sure if maybe that script is also deprecated?

I did mean to remove it, but I did forget to look for other references to the file. I'm not sure this even works any more (or is really meant to be used this way).

I now use the build-push-delete method somewhat documented in the docs/development.md and have pretty much never used the local operator method.

I will also delete that script unless someone else happens to be using this in their development workflow (I assume not since I'm pretty sure I was the one that created this).

leifmadsen:
> > I now use the build-push-delete method somewhat documented in the docs/development.md and have pretty much never used the local operator method.
> 
> That's what I do to. I remember running the local operator in the very beginning, but haven't touched it in awhile. Good to drop this script and the template.

I know this PR is getting a bit out of hand, but if I'm going to do a major refactor I might as well drop things that are likely to be more confusing than useful.

deploy/olm-catalog/service-telemetry-operator/manifests/service-telemetry-operator.clusterserviceversion.yaml
```diff
@@ -39,6 +62,130 @@ spec:
...
+      - kind: ReplicaSets
+        name: ""
+        version: v1
+      - kind: Deployments
+        name: ""
+        version: v1
+      - kind: Services
+        name: ""
+        version: v1
+      specDescriptors:
```
leifmadsen, line 83: image

image

image

image

roles/servicetelemetry/tasks/component_prometheus.yml
```diff
@@ -9,6 +9,7 @@
 
 - name: Create an instance of Prometheus
   k8s:
+    merge_type: merge
```
leifmadsen, line 12: Will need to re-test this. I was working through some issues but I don't think this is necessary.

watches.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-- version: v1alpha1
+- version: v1beta1
```
leifmadsen, line 2: We don't watch multiple versions because when you create a v1alpha1 object it is converted to v1beta1 (not the `spec` though). We just do checks in Ansible looking for the old parameters and drop into a conversion routine. In the future we could look at updating/rewriting the current CR possibly.

Existing objects are also converted over to the new API version automatically, so there is no need to watch multiple objects and have separate entry points.

.travis.yml
```diff
@@ -1,6 +1,6 @@
 env:
   global:
-  - RELEASE_VERSION: v0.17.2
+  - RELEASE_VERSION: v0.19.4
```
leifmadsen, line 3: Using the matching version from the base image results in the bundles showing up as invalid when using a multi-version CRD. Using the latest non-1.x version seems to resolve that issue in Travis-CI and locally when building.

.travis.yml
```diff
@@ -20,8 +20,8 @@ before_install:
 
 install:
 - pip install ansible-lint
+- ansible-galaxy collection install operator_sdk.util
```
leifmadsen, line 23: I install the operator_sdk.util collection here because I was having issues with setting paths or whatever `ansible-lint` was expecting for the vendored collection...

build/stf-run-ci/tasks/setup_base.yml
```diff
@@ -40,7 +40,6 @@
         name: amq7-cert-manager
         source: redhat-operators
         sourceNamespace: openshift-marketplace
-        startingCSV: amq7-cert-manager.v1.0.0
```
leifmadsen, line 43: Just start at the latest version as `startingCSV` will still upgrade through to the latest CSV version in the channel. I'm not aware of ability to pin. The `startingCSV` is probably only useful to us in upgrade testing scenarios I guess?

deploy/crds/infra.watch_v1alpha1_servicetelemetry_cr.yaml
```diff
@@ -1,11 +1,9 @@
...
 metadata:
-  name: stf-default
+  name: default
 spec:
   metricsEnabled: true
   eventsEnabled: false
-  highAvailabilityEnabled: false
-  storageEphemeralEnabled: false
-  graphingEnabled: false
+  graphingEnabled: true
```
leifmadsen, line 8:
```suggestion
  graphingEnabled: false
```

tests/smoketest/minimal-collectd.conf.template
```diff
@@ -11,12 +11,12 @@ LoadPlugin cpu
...
 <Plugin "amqp1">
   <Transport "name">
-    Host "stf-default-interconnect"
+    Host "default-interconnect"
     Port "5672"
     Address "collectd"
     <Instance "telemetry">
         Format JSON
-        PreSettle true
+        PreSettle false
```
csibbitt, line 19: :+1:

csibbitt (approved):
I gave this another once-over and haven't found anything new to comment on. This PR is hard to meaningfully review at this point; let's get it merged ASAP and address any deficiencies in smaller more digestible chunks.

leifmadsen:
> I gave this another once-over and haven't found anything new to comment on. This PR is hard to meaningfully review at this point; let's get it merged ASAP and address any deficiencies in smaller more digestible chunks.

Great thanks for looking this over again. It's definitely a significantly larger refactor than I had been expecting...

I'm going to do one final build/test locally removing this `merge` parameter to the `k8s` module to make sure it's not necessary, and then I'll merge this down and get an upstream release candidate created.

PR #126, opened 2020-09-18
Adjust release deployment scripts

leifmadsen:


leifmadsen:
Since this doesn't really affect anyone but me during the release process, I'm just merging this down :)

PR #127, opened 2020-09-21
Adjust load alerts to account for CPU count

leifmadsen:
Adjust the load alerts in this repository to account for number of CPUs. This is
not necessarily a catch-all net, but should at least keep the alerts notifications
for load levels on nodes to something resembling sanity (especially when nodes differ
greatly in their CPU counts).

### DIFF
--- a/deploy/alerts/alerts.yaml
+++ b/deploy/alerts/alerts.yaml
@@ -187,7 +187,7 @@ spec:
           annotations:
             summary: load longterm (warning)
           expr: >-
-            collectd_load_longterm > 0.7 and collectd_load_longterm < 0.9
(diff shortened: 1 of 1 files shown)


leifmadsen:
Also side question, "does anyone think we should break alerts out into a separate repository, or perhaps migrate it to the `dashboards` repository?"

mrunge (approved):

mrunge:
I don't think alerts should get migrated to dashboards. If you'd want to move it out of STO, then please create a new repo.

Technically, alert definitions are not related to STO, I'd be fine with moving them out.

paramite (approved):

leifmadsen:
Yea I figured a separate repository makes more sense anyways. I only thought dashboards since they are somewhat related. Repositories are cheap though.

PR #129, opened 2020-09-24
Add a script to build and push STO

mrunge:
for local tests.

This builds a new STO and uploads it to the local registry.
Afterwards, the newly built STO is deployed.

### DIFF
--- a/build-and-push.sh
+++ b/build-and-push.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+
+eval $(crc console --credentials | grep "admin, run" | cut -f2 -d"'") --insecure-skip-tls-verify=true
+
+
+REGISTRY=$(oc registry info)
+TOKEN=$(oc whoami -t)
+INTERNAL_REGISTRY=$(oc registry info --internal=true)
+buildah login --tls-verify=false -u openshift -p "${TOKEN}" "${REGISTRY}"
+buildah bud -f build/Dockerfile -t "${REGISTRY}/service-telemetry/service-telemetry-operator:latest" .
+buildah push --tls-verify=false "${REGISTRY}/service-telemetry/service-telemetry-operator:latest"
+
+oc apply -f - <<EOF
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: smartgateway-operator
+  namespace: service-telemetry
+spec:
+  channel: stable
+  installPlanApproval: Automatic
+  name: smartgateway-operator
+  source: infrawatch-operators
+  sourceNamespace: openshift-marketplace
+  startingCSV: smart-gateway-operator.v2.0.0
+EOF
+
+oc delete servicetelemetry stf-default
+
+CSV_VERSION=1.1.0
+
+oc delete csv service-telemetry-operator-v${CSV_VERSION}
+
+oc apply \
+    -f deploy/role_binding.yaml \
+    -f deploy/role.yaml \
+    -f deploy/service_account.yaml \
+    -f deploy/operator_group.yaml
+
+oc apply -f deploy/olm-catalog/service-telemetry-operator/${CSV_VERSION}/infra.watch_servicetelemetrys_crd.yaml
+oc apply -f <(sed "\
+    s|image: .\+/service-telemetry-operator:.\+$|image: ${INTERNAL_REGISTRY}/service-telemetry/service-telemetry-operator:latest|g;
+    s|namespace: placeholder|namespace: service-telemetry|g"\
+    "deploy/olm-catalog/service-telemetry-operator/${CSV_VERSION}/service-telemetry-operator.v${CSV_VERSION}.clusterserviceversion.yaml")
+
+oc apply -f - <<EOF
+apiVersion: infra.watch/v1alpha1
+kind: ServiceTelemetry
+metadata:
+  name: stf-default
+  namespace: service-telemetry
+spec:
+  highAvailabilityEnabled: false
+  eventsEnabled: true
+  metricsEnabled: true
+  storageEphemeralEnabled: false
+  snmpTrapsEnabled: true
+  snmpTrapsTarget: "192.168.36.9"
+EOF


build-and-push.sh
```diff
@@ -0,0 +1,59 @@
...
+  installPlanApproval: Automatic
+  name: smartgateway-operator
+  source: infrawatch-operators
+  sourceNamespace: openshift-marketplace
+  startingCSV: smart-gateway-operator.v2.0.0
+EOF
+
+oc delete servicetelemetry stf-default
+
+CSV_VERSION=1.1.0
```
paramite, line 30: Can we make this as a parameter?

build-and-push.sh
```diff
@@ -0,0 +1,59 @@
...
+spec:
+  channel: stable
+  installPlanApproval: Automatic
+  name: smartgateway-operator
+  source: infrawatch-operators
+  sourceNamespace: openshift-marketplace
+  startingCSV: smart-gateway-operator.v2.0.0
+EOF
+
+oc delete servicetelemetry stf-default
```
paramite, line 28: I'm not sure if the building and pushing script should be deleting this.

build-and-push.sh
```diff
@@ -0,0 +1,59 @@
...
+metadata:
+  name: stf-default
+  namespace: service-telemetry
+spec:
+  highAvailabilityEnabled: false
+  eventsEnabled: true
+  metricsEnabled: true
+  storageEphemeralEnabled: false
+  snmpTrapsEnabled: true
+  snmpTrapsTarget: "192.168.36.9"
```
paramite, line 58: Should we hardcode the IP address?

build-and-push.sh
mrunge: yes, sure.

build-and-push.sh
mrunge: you'd want to remove the old operator and also the deployment in order to test the new operator.

build-and-push.sh
mrunge: This is an external system, where SNMP traps are sent to. This is a config detail and has nothing really to do with the operator itself.

Honestly, I don't care about the address for now.

mrunge:
Thank you for your review. This is probably one of 3-5 scripts to build and push a new STO. I understand this as tool which helped me a lot with testing the snmp traps stuff. I am expecting this to be helpful for anyone playing with or changing STO. I'd expect anybody working with it to change the script, e.g turn off snmp, enable/disable events etc. Not everyone will use crc,....

build-and-push.sh
leifmadsen: If you just source the `metadata.sh` file then you'll get a bunch of parameters. You'll need to adjust this line depending on your relative position, but it already has the `CSV_VERSION` defined and parameterized.

```
REL=$(dirname "$0"); source "${REL}/../build/metadata.sh"
```

build-and-push.sh
leifmadsen: You can also use the `remove_stf.sh` which exists in `deploy/`

leifmadsen:
> Thank you for your review. This is probably one of 3-5 scripts to build and push a new STO. I understand this as tool which helped me a lot with testing the snmp traps stuff. I am expecting this to be helpful for anyone playing with or changing STO. I'd expect anybody working with it to change the script, e.g turn off snmp, enable/disable events etc. Not everyone will use crc,....

There is also the `deploy/configs/` directory which has `ServiceTelemetry` objects you can load into scripts. We don't really use those any more, but that's kind of how we pulled out the environment specific information previously.

build-and-push.sh
```diff
@@ -0,0 +1,59 @@
...
+
+eval $(crc console --credentials | grep "admin, run" | cut -f2 -d"'") --insecure-skip-tls-verify=true
+
+
+REGISTRY=$(oc registry info)
+TOKEN=$(oc whoami -t)
+INTERNAL_REGISTRY=$(oc registry info --internal=true)
+buildah login --tls-verify=false -u openshift -p "${TOKEN}" "${REGISTRY}"
+buildah bud -f build/Dockerfile -t "${REGISTRY}/service-telemetry/service-telemetry-operator:latest" .
+buildah push --tls-verify=false "${REGISTRY}/service-telemetry/service-telemetry-operator:latest"
```
leifmadsen, line 11: This is definitely the faster way to deploy than using the `BuildConfig` setup.

build-and-push.sh
```diff
@@ -0,0 +1,59 @@
...
+
+oc apply -f - <<EOF
+apiVersion: infra.watch/v1alpha1
+kind: ServiceTelemetry
+metadata:
+  name: stf-default
+  namespace: service-telemetry
+spec:
+  highAvailabilityEnabled: false
+  eventsEnabled: true
```
leifmadsen, line 54: This won't result in the events path being setup since I don't see a `Subscription` for ECK.

leifmadsen (commented):
Script seems safe enough and isn't going to cause any issues. I wonder if the more appropriate spot though is to add some of this as a task to `build/stf-run-ci/tasks/` though? Some of the functionality is already dealt with there, the only real thing is doing a local build via buildah and pushing into the local registry. Would probably be pretty straight forward to add that and then add a bit of documentation showing how to run this same style scenario via Ansible?

Anyways, if this seems useful to folks generally it's fine going in with me. Longer term I think I'd like to see us drive towards Ansible and molecule for testing and stand up, then provide some documentation on what commands to run to get a development environment up and running.

build-and-push.sh
mrunge: right. Using the scripts from deploy would remove this here.

leifmadsen:
Thinking about this again, I honestly think this could be accomplished by adding a small playbook that would run the `buildah` stuff locally and then leverage the rest of the automation in `build/stf-run-ci/`. If we wanted to use this script it would need to be migrated to use the ServiceTelemetry/v1beta1 CRD interface.

The migration for that is already done in the `build/stf-run-ci/` environment. I'm afraid of having to maintain all these scripts when we could write it all in to the existing Ansible role...

mrunge:
I have this script, which did the job quite okay when dealing with snmp traps development. I'm not going to invest the time to redo it in ansible.

PR #130, opened 2020-09-25
Add prefix to virtual resources

pleimer:
Adding the prefix options separates out virtual resources used on a machine such that `infrared virsh --cleanup` will not interfere with other VMs on the same host.

### DIFF
--- a/tests/infrared/13/infrared-openstack.sh
+++ b/tests/infrared/13/infrared-openstack.sh
@@ -25,6 +25,8 @@ LIBVIRT_DISKPOOL="${LIBVIRT_DISKPOOL:-/var/lib/libvirt/images}"
 
 TEMPEST_ONLY="${TEMPEST_ONLY:-false}"
 
+PREFIX="{PREFIX:-osp}"
+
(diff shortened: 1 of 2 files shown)


leifmadsen:
Can you make this a parameter that could be passed through? I think this is part of the solution to being able to run multiple clouds on the same infrastructure / hardware.

csibbitt (approved):
I wonder if this makes it easier to run ir and crc on the same machine for whole system testing?

pleimer:
> I wonder if this makes it easier to run ir and crc on the same machine for whole system testing?

Yes it absolutely does. Now when you run the osp deploy scripts, it wont delete the CRC vms and screw up the CRC virtual network

csibbitt:
Thanks, Paul!

leifmadsen:
Oooo! I look forward to adjust our test scripts to stand up an end-to-end environment with co-located OSP and CRC!

leifmadsen:
My biggest interest is running 2 small clouds on the same hardware, and pointing it at STF for multi-cloud testing. Thanks for this!

PR #133, opened 2020-10-19
Add operator-sdk Ansible collection

leifmadsen:
Add the operator-sdk Ansible collection in preparation for Service Telemetry
Operator refactoring.

### DIFF
--- a/collections/ansible_collections/operator_sdk/util/.gitignore
+++ b/collections/ansible_collections/operator_sdk/util/.gitignore
@@ -1,0 +1,1 @@
+operator_sdk-util-*.tar.gz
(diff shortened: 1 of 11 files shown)


leifmadsen:
Nothing is using this yet. I just want to merge and loop it back to the refactoring PR to reduce the size of changes.

csibbitt (approved):
Approving on the assumption that vendoring this code into our tree is the preferred solution.

If this is a workaround of some sort, please add some commentary to the commit message to explain.

leifmadsen:
> Approving on the assumption that vendoring this code into our tree is the preferred solution.
> 
> If this is a workaround of some sort, please add some commentary to the commit message to explain.

I couldn't get the Ansible to dynamically import this at run time, which does seem awkward when running the Operator. In a disconnected deployment that would be impossible. At build time is also awkward downstream since the build services don't have ability to pull from an external interface.

I expect this to be part of a future base-image when we migrate to operator-sdk 1.x which is OCP 4.7 timeframe, so I expect this to be temporary.

PR #134, opened 2020-10-21
Post-refactor clean up

leifmadsen:
Initial round of post-refactor clean up based on operator-sdk 0.19.4 and the
new generate commands. Updates the manifests and metadata per the new release
of operator-sdk and also validates that the scorecard can be run. Adjustments
made based on the scorecard recommendations were made. More to come.


deploy/crds/infra.watch_v1beta1_servicetelemetry_cr.yaml
```diff
@@ -3,13 +3,26 @@ kind: ServiceTelemetry
 metadata:
   name: default
 spec:
+  alerting:
+    enabled: true
+    alertmanager:
+      storage:
+        strategy: ephemeral
```
leifmadsen, line 10: There is a discussion ongoing about what the default should be at (ephemeral vs persistent).

I am considering setting the undefined default to be `persistent` (as it was in 1.0.3) but to set the example CR (which is what you'd get if you didn't override anything when doing a `ServiceTelemetry` object creation from the UI where this CR is loaded) to `ephemeral` to make the user experience a bit nicer when doing a test deployment.

@csibbitt what do you think of that approach? I'm not sure if that actually satisfies the concern of having customers running ephemeral in production though?

deploy/olm-catalog/service-telemetry-operator/manifests/service-telemetry-operator.clusterserviceversion.yaml
```diff
@@ -80,93 +102,101 @@ spec:
       - kind: Services
         name: ""
         version: v1
+      - kind: ServiceMonitors
```
leifmadsen, line 105: This was recommended to be added by the `operator-sdk scorecard` run since we're creating `ServiceMonitor` objects as part of the deployment.

deploy/olm-catalog/service-telemetry-operator/manifests/service-telemetry-operator.clusterserviceversion.yaml
```diff
@@ -80,93 +102,101 @@ spec:
...
+        name: servicemonitors.monitoring.coreos.com
+        version: v1
       specDescriptors:
       - description: Enable Prometheus for metrics backend.
         displayName: Prometheus enabled
         path: backends.metrics.prometheus.enabled
         x-descriptors:
-        - 'urn:alm:descriptor:com.tectonic.ui:fieldGroup:Backends'
+        - urn:alm:descriptor:com.tectonic.ui:fieldGroup:Backends
```
leifmadsen, line 113: Updating the CSV via `operator-sdk generate bundle` resulted in some clean up to the CSV with the 0.19.4 version of operator-sdk, so you'll see a bunch of these changes which were not made by me manually :)

deploy/olm-catalog/service-telemetry-operator/manifests/service-telemetry-operator.clusterserviceversion.yaml
```diff
@@ -284,7 +315,7 @@ spec:
                 - /usr/local/bin/ao-logs
                 - /tmp/ansible-operator/runner
                 - stdout
-                image: image-registry.openshift-image-registry.svc:5000/service-telemetry/service-telemetry-operator:latest
+                image: quay.io/infrawatch/service-telemetry-operator:latest
```
leifmadsen, line 318: Make manually loading in the local CSV a bit easier to consume by not using the local build (which would require some setup to make exist).

deploy/operator.yaml
```diff
@@ -19,14 +19,14 @@ spec:
           - /usr/local/bin/ao-logs
           - /tmp/ansible-operator/runner
           - stdout
-          image: image-registry.openshift-image-registry.svc:5000/service-telemetry/service-telemetry-operator:latest
+          image: quay.io/infrawatch/service-telemetry-operator:latest
```
leifmadsen, line 22: Changing the `image` value here is what will be placed into the `CSV` when the bundle is generated.

deploy/crds/infra.watch_v1beta1_servicetelemetry_cr.yaml
csibbitt: It seems rather strange to have one default behavior from the CLI and a different one if you click through the example defaults in the OLM GUI; so I'm not sure I can get behind that solution either. And yeah, it doesn't really address my concern. My personal opinion on this is two parts:

1. Having a default production setup that doesn't lose my data is more important than a default production setup that "just works" and does lose my data. 
2. If this default is meant to make things easier "in development" then I suggest the patch that I always apply after I `git pull` this project:

```
[csibbitt@laptop service-telemetry-operator (master) ]$ git diff
diff --git a/deploy/quickstart.sh b/deploy/quickstart.sh
index 3bafad7..6899f9a 100755
--- a/deploy/quickstart.sh
+++ b/deploy/quickstart.sh
@@ -5,4 +5,5 @@ oc new-project "${OCP_PROJECT}"
 ansible-playbook \
     --extra-vars namespace="${OCP_PROJECT}" \
     --extra-vars __local_build_enabled=false \
+    --extra-vars __service_telemetry_storage_ephemeral_enabled=true \
     ${REL}/../build/run-ci.yaml
```
 Ideally we'd have a PV based solution that "just works" for some prescribed platform (OCS for example?) and get the best of both worlds. In the meantime I prefer the safe default to the easy one.

leifmadsen: OK that works for me. I will change this in a subsequent PR.

csibbitt (approved):
Approving with reservation noted about the ephemeral storage

mrunge:
I was *just* looking at this.

leifmadsen:
> I was _just_ looking at this.

Ah sorry about that. Feel free to provide additional feedback and I will address it. I have several inbound PRs.

PR #141, opened 2020-10-26
Update lastTransitionTime only when changed for ephemeralStorageEnabled status object

leifmadsen:


leifmadsen:
Tested this to validate that the reconcile loop was no longer looping indefinitely, and that the status field was still getting updated.

PR #144, opened 2020-10-30
Allow for environment file selection

leifmadsen:
Provides environment file selection between the stf and stf+gnocchi
configuration. Removes the PREFIX option which was causing deployment
issues. Drops the size of the controllers to a more reasonable size.


tests/infrared/16.1/infrared-openstack.sh
```diff
@@ -1,116 +0,0 @@
-#!/usr/bin/env bash
-set -e
-
```
mrunge, line 3: Not sure where this is coming from. There is no 16.1 dir(?)

mrunge:
I briefly thought about this yesterday (re. including gnocchi and ceilometer), and also to get rid of the directories per release.
Either we should get rid of all release numbers, or we should probably move common parts into a common directory.
Templates and default configurations in 16.1 and 13 are different, and I wouldn't be too astonished, if there are breaking changes between 16.2 and 16.1, too.

Personally, I wouldn't put too much energy into these scripts, they are plain helper scripts. If you have the time to optimize and to refactor, you're welcome.

tests/infrared/16.1/infrared-openstack.sh
leifmadsen: In your PR there is. This PR is against your PR.

leifmadsen:
> I briefly thought about this yesterday (re. including gnocchi and ceilometer), and also to get rid of the directories per release.
> Either we should get rid of all release numbers, or we should probably move common parts into a common directory.
> Templates and default configurations in 16.1 and 13 are different, and I wouldn't be too astonished, if there are breaking changes between 16.2 and 16.1, too.
> 
> Personally, I wouldn't put too much energy into these scripts, they are plain helper scripts. If you have the time to optimize and to refactor, you're welcome.

That's literally what this PR is doing.

tests/infrared/16.1/infrared-openstack.sh
mrunge: Ugh, I missed this is a PR against my PR. The UI utterly sucks here.

tests/infrared/16/infrared-openstack.sh
```diff
@@ -21,10 +21,10 @@ OSP_VERSION="${OSP_VERSION:-16.1}"
 OSP_TOPOLOGY="${OSP_TOPOLOGY:-undercloud:1,controller:3,compute:2,ceph:3}"
 OSP_MIRROR="${OSP_MIRROR:-rdu2}"
 LIBVIRT_DISKPOOL="${LIBVIRT_DISKPOOL:-/var/lib/libvirt/images}"
+ENVIRONMENT_TEMPLATE="${ENVIRONMENT_TEMPLATE:-stf-connectors.yaml.template}"
 
 TEMPEST_ONLY="${TEMPEST_ONLY:-false}"
-
-PREFIX="${PREFIX:-osp}"
```
csibbitt, line 27: Ah yes I recently ran into trouble with this; though I thought it was because I was moving from a PREFIX'd environment to a non-PREFIX'd one. IIRC @pleimer added this so that we could deploy infrared and CRC on the same machine - is this an abandoned feature, or is there a different solution now?

Do we understand what is wrong with the current usage of PREFIX? It seems like it really should work if we've done it righ.

tests/infrared/16/infrared-openstack.sh
```diff
@@ -47,10 +46,9 @@ ir_run_provision() {
       --host-key "${SSH_KEY}" \
       --image-url "${VM_IMAGE_LOCATION}" \
       --host-memory-overcommit True \
-      -e override.controller.cpu=8 \
-      -e override.controller.memory=32768 \
-      --serial-files True \
-      --prefix "${PREFIX}"
+      -e override.controller.cpu=4 \
```
csibbitt, line 49: I did this manually the other day. Glad to see it here

tests/infrared/16/infrared-openstack.sh
```diff
@@ -122,5 +120,8 @@ else
   stf_create_config
   ir_create_overcloud
   ir_expose_ui
+fi
+
+if ${RUN_WORKLOAD}; then
```
csibbitt, line 125: Ah yes, this ALSO bit me the other day. I think I was missing certain ir plugins or something to make it work

csibbitt (approved):
LGTM except I'd like to understand what the real problem is with the PREFIX option, and whether there are other solutions for what it was intended to help with.

tests/infrared/16/infrared-openstack.sh
leifmadsen: I'm not sure what is wrong with the current usage of PREFIX, but it caused all sorts of deployment troubles for me. I had asked the infrared folks about this, and they said in theory it should work, but that they never use it. I suspect there is a bug in some of the Ansible logic or otherwise, and that it's not actually working.

I'd prefer to just see it dropped until someone can spend time doing some additional testing that doesn't cause the script to fail at the first step (clean up).

tests/infrared/16/infrared-openstack.sh
mrunge: I've never see the launch_workload option working properly without any issue. It always stopped sooner or later, sometimes after uploading the cirrus image, sometimes during floating IP creation.

mrunge (approved):

PR #145, opened 2020-11-04
Adjust smoketests and deployment for SNMPtraps

leifmadsen:
Initial implementation to get quickstart.sh and smoketests.sh updated to account for
snmptraps testing against STF 1.1. Smoketests didn't pass for me but now it completes
without error.


mrunge (approved):
Thank you, looks good.

I know smoketests for snmptraps worked before. Do you recall the concrete error?

leifmadsen:
> Thank you, looks good.
> 
> I know smoketests for snmptraps worked before. Do you recall the concrete error?

The smoketest itself I think was fine (I haven't gotten it to pass yet or send a trap as that wasn't my focus). I did notice I missed the change from `stf-default` to `default` so I fixed that up in the smoketest.sh script.

Our CI runs smoketests against a quickstart created environment, and I noticed quickstart wasn't starting up the SNMPtraps, so that's the main area provided in this patch.

mrunge:
> The smoketest itself I think was fine (I haven't gotten it to pass yet or send a trap as that wasn't my focus). I did notice I missed the change from `stf-default` to `default` so I fixed that up in the smoketest.sh script.

That change (`stf-default` to `default`) bit me in my local deployments here and costed me some time.

Thanks!

csibbitt (approved):
LGTM

PR #149, opened 2020-11-19
Enable libpodstats and sensubility amqp1 transport

pleimer:
Enable libpodstats plugin and configure sensubility to use amqp1 bus to store in ES


pleimer:
This should not be merged until https://github.com/infrawatch/smart-gateway/pull/95 is merged

PR #150, opened 2020-11-24
Fix storageclass selection

csibbitt:
While testing compatibility with Openshift Container Storage, I found that the existing settings for picking a storageclass did not work. I manually tested the `storageClassName` element in the new place in the AlertMan and Prom templates and it worked as expected, so I wrote this up to adjust.


csibbitt:
## Not yet tested

deploy/crds/infra.watch_servicetelemetrys_crd.yaml
```diff
@@ -55,9 +55,6 @@ spec:
                           storageClass:
                             description: Storage class name used for Alertmanager PVC
                             type: string
-                          storageResources:
```
csibbitt, line 58: This param was being used in the wrong part of the template so could never have had any effect. I think it's safe to simply remove it.

deploy/crds/infra.watch_v1beta1_servicetelemetry_cr.yaml
```diff
@@ -33,7 +31,8 @@ spec:
         storage:
           strategy: persistent
           persistent:
-            pvcStorageRequest: 20Gi
+            storageSelector: {}
+            pvcStorageRequest: 20G
```
leifmadsen, line 35: Gi vs G makes a big difference between ElasticSearch and Prometheus interfaces.

csibbitt, line 35: Noticed this size inconsistency. I don't know of any storage system that will automatically re-provision to a smaller size once a PVC is bound, but if there are concerns about this change I could drop it.

deploy/olm-catalog/service-telemetry-operator/manifests/service-telemetry-operator.clusterserviceversion.yaml
```diff
@@ -48,7 +47,8 @@ metadata:
                   "enabled": false,
                   "storage": {
                     "persistent": {
-                      "pvcStorageRequest": "20Gi"
+                      "pvcStorageRequest": "20G",
```
leifmadsen, line 50: Gi vs G is on purpose.

roles/servicetelemetry/defaults/main.yml
```diff
@@ -42,7 +40,9 @@ servicetelemetry_defaults:
         storage:
           strategy: persistent
           persistent:
-            pvc_storage_request: 20Gi
+            storage_class: ""
+            storage_selector: {}
+            pvc_storage_request: 20G
```
leifmadsen, line 45: Gi vs G

deploy/olm-catalog/service-telemetry-operator/manifests/service-telemetry-operator.clusterserviceversion.yaml
```diff
@@ -482,6 +481,8 @@ spec:
                   value: service-telemetry-operator
                 - name: ANSIBLE_GATHERING
                   value: explicit
+                - name: PROMETHEUS_WEBHOOK_SNMP_IMAGE
```
csibbitt, line 484: This is not my change, but it came is as part of the bundle update. It must not have been updated previously after this was added to `build/operator.yaml`.

deploy/olm-catalog/service-telemetry-operator/metadata/annotations.yaml
```diff
@@ -1,6 +1,6 @@
 annotations:
   operators.operatorframework.io.bundle.channel.default.v1: stable
-  operators.operatorframework.io.bundle.channels.v1: latest
+  operators.operatorframework.io.bundle.channels.v1: stable,latest
```
csibbitt, line 3: @leifmadsen I used the bundle update command you showed me - is this the right channels setting (I haven't looked this up yet to be 100% sure I know what it means)

deploy/crds/infra.watch_v1beta1_servicetelemetry_cr.yaml
leifmadsen: Actually this might not be true any more. It used to matter because the request was to the ElasticSearch object directly, but now I'm looking at the docs again, and at some point it seems to have switched to just using volumeClaimTemplates, so fixing the inconsistency is probably the correct thing to do here.

Just needs a test run I guess.

csibbitt:
Removed the 20Gi to 20G change because it resulted in the following:
```
# persistentvolumeclaims "elasticsearch-data-elasticsearch-es-default-0" was not valid:
# * spec.resources.requests.storage: Forbidden: field can not be less than previous value
```

deploy/olm-catalog/service-telemetry-operator/metadata/annotations.yaml
leifmadsen: oh right, sorry, actually you want `--channels latest` and `--default-channel stable`.

* --channels is a list of channels this bundle can be a part of -- we only want HEAD to be in `latest` channel
* --default-channel defines the bundles default preferred channel -- we want this as stable so that we don't auto-install `latest` channel

bundle.Dockerfile
```diff
@@ -4,7 +4,7 @@ LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
 LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=service-telemetry-operator
-LABEL operators.operatorframework.io.bundle.channels.v1=latest
+LABEL operators.operatorframework.io.bundle.channels.v1=stable,latest
```
leifmadsen, line 7: Should probably revert this back to just `latest` since that's what the HEAD bundle should probably target.

csibbitt:
# Testing

1. Started from Leif's already manually deployed STF (no infrawatch subscriptions to mess with things)
1. Deleted existing stf object and all pvcs `oc delete stf default; oc delete pvc --all`
1. Built my operator with `build.sh`
1. Editted `deploy/olm-catalog/service-telemetry-operator/manifests/service-telemetry-operator.clusterserviceversion.yaml` to point to my new image and `oc apply -f <..sto-csv.yaml>`
1. Deployed a default STF from the UI, all defaults except I turned ES on
    1. Verified that the "resources" param is removed, and all three had the "selector" and "class" params
1. Verified the pvcs were created with the default storage class ("standard", Cinder-backed, non-OCS):

    ```bash
    $ oc get pvc
    NAME                                             STATUS       VOLUME                                     CAPACITY   ACCESS MODES       STORAGECLASS   AGE
    alertmanager-default-db-alertmanager-default-0   Bound        pvc-6b74e598-ff3a-4d1d-948f-94d193ae75f7   19Gi       RWO                standard       55s
    elasticsearch-data-elasticsearch-es-default-0    Bound        pvc-db3610dd-4c8a-426d-ac7b-08de2254035f   20Gi       RWO                standard       58s
    prometheus-default-db-prometheus-default-0       Bound        pvc-7054ce9b-4cc4-4368-9b7e-abc5e16d890d   19Gi       RWO                standard       68s
    ```

1. Wipe it out `oc delete stf default; oc delete pvc --all`
1. Redeploy from UI, this time selecting `ocs-storagecluster-ceph-rbd` from the pre-populated storageClass dropdown
1. Verified the pvcs were created with the OCS storageclass:

    ````bash
    $ oc get pvc
    NAME                                             STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS                  AGE    ````
    alertmanager-default-db-alertmanager-default-0   Bound    pvc-d568268e-1864-426d-9daa-4a69ade58862   19Gi       RWO            ocs-storagecluster-ceph-rbd   2m18s    ````
    elasticsearch-data-elasticsearch-es-default-0    Bound    pvc-bba7664e-2e6a-4509-a292-c170d2041546   20Gi       RWO            ocs-storagecluster-ceph-rbd   45s    ````
    prometheus-default-db-prometheus-default-0       Bound    pvc-d132e038-7279-4adc-81f9-cfe835d849fa   19Gi       RWO            ocs-storagecluster-ceph-rbd   56s    ````
    ````

As of this patch, I'd say that STF supports OCS for persistent storage of metrics and events.

I did not test any scenarios involving changing the class of already provisioned storage.

I did not (really) test the selectors, but they appear to be in the right place now since they fail if they are unsupported.

csibbitt:
[Validate OCS as a storage backend for dynamic PVs backing STF and OCP](https://trello.com/c/ZW9kyr6D/1176-validate-ocs-as-a-storage-backend-for-dynamic-pvs-backing-stf-and-ocp)

deploy/olm-catalog/service-telemetry-operator/manifests/service-telemetry-operator.clusterserviceversion.yaml
leifmadsen: Yea I noticed that I don't get the change unless I run it locally in the branch. I must have added it post-1.1 tag :(

It also didn't show up when I had `--output-dir` but worked when it was in the local repo. Thanks for adding that!

leifmadsen (approved):

PR #151, opened 2020-11-30
Elasticsearch datasources for cloud dashboard

pleimer:


roles/servicetelemetry/templates/manifest_grafana_ds.j2
```diff
@@ -0,0 +1,51 @@
...
+      basicAuth: true
+      basicAuthUser: elastic
+      basicAuthPassword: {{ elasticsearch_pass }}
+      database: ceilometer_*
+      jsonData:
+        tlsSkipVerify: true
+        timeField: startsAt
+        esVersion: 70
+{% endif %}
+  name: {{ meta.name }}-ds-prometheus.yaml
```
leifmadsen, line 51: I'd suggest changing this to `{{ meta.name}}-ds-stf.yaml` or something?

roles/servicetelemetry/tasks/component_grafana.yml
```diff
@@ -16,17 +16,29 @@
 
 # Lookup existing datasources
 - block:
-  - name: Set default Prometheus datasource
+
+  - block:
+    - name: Retrieve elastic search secret
```
leifmadsen, line 21: I think this will fail when you don't have ElasticSearch deployed since the Secret won't exist?

leifmadsen: Disregard, I see the `when` now.

pleimer:
CI not working because of the rate limit on docker.io

pleimer:
I don't think there is much we can do besides change the base image of sg-bridge. The BuildConfig for sg-bridge will not look for the centos8 image in the internal registry and I don't see a way to change that. Or we can stick with spotty CI

leifmadsen:
@pleimer did I not migrate to using UBI8 for all the components? (maybe that PR is still open...)

leifmadsen:
> @pleimer did I not migrate to using UBI8 for all the components? (maybe that PR is still open...)

Huh! I didn't do that repo! I did some of the other ones though. Let me see if I can figure out how to migrate sg-bridge to UBI8, which would solve your docker.io issue.

pleimer:
that would be amazing

leifmadsen:
Looking at the logs super quickly, I think the next issue is that the ElasticSearch pod is probably trying to pull an image from docker.io which is failing and causing the smoketest to fail (since you're missing the events store).

pleimer:
Actually the pod was able to come up just fine, I think I found the issue though: `[1]: max virtual memory areas vm.max_map_count [65530] is too low, increase to at least [262144]`

Now I just need to remember how to adjust that

leifmadsen:
> Actually the pod was able to come up just fine, I think I found the issue though: `[1]: max virtual memory areas vm.max_map_count [65530] is too low, increase to at least [262144]`
> 
> Now I just need to remember how to adjust that

Ugh that's right. It was an issue in a recent CRC release (I think specifically 1.17.0 which is OCP 4.5.x I believe).

They said it would be fixed in a later release, but that probably won't help us so much due to the newer releases both being OCP 4.6 (which doesn't have AMQ Cert Manager...)

You're probably going to have to go back to CRC 1.16.0 or something, since they disabled the tuned operator in CRC 1.17 where that issue started to crop up.

leifmadsen:
https://github.com/code-ready/crc/issues/1590

pleimer:
I was able to adjust the value in the VM so, we will see!

/runtest

leifmadsen:
[Add ElasticSearch DataSource for Dashboards when events are enabled in STO](https://trello.com/c/IpBZQ6WE/1209-add-elasticsearch-datasource-for-dashboards-when-events-are-enabled-in-sto)

PR #152, opened 2020-12-02
Fix smoketest

pleimer:


leifmadsen (approved):

leifmadsen:
Certainly looks more correct. Not sure why CI isn't passing though.

pleimer:
the webhook is taking longer than the tests to fire

PR #154, opened 2020-12-04
Fix typo in parameter name

mrunge:
MetricPipelinePublishers is a typo, it should be PipelinePublishers as in 
https://github.com/openstack/tripleo-heat-templates/blob/master/environments/enable-legacy-telemetry.yaml#L27-L28
Otherwise 'gnocchi://?filter_project=service&archive_policy=high' doesn't land in /etc/ceilometer/pipeline.yaml
under publishers, only notifier (for sending data via amqp1).


leifmadsen (approved):
Guess I just guessed when I was creating this? Looked at the upstream templates for both master and train, and looks like it's always been "PipelinePublishers" for metrics.

leifmadsen:
@pleimer pretty sure the testing won't cover this change since it's in `infrared-openstack.sh`.

PR #159, opened 2021-01-07
Provide bundle under stable and stable-1.1

leifmadsen:
Set the STF 1.1 bundle to be available under both stable and stable-1.1 channels, but
change the default channel to stable-1.1. This will help with prep and migration
to versioned channels.

Unversioned channel stable would be deprecated, and no longer referenced in stable-1.2.


leifmadsen:
I'd like to merge these down for our upstream and start aligning things to versions channels. I know this is a bit of a change mid release and I should probably start this next release, but I kind of need to test all this out so I can get the downstream stuff sorted as well. I'm not sure we really have anyone consuming this upstream at the moment other than ourselves.

I'm also going to move the `master` branch to `stable-1.2` in preparation for the next release. Once I branch `master` to `stable-1.2` I'll do the same change but make `master` target `stable-1.3`. Or maybe I should call it something like `canary-1.3` until the branching occurs? (and I'd suggest making `master` be `canary-1.2` until I actually branch it?)

bundle.Dockerfile
```diff
@@ -4,8 +4,8 @@ LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
 LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=service-telemetry-operator
-LABEL operators.operatorframework.io.bundle.channels.v1=latest
-LABEL operators.operatorframework.io.bundle.channel.default.v1=stable
+LABEL operators.operatorframework.io.bundle.channels.v1=stable,stable-1.1
```
csibbitt, line 7: This PR targets the wrong branch. We should not be releasing the master branch as "stable"

leifmadsen: Resolved

csibbitt (approved):

leifmadsen:
[Create stable-1.1 channel and stable-1.2 channel against bundles](https://trello.com/c/2WNMUssd/1237-create-stable-11-channel-and-stable-12-channel-against-bundles)

PR #160, opened 2021-01-08
Set master branch to latest OLM channel

leifmadsen:
Set the master branch to the OLM channel 'latest'. Adjust the current
unreleased CSV version of 1.1.1 to 1.2.0 since we'll not need to release
another 1.1.x release.


csibbitt (approved):
LGTM

leifmadsen:
> LGTM

Thanks, still doing a bit of testing and making sure things are aligned. I woke up today and realized there is a possible thing that could cause this to not work, but that fix will likely be on the `stable-1.1` branch.

Been doing a bunch of research and asking for other folks patterns who have been doing this, and it seems like there is a still a lot of flux, unknowns, and snowflake procedures being used across different products.

leifmadsen:
OK I think this is all correct after additional investigation. I will create a similar change for smart-gateway-operator.

leifmadsen:
[Create stable-1.1 channel and stable-1.2 channel against bundles](https://trello.com/c/2WNMUssd/1237-create-stable-11-channel-and-stable-12-channel-against-bundles)

PR #165, opened 2021-01-21
Refactor sg-core

pleimer:
I am tracking changes to STO for the sg-core here, mostly just so that I can run the CI on the changes without having to build everything myself. Most likely this will not contain any changes.


pleimer:
I decided to put my updates to the performance test here

PR #170, opened 2021-02-03
Resolve issue with ElasticSearch password retrieval

leifmadsen:

Resolves: rhbz#1924195

### DIFF
--- a/roles/servicetelemetry/tasks/component_clouds.yml
+++ b/roles/servicetelemetry/tasks/component_clouds.yml
@@ -31,7 +31,7 @@
 
   - name: Filter out ElasticSearch password for BasicAuth
     set_fact:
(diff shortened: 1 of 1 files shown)


leifmadsen:
This change will need to be cherry-picked to stable-1.1 and stable-1.2 then imported downstream.

leifmadsen:
## Testing

I tested this originally with our default `build/Dockerfile` and everything was fine. At some point either a library in python or ansible seems to have gotten a bit pickier because I can't reproduce this with our default `build/Dockerfile`.

I reproduced the issue with the latest `registry.redhat.io/openshift4/ose-ansible-operator:v4.5`. Issue was something to do with a padding error.

Changing to not returning a list of values seems to have cleared all of this up. I iterated through this a few times with a bunch of changes, but ultimately passing only a single item list seems to have resolved the problem.

csibbitt (approved):

PR #171, opened 2021-02-03
Resolve issue with ElasticSearch password retrieval (#170)

leifmadsen:

Resolves: rhbz#1924195


mrunge:
The change looks okay, I'm am currently running  a fresh deploy including the change.

mrunge:
okay, elasticsearch keeps crashing for me, but it does without this as well.

mrunge (approved):

PR #172, opened 2021-02-03
Resolve issue with ElasticSearch password retrieval (#170)

leifmadsen:

Resolves: rhbz#1924195

### DIFF
--- a/roles/servicetelemetry/tasks/component_clouds.yml
+++ b/roles/servicetelemetry/tasks/component_clouds.yml
@@ -31,7 +31,7 @@
 
   - name: Filter out ElasticSearch password for BasicAuth
     set_fact:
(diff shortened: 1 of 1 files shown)


mrunge (approved):
the change itself looks pretty straightforward.

However, elasticsearch keeps crashing for me, I can not verify it is really working.

leifmadsen:
> the change itself looks pretty straightforward.
> 
> However, elasticsearch keeps crashing for me, I can not verify it is really working.

If you created a CRC environment with defaults (didn't pass `--memory` to the `crc start`) then you're likely running in to a lack of resources, which results in ES not being able to start.

mrunge:
I created an env with 30 Gigs of memory. That was sufficient until last time I looked.

Issue #175, opened 2021-03-01
default-cloud1-coll-meter-smartgateway-.... keeps crashing


mrunge:
With STF 1.2 it seems there is a changed/removed option in sg-core, which is still being used in the container?
 
```
Usage: /sg-core [OPTIONS]

Available options:
flag provided but not defined: -usetimestamp
  -config string
    	configuration file path (default "/etc/sg-core.conf.yaml")
  -cpuprofile string
    	write cpu profile to file


Default configurations:

plugindir: /usr/lib64/sg-core/
loglevel: info
handlererrors: false
transports: []
applications: []
```

Looking into the smart gateway operator, the deployment for `default-cloud1-coll-meter-smartgateway`
lists
```
       - name: smart-gateway
          image: 'quay.io/infrawatch/sg-core:latest'
          args:
            - '-usetimestamp'
            - '-promhost'
            - 0.0.0.0
            - unix
            ```

mrunge:
This is probably fixed by https://github.com/infrawatch/smart-gateway-operator/pull/72/files#diff-6957ee03d95e8f0dd9b1adec826c31595a6a9e512e435375e0a085adbf44d7f1L48

leifmadsen:
Those changes to sg-core are targeting master and should not be in STF 1.2.

pleimer:
The usage you showed is for the sg-core latest, that isn't in the stable-1.2 branch

mrunge:
Yes, the output above is from unstable branch. I was somehow hoping to get to a working deployment. Neither (1.2) nor unstable worked. This is why I've opened the bug.

pleimer:
Unstable builds are now all in alignment. This should no longer be an issue, unless there is a separate bug for 1.2 but the description you gave for this only relates to unstable. Closing for now unless there are issues again

PR #176, opened 2021-03-01
Pull stable operators from redhat-operators

mrunge:
and also stay on stable-1.2 channel

### DIFF
--- a/build/stf-run-ci/tasks/setup_stf.yml
+++ b/build/stf-run-ci/tasks/setup_stf.yml
@@ -29,7 +29,7 @@
         channel: stable-1.2
         installPlanApproval: Automatic
         name: smart-gateway-operator
-        source: redhat-operators
+        source: infrawatch-operators
         sourceNamespace: openshift-marketplace
   when: smart_gateway_operator_subscription_manifest is not defined
 
@@ -45,7 +45,7 @@
         channel: stable-1.2
         installPlanApproval: Automatic
         name: service-telemetry-operator
-        source: redhat-operators
+        source: infrawatch-operators
         sourceNamespace: openshift-marketplace
   when: service_telemetry_operator_subscription_manifest is not defined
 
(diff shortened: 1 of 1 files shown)


mrunge:
Do you want me to drop the infrawatch-operators CatalogSource as well?

leifmadsen:
I'm not entirely sure how I feel about this change. Changing our upstream scripts to refer to the downstream builds and releases I'm not sure makes sense here.

Personally I think I would rather see our scripts be using the latest changes from the branch, or install via the CSV directly or something.

mrunge:
Okay, 1.2 is not installable via infrawatch-operators, and unstable does not install, see https://github.com/infrawatch/service-telemetry-operator/issues/175

pleimer:
@mrunge unstable will fail until https://github.com/infrawatch/smart-gateway-operator/pull/72 merges

leifmadsen (changes requested):
Thinking about this more, I'm going to reject this change. The script is intended to load STF from the upstream sources for testing and development, and pulling from the downstream released version doesn't help accomplish that.

If you need a released/stable version of STF, then you should install via the official documentation and consume it via the redhat-operators catalog. If the issue is that things are not working, then we should resolve the problem.

mrunge:
How are you then testing stable upstream changes, if 1.2 from upstream does not install, and unstable fails to install as well?

leifmadsen:
> How are you then testing stable upstream changes, if 1.2 from upstream does not install, and unstable fails to install as well?

Installing from CSV directly I guess. I am not testing it currently, but that doesn't mean it shouldn't be fixed. My focus has been on downstream release and release management along with being injured. It'll be fixed when it's fixed, but just circumventing the upstream to pull the downstream is not a solution here.

mrunge:
Yeah, I would prefer using upstream first.

leifmadsen:
> Yeah, I would prefer using upstream first.

Sure that makes sense, but sometimes it'll be in a broken state, and will need to be rescued. It is what it is. Anyways, I'm going to close this for now. I'd suggest just importing the CSVs directly or using the downstream release for now until I can get time to do another upstream release.

mrunge:
broken state is okay for unstable, but not for stable branches, imho.

leifmadsen:
I agree, but I can only do so much. Interested in creating new bundles and index images?

PR #181, opened 2021-03-24
Align upstream and downstream base images

leifmadsen:
Better align the upstream and downstream base images so that we have a
better idea that changes we make upstream will work without issue downstream


leifmadsen:
I'm going to fix the linting here the same as I did in master earlier just to keep this from complaining.

pleimer (approved):

PR #186, opened 2021-03-29
[WIP] Additional service monitors to monitor STF itself

csibbitt:


roles/servicetelemetry/templates/manifest_servicemonitors.j2
```diff
@@ -0,0 +1,57 @@
...
+          sourcelabels: []
+        - action: labeldrop
+          regex: namespace
+          sourcelabels: []
+        - action: labeldrop
+          regex: instance
+          sourcelabels: []
+        - action: labeldrop
+          regex: job
+          sourcelabels: []
```
leifmadsen, line 23: These are from a day before I worked on the project. I wonder if we should just try dropping these and building a new set of `metricRelabelings` as we determine? We might end up with the same thing, but at least we'll have worked through why they are here.

leifmadsen (commented):
Looks good so far

roles/servicetelemetry/templates/manifest_servicemonitors.j2
csibbitt: Easy enough to hack these out of our live installation and see if we find those labels valuable on not

PR #187, opened 2021-03-30
Only need 15 OSDs in test configuration

leifmadsen:
For our default test environment of 3 ceph nodes, there are only 15 OSDs configured. We have
16 configured as we start counting at 0. Dropping the ceph-osd.15 list item.


csibbitt (approved):
I'll take your word for it

leifmadsen:
> I'll take your word for it

I also did a visual check across the 3 controllers, and no `15` value was present.

PR #188, opened 2021-03-31
[WIP] hackfest servicemonitors and recordingrules

csibbitt:
This is incomplete but useful.

### DIFF
--- a/roles/servicetelemetry/templates/manifest_recordingrules.j2
+++ b/roles/servicetelemetry/templates/manifest_recordingrules.j2
@@ -1,0 +1,86 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: stf-default
+    role: alert-rules
+  name: stf-data-model-rules
+  namespace: service-telemetry
+spec:
+  groups:
+  - name: stf-data-model-base-fs-util
+    rules:
+    - record: stf_fs_util_pcnt_inodes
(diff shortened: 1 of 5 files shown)


roles/servicetelemetry/templates/manifest_recordingrules.j2
```diff
@@ -0,0 +1,86 @@
...
+    - record: stf_sg_util_ceilometer_msgs_total
+      expr: sg_total_ceilometer_msg_received_count
+    - record: stf_sg_util_collectd_metrics_total
+      expr: sg_total_collectd_metric_decode_count
+    - record: stf_sg_util_collectd_msgs_total
+      expr: sg_total_collectd_msg_received_count
+    - record: stf_sg_util_all_metrics_total
+      expr: sg_total_ceilometer_metric_decode_count + sg_total_collectd_metric_decode_count
+    - record: stf_sg_util_all_msgs_total
+      expr: stf_sg_util_ceilometer_msgs_total + sg_total_collectd_msg_received_count
```
leifmadsen, line 78: I can't seem to get these ones to work. I tried various combinations, but it doesn't seem to like doing addition on them. Just returns an empty set.

Also I would have expected to see the expr being something like `stf_sg_util_collectd_msgs_total + stf_sg_util_ceilometer_msgs_total` but none of the combinations of `stf` or `sg` metrics seems to work.

If I wrap the `stf` rules in `sum()` and then `sum(...) + sum(...)` that seems to work, but not sure if this is exactly what we're going for here.

csibbitt: I'll take a closer look at this one, thanks.

csibbitt:
An idea we might explore again in the future, but not currently a priority

PR #189, opened 2021-04-02
Added a more complete prometheus alert rules file

josecastillolema:
In addition to the existing  prometheus alert rules (`alerts.yaml`) we suggest adding a new one (`alerts_extended.yaml`) with prometheus rules for:
 - ceph
 - virtual layer
 - availability

### DIFF
--- a/deploy/alerts/README.md
+++ b/deploy/alerts/README.md
@@ -6,16 +6,16 @@ amount of standard deviation). They provide a default and may
 require adjustment for specific deployments.
 
 ## alerts_extended.yaml
-Besides `alerts.yaml` alarms from openstack.rules, it includes:
- - ceph.rules: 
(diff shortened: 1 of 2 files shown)


csibbitt:
Thanks very much for this, Joe! We're working on some updates to our internal metrics model + dashboards and alerts in upcoming sprints, so qw will work this in to what we're doing where appropriate.

csibbitt:
It's been a long time since this was submitted, and from a quick glance it still looks valuable. We never did the overhaul of the metrics model, so this may even work as-is.

Testing and documenting this might be a good exercise for new team members.

leifmadsen:
Based on the current lifecycle for STF, I don't believe this will get merged and reviewed going forward. I'm going to close this out as part of overall clean up, but we do appreciate the contribution, and I am sorry it didn't get any attention.

PR #190, opened 2021-04-06
Install Ansible for working CI

leifmadsen:


leifmadsen:
For some reason installing `ansible-lint` alone is not enough. The linting test would fail with a complaint about missing Ansible:

```
ERROR    No module named 'ansible'
FATAL: ansible-lint requires a version of Ansible package >= 2.9, but none was found. Please install a compatible version using the same python interpreter. See https://docs.ansible.com/ansible/latest/installation_guide/intro_installation.html#installing-ansible-with-pip
Error: Process completed with exit code 4.
```

This change results in that error message no longer failing the CI.

PR #193, opened 2021-04-15
Make generating CSVs possible via templated CSV

leifmadsen:
Adjust the deploy/ directory contents so that generating a CSV is
possible for release automation. Makes use of a UNIX timestamp to
provide the final version position after defining the major release
version of the CSV.


leifmadsen:
OK I think I'm happy enough with this now that it could be merged. Eventually I think I want to get this actually building and pushing the resulting bundle images to quay.io, and then updating the index image for us as well. Might need some tweaking for downstream and such as well, variable names, all that good stuff, but at least we have a way of reproducing and templating the bundle contents now.

pleimer (approved):

PR #207, opened 2021-06-03
Adjust grafana datasource name to be more general

pleimer:
Previous naming of the grafandatasource implied only a prometheus datasource was controlled by the object when in fact a couple of elasticsearch datasources were too. This made documentation misleading

### DIFF
--- a/roles/servicetelemetry/tasks/component_grafana.yml
+++ b/roles/servicetelemetry/tasks/component_grafana.yml
@@ -35,7 +35,7 @@
     k8s:
       api_version: integreatly.org/v1alpha1
       name: '{{ meta.name }}-ds-datasource'
-      kind: GrafanaDatasource
+      kind: GrafanaDataSource
       namespace: '{{ meta.namespace }}'
       state: absent
 
--- a/roles/servicetelemetry/tasks/component_grafana.yml
+++ b/roles/servicetelemetry/tasks/component_grafana.yml
@@ -34,7 +34,7 @@
   - name: Remove legacy datasources
     k8s:
       api_version: integreatly.org/v1alpha1
-      name: '{{ meta.name }}-ds-datasource'
+      name: '{{ meta.name }}-ds-prometheus'
       kind: GrafanaDataSource
       namespace: '{{ meta.namespace }}'
       state: absent
--- a/roles/servicetelemetry/tasks/component_grafana.yml
+++ b/roles/servicetelemetry/tasks/component_grafana.yml
@@ -34,7 +34,7 @@
   - name: Remove legacy datasources
     k8s:
       api_version: integreatly.org/v1alpha1
-      name: '{{meta.name}}-ds-datasource'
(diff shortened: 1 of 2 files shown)


roles/servicetelemetry/templates/manifest_grafana_ds.j2
```diff
@@ -1,7 +1,7 @@
 apiVersion: integreatly.org/v1alpha1
 kind: GrafanaDataSource
 metadata:
-  name: {{ meta.name }}-ds-prometheus
+  name: {{ meta.name }}-datasources
```
leifmadsen, line 4: Will the dashboards will need to be adjusted for this? This is just a local naming convention in Grafana right?

leifmadsen: Never mind I looked through the code. This looks good.

leifmadsen (commented):
When this is approved, can you cherry pick this against `stable-1.2` and `stable-1.3` as well?

Also just got thinking here... what happens when we've already defined the datasource previously, and we change this? I think we're going to find we have the same data sources created and listed twice, or one of these is going to fail. At the very least I think we'll have something that is stale.

You're probably going to need to add a look up that looks for the existing DS and then remove it (`state: absent`) when the `{{ meta.name}}-ds-datasource` is present. Make the change, cherry pick this to `stable-1.2` and `stable-1.3`, and then create another patch that removes the check from `master` (as we won't need it for `stable-1.4`).

roles/servicetelemetry/templates/manifest_grafana_ds.j2
pleimer: this won't affect the dashboards at all. This change is purely to make the documentation less confusing

roles/servicetelemetry/tasks/component_grafana.yml
```diff
@@ -23,14 +23,22 @@
...
 
     - name: Decode elasticsearch password
       set_fact:
         elasticsearch_pass: '{{ es_secret.resources[0].data.elastic | b64decode }}'
     when: servicetelemetry_vars.backends.events.elasticsearch.enabled
 
+  - name: Remove legacy datasources
+    k8s:
+      api_version: integreatly.org/v1alpha1
+      name: '{{meta.name}}-ds-datasource'
```
leifmadsen, line 37:
```suggestion
      name: '{{ meta.name }}-ds-datasource'
```

roles/servicetelemetry/tasks/component_grafana.yml
```diff
@@ -23,14 +23,22 @@
...
+      kind: grafanadatasource
```
leifmadsen, line 38:
```suggestion
      kind: GrafanaDatasource
```

pleimer:
This has been tested and the deletion of the previous datasource works as expected

mrunge (approved):

PR #224, opened 2021-07-01
Deploy a sensubility smartgateway

pleimer:
A sensubility smartgateway is configured to listen to container health events from sensubility at a separate QDR  address from all other smartgateways and store the results as metrics into Prometheus. Resolves: rhbz#1956912

Depends-on: https://github.com/infrawatch/sg-core/pull/60

### DIFF
--- a/roles/servicetelemetry/defaults/main.yml
+++ b/roles/servicetelemetry/defaults/main.yml
@@ -73,7 +73,7 @@ servicetelemetry_defaults:
             subscription_address: anycast/ceilometer/metering.sample
             debug_enabled: false
           - collector_type: sensubility
-            subscription_address: sensubility/metrics
+            subscription_address: sensubility/telemetry
             debug_enabled: false
       events:
         collectors:
--- a/roles/servicetelemetry/defaults/main.yml
+++ b/roles/servicetelemetry/defaults/main.yml
@@ -72,6 +72,9 @@ servicetelemetry_defaults:
           - collector_type: ceilometer
             subscription_address: anycast/ceilometer/metering.sample
             debug_enabled: false
+          - collector_type: sensubility
+            subscription_address: sensubility/metrics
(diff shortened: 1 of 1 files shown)


roles/servicetelemetry/defaults/main.yml
```diff
@@ -72,6 +72,9 @@ servicetelemetry_defaults:
           - collector_type: ceilometer
             subscription_address: anycast/ceilometer/metering.sample
             debug_enabled: false
+          - collector_type: sensubility
+            subscription_address: sensubility/metrics
```
pleimer, line 76: I haven't fully decided what this address should be yet, any suggestions?

leifmadsen: Name seems fine to me.

leifmadsen:
Seem to be missing the logic to handle this new type. I underestimated the amount of changes required to support this new feature, as I thought it might just show up in the existing Smart Gateways that were deployed. If we're going to be creating a whole new SG type for this, then I'm not sure this is the correct approach.

Perhaps we should be focused on getting the data set up in sg-core to better match what we already had rather than migrating to a new approach in such a short time frame.

pleimer:
Note: I will be adding smoke test updates to a different PR

csibbitt (approved):

roles/servicetelemetry/defaults/main.yml
leifmadsen: Actually I think I'd like this to align to something we already have a bit. I'd suggest `sensubility/telemetry` to align to `collectd/telemetry`.

leifmadsen, line 76:
```suggestion
            subscription_address: sensubility/telemetry
```

leifmadsen:
Merging as dependency in sg-core has been merged.

PR #236, opened 2021-07-12
Enable Ceilometer events

leifmadsen:
The upstream ceilometer-write-qdr for Train is not in sync with the contents
deployed downstream for ceilometer-write-qdr.yaml. The upstream implements the
Publish parameters, but the current 16.1 deployment does not provide those.
Added appropriate configuration required for RHOSP16.1.

### DIFF
--- a/tests/infrared/16/enable-stf.yaml.template
+++ b/tests/infrared/16/enable-stf.yaml.template
@@ -20,9 +20,9 @@ custom_templates:
         ManagePolling: true
         ManagePipeline: true
 
-        # required to set valid parameter due to typo in ceilometer-write-qdr.yaml
-        # and will be resolved in a future release
+        # enable Ceilometer metrics and events
         CeilometerQdrPublishMetrics: true
+        CeilometerQdrPublishEvents: true
 
         # enable collection of API status
         CollectdEnableSensubility: true


leifmadsen:
Testing is currently in progress. Do not merge, but I expect this will work.

tests/infrared/16/enable-stf.yaml.template
```diff
@@ -20,9 +20,9 @@ custom_templates:
         ManagePolling: true
         ManagePipeline: true
 
-        # required to set valid parameter due to typo in ceilometer-write-qdr.yaml
-        # and will be resolved in a future release
+        # enable Ceilometer metrics and events
         CeilometerQdrPublishMetrics: true
+        CeilometerQdrPublishEvents: true
```
leifmadsen, line 26: Just an FYI that this should go away in the 16.2 timeframe since the changes upstream should be pulled in during full import. It's required for 16.1 because the ceilometer-write-qdr.yaml template is not in sync with upstream Train (missing these parameters).

csibbitt (approved):

leifmadsen:
Testing is completed and the expected configuration is available:

```yaml
sinks:
    - name: event_sink
      transformers:
      triggers:
      publishers:
          - gnocchi://?filter_project=service
          - notifier://172.17.1.11:5666/?driver=amqp&topic=cloudops04-event
```

leifmadsen:
I can confirm events are now coming in with this change.

PR #246, opened 2021-08-17
Use pre-configured pod as main agent in functional CI

pleimer:
In order to limit the number of jobs that can run at once, the Jenkins
job must use a predefined agent rather than defining its own.


pleimer:
test

PR #247, opened 2021-08-17
Use pre-configured pod as main agent in functional CI (#246)

pleimer:
In order to limit the number of jobs that can run at once, the Jenkins
job must use a predefined agent rather than defining its own.

(cherry-picked from commit id 7e072285540e11b5ce74a9420b5be30a232a0f96)


pleimer:
test

PR #248, opened 2021-08-20
Differentiate default component versions for builds in stf-run-ci

pleimer:
Now that we maintain more than one STF version, the CI role must
differentiate between componenet versions when building STF
subcomponents rather than defaulting to the master branch.
Considering not all components always follow the
"stable-<stf-version>" convention for STF releases (sg-bridge &
prometheus-webhook-snmp for STF-1.3), it is easiest to map
components to their stable branch according to the STF version in the
default vars. This map will be adjusted according to the target STF stable
branch when backported

### DIFF
--- a/build/stf-run-ci/defaults/main.yml
+++ b/build/stf-run-ci/defaults/main.yml
@@ -21,3 +21,10 @@ sg_core_image_tag: latest
 sg_bridge_image_tag: latest
 prometheus_webhook_snmp_image_tag: latest
 namespace: service-telemetry
+
+# used when building images to default to correct version branch for STF subcomponents per STF version
+version_branches:
+  sgo: master
+  sg_core: master
+  sg_bridge: master
+  prometheus_webhook_snmp: master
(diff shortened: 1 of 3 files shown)


build/stf-run-ci/defaults/main.yml
```diff
@@ -21,3 +21,10 @@ sg_core_image_tag: latest
 sg_bridge_image_tag: latest
 prometheus_webhook_snmp_image_tag: latest
 namespace: service-telemetry
+
+# used when building images to default to correct version branch for STF subcomponents per STF version
+version_branches:
```
csibbitt, line 26: Just to check my understanding - would this file will be adjusted on a stable-* branch so  that it mentions the corresponding branches in the other repos?

csibbitt (approved):

build/stf-run-ci/defaults/main.yml
csibbitt: I just read PR#249 and yes, it seems I understood correctly :)

PR #249, opened 2021-08-20
Differentiate default component versions for builds in stf-run-ci

pleimer:
Now that we maintain more than one STF version, the CI role must
differentiate between componenet versions when building STF
subcomponents rather than defaulting to the master branch.
Considering not all components always follow the
"stable-<stf-version>" convention for STF releases (sg-bridge &
prometheus-webhook-snmp for STF-1.3), it is easiest to map
components to their stable branch according to the STF version in the
default vars. This map will be adjusted according to the target STF stable
branch when backported

(cherry-picked from commit ID f0562feb0766f5057511df73521e96cbc47babb0)

### DIFF
--- a/build/stf-run-ci/defaults/main.yml
+++ b/build/stf-run-ci/defaults/main.yml
@@ -21,3 +21,11 @@ sg_core_image_tag: latest
 sg_bridge_image_tag: latest
 prometheus_webhook_snmp_image_tag: latest
 namespace: service-telemetry
+
+# used when building images to default to correct version branch for STF subcomponents per STF version
+version_branches:
+  sgo: stable-1.3
+  sg_core: stable-1.3
+  sg_bridge: stable-1.2
+  prometheus_webhook_snmp: stable-1.2
+  
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -76,12 +76,17 @@ spec:
                   name: elasticsearch-data
 """
 
+def working_branch = "stable-1.3"
+
 node('ocp-agent') {
     container('exec') {
         dir('service-telemetry-operator') {
             stage ('Clone Upstream') {
                 catchError(buildResult: 'FAILURE', stageResult: 'FAILURE') {
                     checkout scm
+                    working_branch = sh(script: 'git ls-remote --heads origin | grep $(git rev-parse HEAD) | cut -d / -f 3', returnStdout: true).toString().trim()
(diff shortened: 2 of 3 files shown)


pleimer:
test

build/stf-run-ci/defaults/main.yml
```diff
@@ -21,3 +21,11 @@ sg_core_image_tag: latest
 sg_bridge_image_tag: latest
 prometheus_webhook_snmp_image_tag: latest
 namespace: service-telemetry
+
+# used when building images to default to correct version branch for STF subcomponents per STF version
+version_branches:
+  sgo: stable-1.3
+  sg_core: stable-1.3
+  sg_bridge: stable-1.2
+  prometheus_webhook_snmp: stable-1.2
```
csibbitt, line 30: Hmm, the comment said "Considering not all components always follow the 'stable-' convention..." but it looks like they DO follow that convention. It seems the reason this code needs to exist (right now) is simply that we neglected to create stable-1.3 branches in some of the dependencies.  I still like this concept and think the implementation is valuable, I just feel slightly misled ;)

csibbitt (approved):

build/stf-run-ci/defaults/main.yml
leifmadsen: Yea since prometheus-webhook-snmp (and sg-bridge) didn't change for the STF 1.3 release, corresponding branches were not created. Ultimately the branch naming is probably a bit of a mislead overall since I think the intention was that we'd have per y-stream release component branches, but it appears that's not always the case.

In our downstream productization stream for example, we have an `stf-common` pipeline that takes in those common release components across STF 1.2 and 1.3, which relates to the `stable-1.2` branches currently upstream.

We could look at adjusting this, but basically some components are used across multiple product releases under a single version. I'm not exactly sure how to resolve this other than to duplicate the branching to be consistent, but that feels more confusing and managed overhead than necessary.

For the logging work for example, we'd in theory have a `stable-1.2` and `stable-1.4` branch for `sg-bridge` with no `stable-1.3` release. Or maybe it'd become `stable-2.0` and be used as part of STF 1.4.

pleimer: Yeah I think markdown removed the text after "stable-" in my comment, but it was supposed to indicate that the stable version for the sub component may not always match the STF version. Sorry for the confusion

PR #239, opened 2021-07-29
Add Loki-operator to STO

vyzigold:
This PR adds possibility to deploy STO with loki-operator and deploy Loki and log SG with it.

A few basic configuration options for the loki-operator are exposed through STO configuration. More options can be easily added later when needed.

Build and deployment is temporarily added to the stf-run-ci ansible role for easier development. This won't be needed once the operator is released. The loki-operator is currently under development, so I made a few workarounds to make some things work.
The main workarounds are:

- Remove authentication from /metrics - this seemed to prevent loki microservices from communicating with each other
- Remove operator-sdk build - I wasn't able to build operator-sdk, so I'm downloading a prebuilt binary instead.
- Skip tls verification - There doesn't seem to be an option to verify OCS certificate yet, so the config file is patched to skip it.

I added examples for deploying with minio to crc and with OCS to something like PSI to the docs.

PR #252, opened 2021-08-30
Add Loki-operator to STO

vyzigold:
This PR adds possibility to deploy STO with loki-operator and deploy Loki and log SG with it.

A few basic configuration options for the loki-operator are exposed through STO configuration. More options can be easily added later when needed.

Build and deployment is temporarily added to the stf-run-ci ansible role for easier development. This won't be needed once the operator is released. The loki-operator is currently under development, so I made a few workarounds to make some things work.
The main workarounds are:

    Remove authentication from /metrics - this seemed to prevent loki microservices from communicating with each other
    Remove operator-sdk build - I wasn't able to build operator-sdk, so I'm downloading a prebuilt binary instead.
    Skip tls verification - There doesn't seem to be an option to verify OCS certificate yet, so the config file is patched to skip it.

I added examples for deploying with minio to crc and with OCS to something like PSI to the docs.

### DIFF
--- a/deploy/crds/infra.watch_v1beta1_servicetelemetry_cr.yaml
+++ b/deploy/crds/infra.watch_v1beta1_servicetelemetry_cr.yaml
@@ -34,6 +34,15 @@ spec:
           persistent:
             storageSelector: {}
             pvcStorageRequest: 20Gi
+
+    logs:
+      loki:
+        enabled: false
+        flavor: 1.extra-small
+        replicationFactor: 1
+        storage:
+          objectStorageSecret: test
+          storageClass: standard
   graphing:
     enabled: false
     grafana:
--- a/build/create_standard_pvs.sh
+++ b/build/create_standard_pvs.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+
+# Create the standard storage class
+if ! oc describe sc standard; then
+	oc create -f - <<EOF
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: standard
+provisioner: kubernetes.io/no-provisioner
+reclaimPolicy: Delete
+volumeBindingMode: WaitForFirstConsumer
+EOF
+fi
+
+# Check how many PVs we need to patch
+AVAILABLE_STANDARD_PVS=`oc get pv | grep standard | grep Available | wc -l`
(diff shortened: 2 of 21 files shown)


vyzigold:
This code is the same as the code in #239, but the PR is made from this repository so that the CI works correctly. The differences between #239 and this are:

- This PR is rebased on top of master
- A small merge conflict in stf-run-ci/tasks/clone_repos.yml was resolved
- The last commit, which brings stf-run-ci/tasks/clone_repos.yml up to date with current master

leifmadsen:
I'll do one last pass through this, but I'm sure it's fine :)

deploy/crds/infra.watch_v1beta1_servicetelemetry_cr.yaml
```diff
@@ -34,6 +34,15 @@ spec:
...
             storageSelector: {}
             pvcStorageRequest: 20Gi
+
+    logs:
+      loki:
+        enabled: false
+        flavor: 1.extra-small
+        replicationFactor: 1
+        storage:
+          objectStorageSecret: test
```
mrunge, line 44: can we please add some more obvious secret, rather than `test`?

leifmadsen: What do you suggest? You can make suggested edits on the web interface.

vyzigold: "test" is a name of secret, that comes with the loki-operator and it's preconfigured to work with minio. I guess I could use the stf-run-ci to replace the name of that secret for something else. Does something like "loki-storage-secret" sound OK?

leifmadsen: Is that the standard / default value that loki-operator uses? If so, then I think this is fine as-is. If it's trivial to name it anything you want, then loki-storage-secret sounds find to me. Thanks!

leifmadsen:
I'm merging this because I'm working on CRD migration stuff for OCP 4.9, and I don't want to conflict. I'll merge these changes in to my current working branch and deal with the conflicts there rather than you having to catch this branch up after the fact.

For the change to the secret name that can be dealt with as a separate PR. Thanks for all the hard work on this!

PR #253, opened 2021-08-30
Migrate to CRD API version v1

leifmadsen:
Migrate the CustomResourceDefinition API interface from v1beta1 to v1 in preparation for Kubernetes 1.22 where the v1beta1 API has been deprecated and removed.

### DIFF
--- a/deploy/olm-catalog/service-telemetry-operator/manifests/infra.watch_servicetelemetrys_crd.yaml
+++ b/deploy/olm-catalog/service-telemetry-operator/manifests/infra.watch_servicetelemetrys_crd.yaml
@@ -4,6 +4,8 @@ metadata:
   creationTimestamp: null
   name: servicetelemetrys.infra.watch
 spec:
+  conversion:
+    strategy: None
   group: infra.watch
   names:
     kind: ServiceTelemetry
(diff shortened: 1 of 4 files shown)


pleimer (approved):

deploy/olm-catalog/service-telemetry-operator/manifests/infra.watch_servicetelemetrys_crd.yaml
```diff
@@ -13,259 +15,316 @@ spec:
...
+                type: object
+            type: object
+        type: object
     served: true
     storage: true
+    subresources:
+      status: {}
 status:
   acceptedNames:
     kind: ""
```
csibbitt, line 330: Reviewed these side-by side as whole files and it made way more sense.

csibbitt (approved):
Do we have any CI somewhere that does upgrade tests we could get on board with?

leifmadsen:
> Do we have any CI somewhere that does upgrade tests we could get on board with?

An upgrade CI would be really good, especially for this. I want to try and coordinate this with changes along with SGO as I expect the SGO changes to look pretty small.

I will definitely have to test this going from one API version to another though. Our STF 1.2 to 1.3 upgrades were pretty bad from a customer perspective (I learned a bit too late in to the game about how to better handle transitioning / deprecating our own infra.watch API versions, which we could put in to STF 1.3 but I suspect the number of people who will do an upgrade is relatively small who haven't already).

As far as I can tell this should be a transparent migration, but I've been fooled before.

leifmadsen:
recheck

pleimer (approved):
Wow that's odd, Github was showing me a lot more changes than are actually included here and I was commenting on those non-existent changes. Ignore those comment

leifmadsen:
> Wow that's odd, Github was showing me a lot more changes than are actually included here and I was commenting on those non-existent changes. Ignore those comment

Yea those changes were already merged to `master` last week :) You probably see them here a bit because of the merge / backport to stable-1.3 so that CI all works again.

leifmadsen:
There is a corresponding PR at https://github.com/infrawatch/smart-gateway-operator/pull/99 with a same-named branch so I'm re-running the functional CI with both sets of changes operational.

leifmadsen:
recheck

leifmadsen:
I'm dropping the backport labels from this as I think we'll have an STF 1.4 soon enough to not need to backport this.

leifmadsen:
Sorry, last set of changes, reset the review one last time.

csibbitt (approved):

pleimer (approved):
such a small change but it looks so large in the Github diff

PR #254, opened 2021-08-31
Add baseImage parameter to Grafana parameter

leifmadsen:
Add the baseImage parameter to the graphing.grafana spec parameters of
the ServiceTelemetry object. Defaults to docker.io/grafana/grafana:8.1.2
but can be overridden at deployment time.

Resolves: rhbz#1998551

### DIFF
--- a/deploy/crds/infra.watch_servicetelemetrys_crd.yaml
+++ b/deploy/crds/infra.watch_servicetelemetrys_crd.yaml
@@ -258,6 +258,9 @@ spec:
                 adminUser:
                   description: Grafana admin user
                   type: string
+                baseImage:
+                  description: Path to the base container image used to instantiate a Grafana instance
+                  type: string
               type: object
           type: object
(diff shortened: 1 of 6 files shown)


pleimer (approved):
not tested but looks good to me

leifmadsen:
Currently blocked on some other work, but will get back to this hopefully today and provide a test report.

leifmadsen:
recheck

leifmadsen:
Just doing some final testing on this and making sure the values populate and display the way I want in the UI. Will then double check that it actually works :)

leifmadsen:
Looks good. Just need to verify everything and this is ready.
image

leifmadsen:
Verified to work. Was going to hack on stf-run-ci as well in this patch to get Grafana Operator and Grafana set up, but that feels like an overreach for what I'm doing here. Will definitely circle back soon and add Grafana Operator to the testing system though.

Issue #255, opened 2021-08-31
Installation of loki-operator blocked by Docker Hub rate limit


leifmadsen:
When deploying via `stf-run-ci` the builds of loki-operator results in a pull of `golang` image, which is hosted on hub.docker.com. If the deployment of this `run-ci.yaml` is run behind a corporate firewall, there is an extremely likely chance the deployment will fail.

The deployment of loki-operator needs to be wrapped so it can skipped, or a path parameter added that allows a different base image for loki-operator to be specified and in-line replaced in the build/working/loki-operator/Dockerfile.

This is a high priority issue as this will quickly block our ability to run CI.

leifmadsen:
```
Pulling image golang:1.16 ...
Warning: Pull failed, retrying in 5s ...
Warning: Pull failed, retrying in 5s ...
Warning: Pull failed, retrying in 5s ...
error: build error: failed to pull image: After retrying 2...g and upgrading: https://www.docker.com/increase-rate-limit
```

PR #256, opened 2021-08-31
Replace golang image in loki-operator

vyzigold:
I added a replace to the stf-run-ci, which replaces the golang base image used to build the loki-operator for the same image hosted on quay.io instead of docker.io. This should make the loki-operator buildable behind Red Hat firewall and it should fix the #255 issue.

### DIFF
--- a/.jenkins/agent/Dockerfile
+++ b/.jenkins/agent/Dockerfile
@@ -1,7 +1,7 @@
 FROM quay.io/openshift/origin-jenkins-agent-base:latest
-
-COPY install-sdk.sh /install-sdk.sh
-RUN /install-sdk.sh
-RUN dnf install -y ansible && \
+RUN curl -LO "https://github.com/operator-framework/operator-sdk/releases/download/v0.19.4/operator-sdk-v0.19.4-x86_64-linux-gnu" && \
+    chmod +x operator-sdk-v0.19.4-x86_64-linux-gnu && mv operator-sdk-v0.19.4-x86_64-linux-gnu /usr/local/bin/operator-sdk
+RUN dnf install -y ansible golang && \
+    dnf groupinstall -y "Development Tools" -y && \
     ansible-galaxy collection install community.kubernetes && \
     python -m pip install openshift
--- a/build/stf-run-ci/tasks/clone_repos.yml
+++ b/build/stf-run-ci/tasks/clone_repos.yml
@@ -69,9 +69,11 @@
         repo: https://github.com/viaq/loki-operator
         dest: working/loki-operator
         version: "{{ loki_operator_branch | default(branch, true) }}"
+        force: yes
   rescue:
     - name: "Get {{ version_branches.loki_operator }} branch because same-named doesn't exist"
       git:
         repo: https://github.com/viaq/loki-operator
         dest: working/loki-operator
         version: "{{ version_branches.loki_operator }}"
+        force: yes
--- a/build/stf-run-ci/tasks/main.yml
+++ b/build/stf-run-ci/tasks/main.yml
@@ -81,6 +81,12 @@
       regexp: "golang:1.16"
       replace: "{{ __golang_image_path }}"
 
+  - name: Replace Loki image
+    replace:
+      path: "{{ playbook_dir }}/working/loki-operator/internal/manifests/var.go"
+      regexp: "docker.io/grafana/loki:\\d\\.\\d\\.\\d"
+      replace: "{{ __loki_image_path }}"
+
   - name: Create builds and artifacts
     include_tasks: create_builds.yml
     loop:
--- a/build/stf-run-ci/tasks/main.yml
+++ b/build/stf-run-ci/tasks/main.yml
@@ -79,7 +79,7 @@
     replace:
       path: "{{ playbook_dir }}/working/loki-operator/Dockerfile"
       regexp: "golang:1.16"
-      replace: "quay.io/jwysogla/golang:latest"
+      replace: "{{ __golang_image_path }}"
 
   - name: Create builds and artifacts
     include_tasks: create_builds.yml
--- a/build/stf-run-ci/tasks/main.yml
+++ b/build/stf-run-ci/tasks/main.yml
@@ -49,6 +49,9 @@
   - name: Get new operator sdk
     command: ./get_new_operator_sdk.sh "{{ new_operator_sdk_version }}"
 
+  - name: Get new go
+    command: ./get_go.sh "{{ new_go_version }}"
+
   - name: Setup supporting Operator subscriptions
     include_tasks: setup_base.yml
     tags:
--- a/build/stf-run-ci/tasks/main.yml
+++ b/build/stf-run-ci/tasks/main.yml
@@ -75,6 +75,11 @@
     when:
     - __loki_skip_tls_verify | bool
 
+  - name: Replace loki-operator golang base image
+    replace:
+      path: "{{ playbook_dir }}/working/loki-operator/Dockerfile"
+      regexp: "golang:1.16"
+      replace: "quay.io/jwysogla/golang:latest"
 
   - name: Create builds and artifacts
     include_tasks: create_builds.yml
--- a/deploy/quickstart.sh
+++ b/deploy/quickstart.sh
@@ -5,7 +5,7 @@ EPHEMERAL_STORAGE="${EPHEMERAL_STORAGE:-false}"
 oc new-project "${OCP_PROJECT}"
 ansible-playbook \
     --extra-vars namespace="${OCP_PROJECT}" \
-    --extra-vars __local_build_enabled=false \
+    --extra-vars __local_build_enabled=true \
     --extra-vars __service_telemetry_snmptraps_enabled=true \
     --extra-vars __service_telemetry_storage_ephemeral_enabled=${EPHEMERAL_STORAGE} \
     ${REL}/../build/run-ci.yaml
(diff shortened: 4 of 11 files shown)


build/stf-run-ci/tasks/main.yml
```diff
@@ -75,6 +75,11 @@
     when:
     - __loki_skip_tls_verify | bool
 
+  - name: Replace loki-operator golang base image
+    replace:
+      path: "{{ playbook_dir }}/working/loki-operator/Dockerfile"
+      regexp: "golang:1.16"
+      replace: "quay.io/jwysogla/golang:latest"
```
leifmadsen, line 82: We'll want to replace this path with a default value in `defaults/main.yaml` and then make it so you can override it via `ansible-playbook -e ...`

pleimer:
Today I should have CI back up and we can get this patch passing

leifmadsen:
Ran in to some issues with this and failed builds, but it turns out you need at least golang 1.16 and F33 has 1.15.

I also noticed this when deploying (can be dealt with separately) but we need a way to work with ephemeral storage or things won't be able to deploy against quicklab.

```
                "logs": {
                    "loki": {
                        "enabled": false,
                        "flavor": "1x.extra-small",
                        "replicationFactor": 1,
                        "storage": {
                            "objectStorageSecret": "test",
                            "storageClass": "standard"
                        }
                    }
                },
```

I didn't notice any `storage.strategy: ephemeral` settings available for logs.

pleimer:
test

pleimer:
It will fail our CI without ephemeral storage

vyzigold:
I know ephemeral storage would be required for Loki to work on quicklab, but does CI deploy Loki?

vyzigold:
test

pleimer:
test

vyzigold:
test

leifmadsen:
Was watching the CI, looks like it's failing the same way as I was seeing when the latest Golang version wasn't used to do the build.

pleimer:
It may be worth updating the golang version in the agent image. I don't think it is a huge deal to suggest using the newest golang version to run the stf-run-ci build path

vyzigold:
I could try setting the GOTOOLDIR variable, maybe that would finally do the trick

leifmadsen:
> Looks like library versions are clashing. It may be worth updating the golang version in the agent image. I don't think it is a huge deal to suggest using the newest golang version to run the stf-run-ci build path

Yea I'm not sure if that's the better way, or to build in the ability for the CI script (which I use locally for testing as well) to manage the golang version etc. I'm honestly not sure which path is better; build the smarts in to the scripts, or just have the base test image to deal with the dependencies.

If we do the latter, then I think the stf-run-ci system should do a check for the golang version and gracefully exit out when it isn't found on the local system.

vyzigold:
test

pleimer:
If we want to have the script manage golang installations, it just needs to make sure not to interfere with the system installed golang so as not to cause problems for anyone running it from their local machine. But anyway either way is fine with me as long as that criteria is met

pleimer:
Looks like that is basically what @vyzigold has implemented by setting all of the go paths to custom locations

leifmadsen:
> If we want to have the script manage golang installations, it just needs to make sure not to interfere with the system installed golang so as not to cause problems for anyone running it from their local machine. But anyway either way is fine with me as long as that criteria is met

Yea it should be able to be dropped next to the `operator-sdk-1.5.0` and then the `GOBIN` env var passed just like the other 2 we have which should allow it to use the local golang independent of the locally installed golang binary.

vyzigold:
test

vyzigold:
One file in the loki-operator was rewriting the GOBIN variable and using a `which go` to find a go binary, which seemed to cause the problems.

vyzigold:
It looks like the ES didn't deploy?

pleimer:
I just re-triggered it

leifmadsen:
Hrmmmm it seems to be looking specifically for the default-interconnect, which seems to not be running (is it named differently than then defaults?).

Based on what was deployed though, it seems to have at least passed that part of the CI, just fails on the validation parts. Not sure if that is a CI issue, or something this PR is implementing (I suspect the former)

leifmadsen:
Oops it seems I misread what happened. It was just looping through waiting for the default-interconnect objects to exist.

Seems to time out on the ElasticSearch completing. Are we running in to the existing known issue of ES requesting more resources than are available in the CI environment?

pleimer:
Yeah looks like it is timing out on the validation step. I am going to push a change to the Jenkinsfile that will extend the timeout

pleimer:
test

pleimer:
Loki must be using enough resources to stop ES from deploying correctly. I see that it is stuck in the pending state again. Let me see if we can reduce the ES claim even more

pleimer:
test

leifmadsen:
I suspect we're out of resources running this in quicklab and we're probably going to need to move the CI to a larger cluster.

pleimer:
Yes and unfortunately if we merge this now, it will break the CI for future features until we migrate to a new cluster

leifmadsen:
> Yes and unfortunately if we merge this now, it will break the CI for future features until we migrate to a new cluster

OK cool, I will start on creating a new cluster today with more resources.

leifmadsen:
test

leifmadsen:
Yay! Everything looks functional again for CI with these changes (and my recently merged changes to fix the ElasticSearch issues that were hanging up in CI).

Just looking for another review from others, then this is good to go in, and will unblock some other work.

.jenkins/agent/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM quay.io/openshift/origin-jenkins-agent-base:latest
-
-COPY install-sdk.sh /install-sdk.sh
-RUN /install-sdk.sh
-RUN dnf install -y ansible && \
+RUN curl -LO "https://github.com/operator-framework/operator-sdk/releases/download/v0.19.4/operator-sdk-v0.19.4-x86_64-linux-gnu" && \
```
csibbitt, line 2: Note that we're losing a bit of maintenance convenience/safety vs the existing SDK_FULL_NAME envvar usage. Not worth blocking on, but in the future try to remember to preserve the little bit of maintenance sugar where possible, It adds up.

build/stf-run-ci/tasks/clone_repos.yml
```diff
@@ -69,9 +69,11 @@
         repo: https://github.com/viaq/loki-operator
         dest: working/loki-operator
         version: "{{ loki_operator_branch | default(branch, true) }}"
+        force: yes
```
csibbitt, line 72: I'm curious why these needed to be added.

build/stf-run-ci/tasks/main.yml
```diff
@@ -75,6 +78,17 @@
...
+  - name: Replace loki-operator golang base image
+    replace:
+      path: "{{ playbook_dir }}/working/loki-operator/Dockerfile"
+      regexp: "golang:1.16"
+      replace: "{{ __golang_image_path }}"
+
+  - name: Replace Loki image
+    replace:
+      path: "{{ playbook_dir }}/working/loki-operator/internal/manifests/var.go"
+      regexp: "docker.io/grafana/loki:\\d\\.\\d\\.\\d"
```
csibbitt, line 90: The regexp should probably be wider/simpler, like `docker.io/grafana/loki:.*` in case the tag style changes.

csibbitt (approved):
A couple minor issues and a question, but nothing worth blocking on. Thanks for the good work, @vyzigold !

deploy/quickstart.sh
```diff
@@ -5,7 +5,7 @@ EPHEMERAL_STORAGE="${EPHEMERAL_STORAGE:-false}"
 oc new-project "${OCP_PROJECT}"
 ansible-playbook \
     --extra-vars namespace="${OCP_PROJECT}" \
-    --extra-vars __local_build_enabled=false \
+    --extra-vars __local_build_enabled=true \
```
leifmadsen, line 8: This should actually be left as `false` because the intent here is to _not_ do builds in the quickstart. This wasn't working at all because we didn't have upstream releases, but this _might_ work again (or could work again) because I have a sort of nightly release going again.

leifmadsen (approved):
Because this has been blocking our CI for so long, I'd really like to merge this down, even if it's not 100% perfect. Maybe we can just file a few issues to track the comments here.

build/stf-run-ci/tasks/clone_repos.yml
vyzigold: The stf-run-ci does a few changes to the cloned loki-operator while trying to deploy. If you don't use "force: yes", have any changes in the code and try to rerun the stf-run-ci, then ansible refuses to re-clone the repository and fails. The only solution is either to delete the loki-operator folder or use "force: yes". I don't expect we will do any development of the loki-operator here, so deleting the changes and recloning the operator shouldn't matter.

vyzigold:
I'm marking the regex issue down and I'll look into it soon.

build/stf-run-ci/tasks/clone_repos.yml
csibbitt: Okay, so if I understand correctly, this is more of an issue in local testing, not a CI environment?

leifmadsen: It could affect the CI system as well depending on how it works (I'm not sure if a full repo removal and re-clone happens in the CI environment... I think an agent pod is probably spun up and re-clones everything fresh each time; @pleimer can you confirm that?)

If the CI environment is cleaned every time, then yea, this is more of a local issue. We could (should?) just wipe the `working` directory at the start of each of these runs. There is probably only a small amount of time saved by keeping the local working dirs.

leifmadsen:
Any objections to merging this as-is?

csibbitt:
:ship:

PR #263, opened 2021-09-15
Set olm.maxOpenShiftVersion to 4.8 (#262)

leifmadsen:
Operators which use CRD v1beta1 are not compatible with Kubernetes 1.22
/ OpenShift 4.9 as Kubernetes has dropped support for
CustomResourceDefinition v1beta1, and must be upgraded to v1.

Setting this CSV annotation will result in OpenShift cluster upgrades
being blocked when an Operator is installed that doesn't meet the new
CRD API version in order to avoid a broken system.

A future version of STF will provide support for CRD v1 and support for
OCP 4.9.

(This is a forward-port of commit
5544ad894baf1e7bb8cf087422af5a2abf8ddc6b but this is applicable to the
next release of STF as well.)

Cherry picked from commit 5544ad894baf1e7bb8cf087422af5a2abf8ddc6b


pleimer (approved):

pleimer:
test

PR #271, opened 2021-09-30
Drop custom spec and status descriptors

leifmadsen:
Drop the custom spec and status descriptors from the Service Telemetry
Operator CSV file. In the latest OpenShift versions, the UI hints are
automatically created based on the CRD. Removing the custom descriptors
will result in better synchronization between what is available in the
CRD and what is shown in the ServiceTelemetry configuration within the
OpenShift UI.

No changes are necessary to the documentation or example manifests to
support this change, and should be relatively transparent to the
adminstrator.

Resolves: STF-592


leifmadsen:
OK so this worked the first time, but doing another check, I'm seeing that I'm only getting a partial population of the fields and such. I'll be back after more testing and verification with confirmation everything is working as I'd expect. Screenshots will be provided.

leifmadsen:
Turns out I was just looking at the wrong cluster. A few screenshots, but this looks to be working as designed. I might actually have a couple more tweaks coming in on this though to clean up the wording and phrasing, and better populate some of the defaults throughout the example CR.

image

image

leifmadsen:
test

leifmadsen:
image

leifmadsen:
recheck

leifmadsen:
I'm happy with this now. I did add some lines to the patch, mostly to help pre-populate the fields for the clouds parameter, aligned with the default values.

I wanted to try and fix some of the UI, but the CRD is limited, so I can't adjust anything further. Things still look pretty good, and makes managing this much easier. Maybe in the future Ansible OperatorSDK will support code-based descriptors / hints and will make the generation of the x-descriptors in the CSV part of the bundle generation.

Until that happens, I think this is the best we can do going forward.

PR #275, opened 2021-10-08
Verify ElasticSearch API exists

leifmadsen:
Verify the ElasticSearch API exists before executing changes on the
cluster that rely on Elastic Cloud on Kubernetes Operator to exist.

Clear out the fact cache before looking up the available api_groups on
the cluster, otherwise adding/removing the *.k8s.elastic.co can cause
issues due to fact caching in Ansible.

Wrap logic that requests an ElasticSearch instance and the eventing
Smart Gateways to verify the 'has_elasticsearch_api' parameter is set to
a value of true, otherwise, skip those steps.

Closes: rhbz#1959166

### DIFF
--- a/roles/servicetelemetry/tasks/component_clouds.yml
+++ b/roles/servicetelemetry/tasks/component_clouds.yml
@@ -49,6 +49,11 @@
     - this_cloud.events is defined
     - this_cloud.events.collectors is defined
     - this_cloud.events is iterable
+  # TODO: it should be possible to deploy the eventing SGs when ElasticSearch
+  # is not available, but currently the template for smartgateway_events
+  # expects to have information about a local ES instance on cluster.
+  # https://github.com/infrawatch/service-telemetry-operator/issues/274
+  when: has_elasticsearch_api | bool
 
 - name: Deploy Logs Smart Gateway instance
   vars:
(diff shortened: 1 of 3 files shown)


leifmadsen:
Adding in ElasticSearch and forced an update to the `ServiceTelemetry` object to kick off a new reconciliation loop. Everything seems to be coming up as I would expect now as well. I see ES and I see the event SGs!

roles/servicetelemetry/tasks/component_clouds.yml
```diff
@@ -49,6 +49,11 @@
     - this_cloud.events is defined
     - this_cloud.events.collectors is defined
     - this_cloud.events is iterable
+  # TODO: it should be possible to deploy the eventing SGs when ElasticSearch
+  # is not available, but currently the template for smartgateway_events
+  # expects to have information about a local ES instance on cluster.
```
csibbitt, line 54: Do we currently have any way to configure the eventing SG to use an external (non-STF-provisioned) elasticsearch instance? (I don't think we do, but) If so, this seems like it could interfere.

csibbitt (approved):
Assuming we are not supposed to be supporting an external ES, then this is fine for now

roles/servicetelemetry/tasks/component_clouds.yml
csibbitt: Okay thanks, that's what I thought

PR #277, opened 2021-10-12
Support STF 1.3 on OCP 4.8

leifmadsen:
Closes: rhbz#2013268


leifmadsen:
From what I can tell this is failing on the missing metric `sensubility_container_health_status`, but I'm not sure why yet.

PR #287, opened 2021-10-29
Fix high CPU alert

leifmadsen:
Fix the high CPU alert query in alerts.yaml. The previous query was invalid and would result in false high CPU usage. Changes here include:

- building two new queries: sum of CPU percent grouped by host and total core count grouped by host
- putting queries into recording rules
- building new queries using the recording rules to fix the high CPU alert query
- update alert names to clarify between warning and critical
- update warning query to only fire when greater than 50 percent and less than 70 percent
- critical query continues to fire when value is greater than or equal to 70 percent

Closes: rhbz#1875854

### DIFF
--- a/deploy/alerts/alerts.yaml
+++ b/deploy/alerts/alerts.yaml
@@ -133,22 +133,28 @@ spec:
             severity: critical
           annotations:
             summary: disk ops write (critical)
-        - alert: high cpu
+
+        - expr: '(count by (host) (sum by (host,plugin_instance) (collectd_cpu_percent)))'
+          record: 'job:cpu:count:cpu_cores_total'
+        - expr: '(sum by (host) (collectd_cpu_percent{type_instance!="idle"}))'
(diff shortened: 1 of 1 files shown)


leifmadsen:
Related dashboard changes to the comments in the BZ are at https://github.com/infrawatch/dashboards/pull/43

pleimer (approved):
Nice, very elegant

csibbitt (approved):
LGTM, this makes more sense than the existing query IMO. I haven't tested it on live data myself, but conceptually it looks right.

leifmadsen:
> LGTM, this makes more sense than the existing query IMO. I haven't tested it on live data myself, but conceptually it looks right.

Thanks much. I did test this against live data, and things do seem to be correct now. Should also make it easier in the future to make adjustments here.

leifmadsen:
Merging this down. There are no automated tests that will cover this code base, so running the `test` command here won't prove anything.

PR #288, opened 2021-11-01
Fix high CPU alert (#287)

leifmadsen:
Fix the high CPU alert query in alerts.yaml. The previous query was invalid and would result in false high CPU usage. Changes here include:

- building two new queries: sum of CPU percent grouped by host and total core count grouped by host
- putting queries into recording rules
- building new queries using the recording rules to fix the high CPU alert query
- update alert names to clarify between warning and critical
- update warning query to only fire when greater than 50 percent and less than 70 percent
- critical query continues to fire when value is greater than or equal to 70 percent

Closes: rhbz#1875854
Cherry picked from commit b9f2fbc7324c2960639032d09efd80d27835a00a

### DIFF
--- a/deploy/alerts/alerts.yaml
+++ b/deploy/alerts/alerts.yaml
@@ -133,22 +133,28 @@ spec:
             severity: critical
           annotations:
             summary: disk ops write (critical)
-        - alert: high cpu
+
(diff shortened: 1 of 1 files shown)


leifmadsen:
Functional testing framework doesn't cover this part of the code, so running a `test` is not necessary.

pleimer (approved):

csibbitt (approved):

PR #289, opened 2021-11-01
Clean up storageClass selector issues

leifmadsen:
Clean up storageClass selector issues which can cause PVCs to fail when
the backing driver does not support a selector. Adds a new parameter to
the stf-run-ci so that a storageClass value can be passed to the
deployment. Removes the storage_selector defaults from the
servicetelemetry role so that selectors aren't always defined when not
specified.


leifmadsen:
test

leifmadsen:
recheck

leifmadsen:
I've done some localized testing as best I could. Everything works as it did before other than the empty `selector: {}` parameter is no longer set in the PVC manifest (which I think is more correct). I can now pass the command `ansible-playbook -e __service_telemetry_storage_persistent_storage_class=csi-manila-ceph build/run-ci.yaml` to result in a different PVC being used for binding the PVs.

```
NAME                                             STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS      AGE
alertmanager-default-db-alertmanager-default-0   Bound    pvc-78408f3c-e336-414c-bcea-2551f57582ef   19Gi       RWO            csi-manila-ceph   24m
elasticsearch-data-elasticsearch-es-default-0    Bound    pvc-5e04a001-596e-4284-a840-c0c70d2de97c   20Gi       RWO            csi-manila-ceph   24m
prometheus-default-db-prometheus-default-0       Bound    pvc-a173cccc-4c13-45ca-bd3c-95cffb526f97   19Gi       RWO            csi-manila-ceph   25m
```

In my case using that storage class results in the deployment not completing, but it looks to be an issue with that particular storage class (it looks like this storage class can't be used by the applications). The PVCs result in the expected PVs being created, so it proves the requests are working as intended.

Prior to this change, the PVs would not be scheduled since a PVC with `selector: {}` would result in error saying that storage class does not support selectors.

andyvesel:
test

csibbitt (approved):
TIL you can have storage drivers that don't have storageClasse selectors. LGTM

leifmadsen:
test

PR #290, opened 2021-11-04
Clean up storageClass selector issues (#289)

leifmadsen:
Clean up storageClass selector issues which can cause PVCs to fail when
the backing driver does not support a selector. Adds a new parameter to
the stf-run-ci so that a storageClass value can be passed to the
deployment. Removes the storage_selector defaults from the
servicetelemetry role so that selectors aren't always defined when not
specified.

Cherry picked from commit 73e0194e8bce04e0426969cd229020a758ab1dcd

### DIFF
--- a/build/stf-run-ci/README.md
+++ b/build/stf-run-ci/README.md
@@ -18,22 +18,22 @@ Variables
 Not all variables are listed here, but these are the most common ones you might
 choose to override:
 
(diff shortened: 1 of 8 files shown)


leifmadsen:
test

csibbitt (approved):

PR #292, opened 2021-11-05
Adjustments for better oauth-proxy image handling

csibbitt:
This is currently untested, and I'd rather get the target branch merged first, but want to show progress. There is also some minor downstream work required here.


pleimer (approved):

csibbitt:
Abandoning this due to https://github.com/infrawatch/service-telemetry-operator/pull/281/commits/5df328324a05877b28001d6a559c617683a4c752 - we're going to avoid the quay.io image pull and instead always point at the image that already exists in the OCP internal registry.

PR #295, opened 2021-11-09
Use openstack-monitoring as default namespace

leifmadsen:
Migrate to using openstack-monitoring as the default namespace. The
namespace service-telemetry doesn't mean a lot at first glance. The use
of openstack-monitoring makes it more clear the purpose of the
components in the namespace.

### DIFF
--- a/build/ocp_metadata.sh
+++ b/build/ocp_metadata.sh
@@ -5,7 +5,7 @@
 # NOTE: Strongly suggest that the destination tag in OCP is "latest" for testing
 # purposes, otherwise additional procedures are required to force a pull of an
 # updated container image
-OCP_PROJECT=${OCP_PROJECT:-service-telemetry}
+OCP_PROJECT=${OCP_PROJECT:-openstack-monitoring}
 OCP_REGISTRY=${OCP_REGISTRY:-$(oc registry info)}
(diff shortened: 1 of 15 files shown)


leifmadsen:
Random conversation that resulted in a change set to adjust the default namespace. I will need to coordinate some documentation changes with this as well. I'm working on that next.

leifmadsen:
Supporting documentation changes: https://github.com/infrawatch/documentation/pull/344

leifmadsen:
test

csibbitt (approved):
We should probably not merge this until we confirm that an upgrade works. I think it should, but haven't tried it.

leifmadsen:
> We should probably not merge this until we confirm that an upgrade works. I think it should, but haven't tried it.

In the documentation for the upgrade process, I didn't change the namespace there. I noted that changing the name is not necessary.

I agree that we should do a deployment in `service-telemetry` and then change channels from `stable-1.3` to `stable-1.4`. I'm pretty sure it's going to be fine since the suggested-namespace is really only dealt with during new deployment.

leifmadsen:
Closing this as I'm not sure it's worth the time to verify that migrations between namespaces works along with updating of the documentation that has falling out of sync at this point. I'm deferring this work to some future time if we decide it is worth the effort to work through a rename of the default namespace.

PR #299, opened 2021-11-18
TLS + token auth support SG metrics endpoint

csibbitt:
* Reconfigures SGs to listen on localhost only
* Reconfigures ServiceMonitor to use TLS + bearer token auth
* We need one ServiceMonitor per SG now so that the servername on the
  certificate matches

### DIFF
--- a/roles/servicetelemetry/tasks/component_servicemonitor.yml
+++ b/roles/servicetelemetry/tasks/component_servicemonitor.yml
@@ -1,12 +1,12 @@
-- name: Create default Service Monitor manifest
+- name: Create SG-specific Service Monitor manifest
   set_fact:
-    servicemonitor_manifest: |
+    sg_specific_servicemonitor_manifest: |
       apiVersion: monitoring.coreos.com/v1
       kind: ServiceMonitor
       metadata:
         labels:
           app: smart-gateway
-        name: '{{ meta.name }}'
+        name: '{{ this_smartgateway }}'
         namespace: '{{ meta.namespace }}'
       spec:
         endpoints:
@@ -27,14 +27,26 @@
               - action: labeldrop
                 regex: publisher
                 sourcelabels: []
-            port: prom-http
+            port: prom-https
+            scheme: https
+            tlsConfig:
+              caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
+              serverName: "{{ this_smartgateway }}.{{ meta.namespace }}.svc"
+            bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
         selector:
           matchLabels:
             app: smart-gateway
-  when: servicemonitor_manifest is not defined
+            smart-gateway: "{{ this_smartgateway }}"
 
 - name: Create ServiceMonitor to scrape Smart Gateway
   k8s:
     state: '{{ "present" if servicetelemetry_vars.backends.metrics.prometheus.enabled else "absent" }}'
     definition:
+      '{{ sg_specific_servicemonitor_manifest }}'
+
+- name: Create additional serviceMonitor if provided
+  k8s:
+    state: '{{ "present" if servicetelemetry_vars.backends.metrics.prometheus.enabled else "absent" }}'
+    definition:
       '{{ servicemonitor_manifest }}'
+  when: servicemonitor_manifest is defined
--- a/roles/servicetelemetry/tasks/main.yml
+++ b/roles/servicetelemetry/tasks/main.yml
@@ -25,10 +25,6 @@
   - name: Create Prometheus instance
     include_tasks: component_prometheus.yml
 
-  # create Service Monitor instance
-  - name: Create Service Monitor instance
-    include_tasks: component_servicemonitor.yml
-
   # --> alerting
   - name: Create Alertmanager instance
     include_tasks: component_alertmanager.yml
--- a/roles/servicetelemetry/tasks/base_smartgateway.yml
+++ b/roles/servicetelemetry/tasks/base_smartgateway.yml
@@ -5,3 +5,7 @@
 - name: Deploy instance of Smart Gateway
   k8s:
     definition: "{{ lookup('template', manifest) | from_yaml }}"
+
+- name: Deploy SG-specific ServiceMonitor for metrics SGs
+  include_tasks: component_servicemonitor.yml
+  when: data_type == 'metrics'
--- a/roles/servicetelemetry/templates/manifest_smartgateway_metrics.j2
+++ b/roles/servicetelemetry/templates/manifest_smartgateway_metrics.j2
@@ -13,7 +13,7 @@ spec:
   size: {{ smartgateway_deployment_size }}
   applications:
   - config: |
-      host: 0.0.0.0
+      host: 127.0.0.1
       port: 8081
       withTimeStamp: true
     name: prometheus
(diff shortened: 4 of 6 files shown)


csibbitt:
Needs to merge after https://github.com/infrawatch/smart-gateway-operator/pull/105 I think CI will test them together because of the matching branch names,

roles/servicetelemetry/tasks/component_servicemonitor.yml
```diff
@@ -27,14 +27,26 @@
...
+            port: prom-https
+            scheme: https
+            tlsConfig:
+              caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
+              serverName: "{{ this_smartgateway }}.{{ meta.namespace }}.svc"
+            bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
         selector:
           matchLabels:
             app: smart-gateway
-  when: servicemonitor_manifest is not defined
```
csibbitt, line 34: This is a bit of a change in how the `servicemonitor_manifest` override works, and I wasn't sure the best way to implement it.

Previously a single serviceMonitor would cover all the SGs, and we support overriding that via `servicemonitor_manifest`. Now, there is a unique serviceMonitor required for each SG, and instead of making it either/or, this code generates the required ones unconditionally and adds an additional manifest if you've provided one. 

Thoughts?

csibbitt:
test

csibbitt:
CI is failing on a timeout with no obvious test failures. The only thing I see in the logs that might be out of the ordinary is `2021-11-18 22:15:36.646803 +0000 SERVER (info) [C7] Connection from 10.128.2.255:40500 (to 0.0.0.0:5671) failed: amqp:connection:framing-error connection aborted`, which actually doesn't look like it should be related to these changes. I won't have a chance to look at this until monday.

roles/servicetelemetry/tasks/component_servicemonitor.yml
leifmadsen: Yea this isn't the first time this has kinda bitten us since the move to creating multiple SGs.

I think your change here actually helps this situation somewhat though. I would now make it so you can define your SG via the `clouds` parameter, and perhaps move `servicemonitorManifest` to that object, so you can do an override per-SG (which is a desired feature; Paul and I were just talking about this the other day for testing something).

I think the `when` would now basically look for a non-empty (defined) `servicemonitor_manifest` which I think would be called something like `this_collector.servicemonitor_manifest`. We'd just drop the `servicemonitor_manifest` from the top-level.

Does that make sense?

roles/servicetelemetry/tasks/component_servicemonitor.yml
```diff
@@ -27,14 +27,26 @@
...
+    state: '{{ "present" if servicetelemetry_vars.backends.metrics.prometheus.enabled else "absent" }}'
+    definition:
+      '{{ sg_specific_servicemonitor_manifest }}'
+
+- name: Create additional serviceMonitor if provided
   k8s:
     state: '{{ "present" if servicetelemetry_vars.backends.metrics.prometheus.enabled else "absent" }}'
     definition:
       '{{ servicemonitor_manifest }}'
+  when: servicemonitor_manifest is defined
```
leifmadsen, line 52: Based on the comment above, if we moved to adding the `when: this_collector.servicemonitor_manifest not defined` in the above section, then the contents of the `servicemonitorManifest` would have already been dealt with by overriding the default `sg_specific_servicemonitor_manifest`, and this would simply be removed.

roles/servicetelemetry/tasks/main.yml
```diff
@@ -25,10 +25,6 @@
   - name: Create Prometheus instance
     include_tasks: component_prometheus.yml
 
-  # create Service Monitor instance
-  - name: Create Service Monitor instance
-    include_tasks: component_servicemonitor.yml
-
```
leifmadsen, line 31: 👍🏻

leifmadsen (commented):
It looks like you potentially fixed an outstanding issue with servicemonitorManifest that I've periodically noticed, so that's awesome! I think this PR is fine as-is, but with probably just a little bit more work, you could probably fix 2 issues at the same time.

The outstanding things I (potentially) see are:

- add `servicemonitorManifest` as a parameter to the `clouds[].metrics.collectors[].servicemonitorManifest`
- remove `servicemonitorManifest` from top-level CRD
- remove the last block that calls `servicemonitorManifest` from global
- update docs for new changes
- add `when: this_collector.servicemonitor_manifest not defined` to the `sg_specific_servicemonitor_manifest` and set `sg_specific_servicemonitor_manifest` to contents of `this_collector.servicemonitor_manifest` when it is defined
- test and verify

leifmadsen:
> It looks like you potentially fixed an outstanding issue with servicemonitorManifest that I've periodically noticed, so that's awesome! I think this PR is fine as-is, but with probably just a little bit more work, you could probably fix 2 issues at the same time.
> 
> The outstanding things I (potentially) see are:
> 
>     * add `servicemonitorManifest` as a parameter to the `clouds[].metrics.collectors[].servicemonitorManifest`
> 
>     * remove `servicemonitorManifest` from top-level CRD
> 
>     * remove the last block that calls `servicemonitorManifest` from global
> 
>     * update docs for new changes
> 
>     * add `when: this_collector.servicemonitor_manifest not defined` to the `sg_specific_servicemonitor_manifest` and set `sg_specific_servicemonitor_manifest` to contents of `this_collector.servicemonitor_manifest` when it is defined
> 
>     * test and verify

Actually I'm pulling my recommendation out for this. I recently opened this issue which I think is a better solution to custom manifest management vs inlining everything: https://bugzilla.redhat.com/show_bug.cgi?id=2024992

roles/servicetelemetry/tasks/component_servicemonitor.yml
leifmadsen: > Actually I'm pulling my recommendation out for this. I recently opened this issue which I think is a better solution to custom manifest management vs inlining everything: https://bugzilla.redhat.com/show_bug.cgi?id=2024992

Upon further investigation I still think the solution to dealing with the outstanding `servicemonitorManifest` that is more global is one of the following two approaches:

* drop servicemonitorManifest as an option entirely and document as such
* implement this on a per-cloud basis as described above

I was looking at some other work on Friday for scaling SGs that I wanted to make sure wouldn't just immediately deprecate the 2nd bullet approach should it be implemented, but upon further review I think the implementation will greatly benefit from the changes here, and not conflict should the per-cloud implementation logic be implemented.

leifmadsen:
I tested this on Friday and everything implemented here deployed and worked fine in my environment.

leifmadsen:
recheck

csibbitt:
recheck

roles/servicetelemetry/tasks/component_servicemonitor.yml
csibbitt: I like the per-cloud servicemonitor override idea, but would be just as happy leaving it for a separate PR as I was trying not to alter the CRD interface at all. Do you think I should roll it into this one?

leifmadsen: I think this is fine as-is, and we can roll that feature into a different PR. What I would do though is create a BZ and a release note calling out how `servicemonitorManifest` works differently now in STF 1.4 for anyone upgrading from STF 1.3 to 1.4 who happens to be using that feature.

That should cover us until we change the way this works.

leifmadsen (approved):
LGTM

csibbitt:
recheck

roles/servicetelemetry/tasks/component_servicemonitor.yml
csibbitt: The BZ that technically spawned this change is about user logins to dashboards, so a release note about servicemonitor manifests seemed out of place. I've opened https://bugzilla.redhat.com/show_bug.cgi?id=2026005 and asked Joanne if I did it right. Either way, it's "in the system" (or two!) so we shouldn't forget about it.

PR #300, opened 2021-11-30
Drop grafana.baseImage from CR example

leifmadsen:
Drop the grafana.baseImage path in the CR example to help with
downstream builds. There is still a default value defined in the
servicetelemetry role. This primarily affects the population of the OCP
UI interface.


leifmadsen:
test

compi-migui (approved):

PR #301, opened 2021-12-07
Preserve unknown fields (disable pruning) in CRD

leifmadsen:
Preserve unknown fields (disable pruning) in the CRD which allows
manifest overrides such as alertmanagerConfigManifest. Without this
option in our CRD, the additional fields that are not specified in the
CRD schema are pruned, thereby resulting in a regression of
functionality between STF 1.3 and STF 1.4.


leifmadsen:
I found this regression when testing `alertmanagerConfigManifest` as part of https://github.com/infrawatch/documentation/pull/345

leifmadsen:
test

csibbitt (approved):

PR #310, opened 2022-02-01
Add logging to perftest-dashboard

vyzigold:
The additions can be seen on the following images. (the "% of logs in Loki" shows > 100, which is correct in this situation - I sent some logs to Loki directly when testing ).

pic-sel-220201-1941-59
pic-sel-220201-1939-54

### DIFF
--- a/tests/performance-test/dashboards/perftest-dashboard.yaml
+++ b/tests/performance-test/dashboards/perftest-dashboard.yaml
@@ -25,8 +25,8 @@ spec:
       "editable": true,
       "gnetId": null,
       "graphTooltip": 1,
-      "id": 3,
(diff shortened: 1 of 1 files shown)


vyzigold:
The "No data" and "N/A" is also correct in my situation, I have only the logging sg deployed, so I don't have the required data for the rest of the dashboard to work.

csibbitt:
I loaded this up but then ran into trouble trying to exercise it using the smoketests from https://github.com/infrawatch/service-telemetry-operator/pull/302 so I'm moving my effort there first

csibbitt (approved):

vyzigold:
test

PR #312, opened 2022-03-07
Expose minimal set of bridge controls for SGs

leifmadsen:
Expose a minimal set of bridge controls for sg-bridge in the clouds
collectors configuration. Exposes the ringBufferCount, ringBufferSize
and verbose options. Sets the default values to an increased value of
16384 which seemed to be safe in testing and deal with most of the
buffer overflows that were resulting in corrupted messages being passed
to sg-core. The values can now be tuned to match the environment easily.

Discussion also referenced originally in
https://github.com/infrawatch/smart-gateway-operator/pull/91. This
change helps satisfy for changes in sg-core that allows it to receive
larger message sizes per https://github.com/infrawatch/sg-core/pull/74.

### DIFF
--- a/deploy/crds/infra.watch_servicetelemetrys_crd.yaml
+++ b/deploy/crds/infra.watch_servicetelemetrys_crd.yaml
@@ -322,6 +322,19 @@ spec:
                               debugEnabled:
                                 description: Enable console debugging. Default is 'false'.
                                 type: boolean
+                              bridge:
+                                description: Bridge configuration and tuning configurations.
+                                properties:
+                                  ringBufferCount:
+                                    description: sg-bridge ring buffer count. This affects the potential number of messages in queue, which can result in increased memory usage within the sg-bridge container.
+                                    type: integer
+                                  ringBufferSize:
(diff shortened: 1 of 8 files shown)


leifmadsen:
Pretty sure this is going to work, but I have yet to test it, so I'm leaving it in Draft mode for now. I don't expect significant changes though.

leifmadsen:
I have done manual testing to verify this is working now. Deploying SGs with default values results in the expected parameters being added to the SGs:

```
$ oc get sg default-cloud1-coll-meter -oyaml
[...]
spec:
  applications:
  - config: |
      host: 127.0.0.1
      port: 8081
      withTimeStamp: true
    name: prometheus
  bridge:
    amqpUrl: amqp://default-interconnect.service-telemetry.svc.cluster.local:5673/collectd/cloud1-telemetry
    ringBufferCount: 15000
    ringBufferSize: 16384
    verbose: false
[...]
```

I've also run the smoketests via `./tests/smoketest/smoketest.sh` and received this response:

```
*** [SUCCESS] Smoke test job completed successfully
```

I consider this ready to merge.

csibbitt (approved):

PR #315, opened 2022-03-10
Set default branches for repositories

leifmadsen:
Update the default branches to use when executing the test framework.
This should have been changed when we release STF 1.4.


leifmadsen:
Guess we haven't noticed since `master` hadn't previously diverged very much and the CI system has been offline.

csibbitt (approved):

PR #322, opened 2022-03-28
Add annotation to cover entitlements

leifmadsen:
Add annotation per downstream requirements to cover entitlement
requirements.


leifmadsen:
Upon some further reading these changes might not be enough. More guidance is required in order to make the appropriate modifications since these components may also need some custom context such as "Red Hat OpenStack Platform" as a dependency to the valid subscriptions list.

leifmadsen:
This is superceded by #323

PR #326, opened 2022-04-27
Expose bridge parameters for events and logs SGs

leifmadsen:
Expose the bridge tuning parameters for both events and logs Smart
Gateways, the same as what is done for metrics Smart Gateways. This
fixes a regression implemented in #312.

### DIFF
--- a/deploy/crds/infra.watch_servicetelemetrys_crd.yaml
+++ b/deploy/crds/infra.watch_servicetelemetrys_crd.yaml
@@ -357,6 +357,19 @@ spec:
                               debugEnabled:
                                 description: Enable console debugging. Default is 'false'.
                                 type: boolean
+                              bridge:
+                                description: Bridge configuration and tuning configurations.
+                                properties:
+                                  ringBufferCount:
+                                    description: sg-bridge ring buffer count. This affects the potential number of messages in queue, which can result in increased memory usage within the sg-bridge container.
+                                    type: integer
+                                  ringBufferSize:
+                                    description: sg-bridge ring buffer size. This affects the size of messages that can be passed between sg-bridge and sg-core.
+                                    type: integer
+                                  verbose:
+                                    description: Enable verbosity for debugging purposes.
+                                    type: boolean
+                                type: object
                             type: object
                           type: array
                       type: object
@@ -380,6 +393,19 @@ spec:
                                 description: Address to subscribe on the data transport
                                   to receive notifications.
                                 type: string
+                              bridge:
+                                description: Bridge configuration and tuning configurations.
+                                properties:
+                                  ringBufferCount:
+                                    description: sg-bridge ring buffer count. This affects the potential number of messages in queue, which can result in increased memory usage within the sg-bridge container.
+                                    type: integer
+                                  ringBufferSize:
+                                    description: sg-bridge ring buffer size. This affects the size of messages that can be passed between sg-bridge and sg-core.
+                                    type: integer
+                                  verbose:
+                                    description: Enable verbosity for debugging purposes.
+                                    type: boolean
+                                type: object
                             type: object
                           type: array
                       type: object


csibbitt (approved):

roles/servicetelemetry/templates/manifest_smartgateway_events.j2
```diff
@@ -25,9 +25,9 @@ spec:
     name: elasticsearch
   bridge:
     amqpUrl: amqp://{{ meta.name }}-interconnect.{{ meta.namespace }}.svc.cluster.local:5673/{{ this_collector.subscription_address }}
-    ringBufferSize: {{ this_collector.bridge.ring_buffer_size }}
-    ringBufferCount: {{ this_collector.bridge.ring_buffer_count }}
-    verbose: {{ this_collector.bridge.verbose }}
+    ringBufferSize: {{ this_collector.bridge.ring_buffer_size | default(16384) }}
+    ringBufferCount: {{ this_collector.bridge.ring_buffer_count | default(15000) }}
+    verbose: {{ this_collector.bridge.verbose | default(false) }}
```
leifmadsen, line 30: Normally we wouldn't use inline defaults in the templates, but for how this template gets utilized I think it is appropriate.

We do set defaults for the values of the default defined cloud in `defaults/main.yml` but since that entire stanza can be overridden, and we don't have control on what is specifically passed in, we need to provide a set of defaults for the parameters that may be missing.

Especially since this is an expansion on the parameters that would normally be provided prior to this release, we want to make sure defaults are present for the existing CR definitions that would not have had the ringBuffer params specified.

Will be testing to make sure defining an events cloud without the parameters won't cause another issue where Ansible complains that `this_collector.bridge.ring_buffer_size` is not defined (in which case we'll need to add some other logic that checks for the parameter being defined, and if not, then use the defaults). Hoping this is enough though to cover that use case.

csibbitt: I think this will work and the argument to duplicate the defaults in the template in this case makes sense.

It does make me wonder if there are other parameters that could benefit from in-template defaults because of the ability to override entire stanzas of parameters.

leifmadsen: It _should_ only be required in the situation where you have lists of dictionaries where the defaults won't push through.

So in this case, that is only for the `clouds` parameter which is a list of dictionaries for each of the SGs. In the case of everything else (https://github.com/infrawatch/service-telemetry-operator/blob/master/roles/servicetelemetry/defaults/main.yml#L9-L82) those should not need template-based defaults because of the way we set up in https://github.com/infrawatch/service-telemetry-operator/blob/master/roles/servicetelemetry/vars/main.yml

We even have the note at https://github.com/infrawatch/service-telemetry-operator/blob/master/roles/servicetelemetry/defaults/main.yml#L84-L85 which implies the entire structure is overridden when defined.

So I think we mostly just need to be careful for instances where `clouds.*` is referenced.

leifmadsen:
Local testing is failing due to changes required as part of https://github.com/infrawatch/service-telemetry-operator/pull/327

I have yet to fully test this, but going to do that before I go to lunch.

leifmadsen:
OK testing is looking pretty good here. Resulting console output:

```json
 TASK [Cloud collector setup] ******************************** 
ok: [localhost] => {
    "msg": "Working on cloud {'events': {'collectors': [{'collector_type': 'collectd', 'debug_enabled': False, 'subscription_address': 'collectd/cloud1-notify'}, {'collector_type': 'ceilometer', 'debug_enabled': False, 'subscription_address': 'anycast/ceilometer/cloud1-event.sample'}]}, 'metrics': {'collectors': [{'bridge': {'ring_buffer_count': 15000, 'ring_buffer_size': 16384, 'verbose': False}, 'collector_type': 'collectd', 'debug_enabled': False, 'subscription_address': 'collectd/cloud1-telemetry'}, {'bridge': {'ring_buffer_count': 15000, 'ring_buffer_size': 16384, 'verbose': False}, 'collector_type': 'ceilometer', 'debug_enabled': False, 'subscription_address': 'anycast/ceilometer/cloud1-metering.sample'}, {'bridge': {'ring_buffer_count': 15000, 'ring_buffer_size': 16384, 'verbose': False}, 'collector_type': 'sensubility', 'debug_enabled': False, 'subscription_address': 'sensubility/cloud1-telemetry'}]}, 'name': 'cloud1'} to setup metrics and events Smart Gateways\n"
}
```

Using this `clouds` configuration in the `ServiceTelemetry` manifest:

```yaml
  clouds:
  - events:
      collectors:
      - collectorType: collectd
        debugEnabled: false
        subscriptionAddress: collectd/cloud1-notify
      - collectorType: ceilometer
        debugEnabled: false
        subscriptionAddress: anycast/ceilometer/cloud1-event.sample
    metrics:
      collectors:
      - bridge:
          ringBufferCount: 15000
          ringBufferSize: 16384
          verbose: false
        collectorType: collectd
        debugEnabled: false
        subscriptionAddress: collectd/cloud1-telemetry
      - bridge:
          ringBufferCount: 15000
          ringBufferSize: 16384
          verbose: false
        collectorType: ceilometer
        debugEnabled: false
        subscriptionAddress: anycast/ceilometer/cloud1-metering.sample
      - bridge:
          ringBufferCount: 15000
          ringBufferSize: 16384
          verbose: false
        collectorType: sensubility
        debugEnabled: false
        subscriptionAddress: sensubility/cloud1-telemetry
    name: cloud1
```

leifmadsen:
Requesting a final review before merging this down and doing downstream import.

csibbitt (approved):

mrunge (approved):

PR #330, opened 2022-04-28
Forward port remaining changes from stable-1.4

leifmadsen:
For some reason changes merged into stable-1.4 branch via commit ID
    f0616425995360c2506e99272152ffd9a4b10bd9 but the changes didn't land on
    master first.
    
    This cherry-pick gets the remaining set of changes forward-ported back
    to master in order to avoid a future regression.
    
    Cherry picked from commit f0616425995360c2506e99272152ffd9a4b10bd9


leifmadsen:
> This PR confuses me appears to only contain the linting changes, not any of the role binding or bridge controls mentioned in the description. I'm okay with the actual changes, just wondering if this was an oops of somesort,

That's the set of changes that were part of the cherry-pick commit message. I effectively duplicated the changes in this PR (which was oddly only merged into stable-1.4...), so the commit message itself is actually invalid. I should have actually modified the commit message to better call out what was expected.

I might still do that, but I'll have to force-push the changes here (which I think is ok, no one is going to be working again this PR). I'll make the commit message better reflect why these changes exist.

mrunge (approved):

PR #334, opened 2022-05-05
Disable deployment of Loki by default

leifmadsen:
Don't execute deployment of Loki or the Loki Operator and other systems
by default. Product deployment of STF does not include Loki or Loki
Operator, so this better aligns upstream system testing to what the
product deploys.


csibbitt (approved):

vyzigold (approved):

leifmadsen:
I did an A/B test of this with and without the `--extra-vars sgo_branch=rhbz2082310`. Cleared out all clusterroles and clusterrolebindings in between. Verified the proxy-auth cluster roles that would have been created by the Loki path were also removed.

Without changes in SGO, failure happens which is a success.
With changes in SGO, no failure happens which is the desired outcome.

Merging this down.

PR #338, opened 2022-05-11
Version updates for new CI features

csibbitt:
I revived upstream CI earlier today by rebuilding this container (possibly with a new module installation?).

It didn't work with my latest code from https://github.com/infrawatch/service-telemetry-operator/pull/336 though.

I needed kubernetes.core >= 2.2.0 which wasn't coming along with our current setup, so I found this solution that works.

This image has already been built and is live on the CI server, so I'm going to self-merge this, but wanted to draw attention to the tooling in use in upstream CI to support the latest features, namely:

Python 3.8 modules
--
ansible-core-2.12.5
kubernetes-23.6.0
openshift-0.13.1

Ansible Core 2.12 Collections
--
kubernetes.core:2.3.1
community.general:4.8.0

@elfiesmelfie , I'm not sure how this compares to your jumphost environment, but I know the jenkins job that I currently use to call your roles asks to install "ansible >= 2.9" which I think will do 2.10 at best (I haven't checked what it actually installs yet). I run 2.10 locally, but had trouble getting it to work in the Jenkins container, so just kept rolling forward with ansible-core. 

As long as what we end up with downstream CI works, I don't care too much if it matches what I've produced here. Notably, it's prety normal for upstream to be ahead by a couple versions so we get a heads-up of changes.

CC: @mgirgisf

### DIFF
--- a/.jenkins/agent/Dockerfile
+++ b/.jenkins/agent/Dockerfile
@@ -1,7 +1,8 @@
 FROM quay.io/openshift/origin-jenkins-agent-base:latest
 RUN curl -LO "https://github.com/operator-framework/operator-sdk/releases/download/v0.19.4/operator-sdk-v0.19.4-x86_64-linux-gnu" && \
     chmod +x operator-sdk-v0.19.4-x86_64-linux-gnu && mv operator-sdk-v0.19.4-x86_64-linux-gnu /usr/local/bin/operator-sdk
-RUN dnf install -y ansible golang && \
+RUN dnf install -y ansible golang python38 && \
     dnf groupinstall -y "Development Tools" -y && \
-    ansible-galaxy collection install community.kubernetes && \
-    python -m pip install openshift kubernetes
+    alternatives --set python /usr/bin/python3.8 && \
+    python -m pip install openshift kubernetes "ansible-core~=2.12" && \
+    ansible-galaxy collection install -f kubernetes.core community.general


.jenkins/agent/Dockerfile
```diff
@@ -1,7 +1,8 @@
 FROM quay.io/openshift/origin-jenkins-agent-base:latest
 RUN curl -LO "https://github.com/operator-framework/operator-sdk/releases/download/v0.19.4/operator-sdk-v0.19.4-x86_64-linux-gnu" && \
     chmod +x operator-sdk-v0.19.4-x86_64-linux-gnu && mv operator-sdk-v0.19.4-x86_64-linux-gnu /usr/local/bin/operator-sdk
```
csibbitt, line 3: I'm thinking we might be able to remove this entirely to prevent confusion. We have code in the CI ansible to manage operator-sdk (and golang for that matter), so AFAICT this version is never actually used.

.jenkins/agent/Dockerfile
```diff
@@ -1,7 +1,8 @@
...
-RUN dnf install -y ansible golang && \
+RUN dnf install -y ansible golang python38 && \
     dnf groupinstall -y "Development Tools" -y && \
-    ansible-galaxy collection install community.kubernetes && \
-    python -m pip install openshift kubernetes
+    alternatives --set python /usr/bin/python3.8 && \
+    python -m pip install openshift kubernetes "ansible-core~=2.12" && \
+    ansible-galaxy collection install -f kubernetes.core community.general
```
csibbitt, line 8: community.kubernetes has been replaced with kubernetes.core, and the `make` module we use in loki building (even if it's turned off) is now part of community.general which must be installed explicitly when using ansible-core

.jenkins/agent/Dockerfile
leifmadsen: I agree. If we can keep the installation of dynamic tooling separate of the base image, that should help keep confusion down.

In downstream we need to install the operator-sdk for the bundle generation, so if we stick to that in stf-run-ci, it should make more sense. I believe it's the `generate_bundle.sh` script that gets used across both upstream and downstream. I'm hoping to build some GitHub Actions that will generate bundles and push them to quay.io on merge to main and stable branches.

Then at least we have a path where the streams can more naturally cross.

As you said, it's natural for upstream CI systems to be more targeting the future, and for downstream to handle the stable. If the build system converges at some point, then I think that's a win.

csibbitt: @andyvesel Are you comfortable enough with how this works to take on the task of removing this operator-sdk installation and verifying that the jobs still function correctly? I think it's probably trivial, I just want to be sure you are comfortable with how this container is built for use, and checking on the results. Ping me if you decide to do it, I want to point you to an important checkbox in Jenkins for when we are updating the agent container.

elfiesmelfie:
> @elfiesmelfie , I'm not sure how this compares to your jumphost environment, but I know the jenkins job that I currently use to call your roles asks to install "ansible >= 2.9" which I think will do 2.10 at best (I haven't checked what it actually installs yet). I run 2.10 locally, but had trouble getting it to work in the Jenkins container, so just kept rolling forward with ansible-core.

I have added a requirements.yml for the ansible modules, and a task for the pip reqs, so these pip packages and ansible modules do end up on the jumphost

.jenkins/agent/Dockerfile
andyvesel: @csibbitt  I don't see anything non-trivial, so yes, I can figure this out. I will ping you for more details then

PR #342, opened 2022-06-02
Replacing lookup('k8s'... with k8s_info

csibbitt:
* Lookups always execute on the ansible control machine
* We think this will allow us to run this when `hosts: jumphost`

### DIFF
--- a/build/stf-run-ci/tasks/main.yml
+++ b/build/stf-run-ci/tasks/main.yml
@@ -26,14 +26,8 @@
   register: node_info
 
 - name: Find out if we are using crc by looking at the node hostnames
-  block:
-    - name: Try finding out for cluster with more than 1 node
-      set_fact:
-        is_crc: "{{ True if 'crc' in node_info.resources[0].metadata.labels[\"kubernetes.io/hostname\"] else False }}"
-  rescue:
-    - name: Try finding out for cluster with only 1 node
-      set_fact:
-        is_crc: "{{ True if 'crc' in node_info.resources.metadata.labels[\"kubernetes.io/hostname\"] else False }}"
+  set_fact:
+    is_crc: "{{ True if 'crc' in node_info.resources[0].metadata.labels[\"kubernetes.io/hostname\"] else False }}"
 
 - name: Clean up any existing global artifacts
   include_tasks: pre-clean.yml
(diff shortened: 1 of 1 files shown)


csibbitt:
test

csibbitt:
test

build/stf-run-ci/tasks/main.yml
```diff
@@ -20,19 +20,20 @@
...
   block:
     - name: Try finding out for cluster with more than 1 node
       set_fact:
-        is_crc: "{{ True if 'crc' in nodes[0].metadata.labels[\"kubernetes.io/hostname\"] else False }}"
+        is_crc: "{{ True if 'crc' in node_info.resources[0].metadata.labels[\"kubernetes.io/hostname\"] else False }}"
   rescue:
     - name: Try finding out for cluster with only 1 node
       set_fact:
-        is_crc: "{{ True if 'crc' in nodes.metadata.labels[\"kubernetes.io/hostname\"] else False }}"
+        is_crc: "{{ True if 'crc' in node_info.resources.metadata.labels[\"kubernetes.io/hostname\"] else False }}"
```
csibbitt, line 36: I didn't actually test single-node CRC detection, but I think this rescue block is no longer necessary because k8s_info returns an array of resources in either case.

mgirgisf (commented):
I didn't check its functionality , but I review from syntex point of view from ansible k8s_info module and we have kubernetes.core installed with the dockerfile .

build/stf-run-ci/tasks/main.yml
leifmadsen: I don't remember this rescue block and why it was necessary. I'm not sure what `nodes` data looked like vs `k8s_info` and the difference between them on single vs multi-node deployment. I'm fine with removing if tested.

leifmadsen (approved):
Looks good as-is to me. 👍🏻

build/stf-run-ci/tasks/main.yml
csibbitt: The lookup would return an array if there were multiple, but just a single hash if there was one.

csibbitt:
test

build/stf-run-ci/tasks/main.yml
leifmadsen: Ahhhh ok, thanks!

csibbitt:
> This needs to be backported to 1.4 as well. The same lookup block appears on that branch. The lookup was added after 1.3 , so the backport doesn't need to go that far

On it

csibbitt:
Backport: https://github.com/infrawatch/service-telemetry-operator/pull/346

PR #343, opened 2022-06-06
Adding a base_dir variable

csibbitt:
* So that shell scripts can be found when running remotely

### DIFF
--- a/build/stf-run-ci/tasks/main.yml
+++ b/build/stf-run-ci/tasks/main.yml
@@ -39,8 +39,7 @@
 
 - name: Set default base dir if not provided
   set_fact:
-    name: base_dir
-    value: "{{ playbook_dir }}"
+    base_dir: "{{ playbook_dir }}"
   when: base_dir | length == 0
 
 - name: Get new operator sdk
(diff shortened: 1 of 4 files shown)


csibbitt:
test

build/stf-run-ci/tasks/main.yml
```diff
@@ -37,9 +37,15 @@
   tags:
     - deploy
 
+- name: Set default base dir if not provided
+  set_fact:
+    name: base_dir
+    value: "{{ playbook_dir }}"
```
elfiesmelfie, line 43: This should be: 
 
```suggestion
    base_dir: "{{ playbook_dir }}"
```

You've set two facts here called `name` and `value` with the values `"base_dir"` and `playbook_dir`
base_dir remains unset and the functionality remains the same as before.

elfiesmelfie (changes requested):

build/stf-run-ci/tasks/main.yml
elfiesmelfie: I'm just going to commit this change to get it tested. I am >90% certain it'll resolve the error seen in the ci, which appeared to still be using the relative path.

csibbitt: Haha, how embarrassing! Thanks for the patch.

leifmadsen (approved):

PR #346, opened 2022-06-07
Replacing lookup('k8s'... with k8s_info (#342)

csibbitt:
* Lookups always execute on the ansible control machine
* We think this will allow us to run this when `hosts: jumphost`
* Remove unnecessary rescue block

Cherry picks changes from: 32f7e00


leifmadsen (approved):

csibbitt:
test

PR #350, opened 2022-06-09
Update to operator-sdk 1.11 on older branches

csibbitt:
I have no idea if this is okay or not, but I observed that this is different from the master branch and prevents me from using `operator-sdk run bundle` from CI automation when working from this branch. I'm opening this PR just to test the change and discuss whether it's safe or not.

### DIFF
--- a/build/stf-run-ci/defaults/main.yml
+++ b/build/stf-run-ci/defaults/main.yml
@@ -23,7 +23,7 @@ sto_image_tag: latest
 sg_core_image_tag: latest
 sg_bridge_image_tag: latest
 prometheus_webhook_snmp_image_tag: latest
-new_operator_sdk_version: v1.5.0
+new_operator_sdk_version: v1.11.0
 namespace: service-telemetry
 pull_secret_registry:
 pull_secret_user:


csibbitt:
test

leifmadsen:
I would almost be surprised if this works at all. I abandoned this due to the number of external changes required to get everything operational again, and since we're probably not going to be doing a lot of feature development here, I didn't see the value in continuing on this effort.

The one potential thing that might make this worthwhile is that `make bundle` becomes much simpler under this model than with how we use operator-sdk 0.19.4 (and obviously more in line with how other Operators are now built).

leifmadsen:
> I would almost be surprised if this works at all. I abandoned this due to the number of external changes required to get everything operational again, and since we're probably not going to be doing a lot of feature development here, I didn't see the value in continuing on this effort.
> 
> The one potential thing that might make this worthwhile is that `make bundle` becomes much simpler under this model than with how we use operator-sdk 0.19.4 (and obviously more in line with how other Operators are now built).

Disregard this comment. For some reason I thought this was a re-opening of my operator-sdk migration work. That is not the case at all.

leifmadsen (approved):

elfiesmelfie (approved):

PR #356, opened 2022-06-21
Rewrite Jenkinsfile to declarative syntax and skip CI on certain files

vyzigold:
This PR resolves the STF-618 and STF-576 issues.

I changed the syntax from scripted to declarative, which should make adding new steps to jenkinsfile a bit easier. I still needed to use quite a few "script" tags, because for example you can't interact with openshift from declarative directly. I read in the plugin readme, that they are looking for ways to allow that.

The CI works the same as before with these small changes:

 - When switching to declarative syntax, it automatically added a new stage at the beginning: "Declarative: Checkout SCM"pic-sel-220621-1731-55

 - When a stage fails. With the old Jenkinsfile, the stages after that would get executed and they would just return right after they started. Now these stages get skipped completely (except the cleanup stage), as can be seen on the following image (I deliberately edited a few files to fail each stage).
pic-sel-220621-1731-35

 - The naming scheme of ocp agents changed from "ocp-agent" to "{repo}-{branch}-{build number}-{some hash}" truncated to 64 character from the front. For example: "ce-telemetry-operator-jwysogla-declarative-jenkinsfile-35-lddv1". This might be useful for future debugging of the CI.

Another small change (the STF-576), I added a regex to the beginning of the Jenkinsfile. Each times the CI is started, it first checks which files changed and executes the pipeline only if one of the files matches the regex. So when we change the README for example, the CI just succeeds instantly.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,7 @@
 #!/usr/bin/env groovy
 
 
-def tested_files = "build/.*|deploy/.*|roles/.*|tests/smoketest/.*|Makefile|watches.yaml|Jenkinsfile"
+def tested_files = "\"build/.\\*|deploy/.\\*|roles/.\\*|tests/smoketest/.\\*|Makefile|watches.yaml|Jenkinsfile\""
 
 // can't just use BUILD_TAG because qdr operator limits name of resources to 60 chars
 def namespace = env.JOB_BASE_NAME + '-' + env.BUILD_NUMBER
@@ -90,15 +90,13 @@ pipeline {
 			defaultContainer 'exec'
 		}
 	}
+	environment {
+		run_ci = sh(script: "git fetch origin ${env.CHANGE_TARGET} && git diff --name-only origin/${env.CHANGE_TARGET} | egrep ${tested_files}", returnStatus: true)
+	}
 	stages {
 		stage('Clone Upstream') {
 			when {
-				anyOf {
-					changeset pattern: "${tested_files}", comparator: "REGEXP"
-					expression {
-						currentBuild.previousBuild.result != "SUCCESS" && currentBuild.previousBuild.result != null
-					}
-				}
+				environment name: 'run_ci', value: '0'
 			}
 			steps {
 				dir('service-telemetry-operator') {
@@ -119,14 +117,9 @@ pipeline {
 		}
 		stage('Create project') {
 			when {
+				environment name: 'run_ci', value: '0'
 				expression {
 					currentBuild.result == null
-				}
-				anyOf {
-					changeset pattern: "${tested_files}", comparator: "REGEXP"
-					expression {
-						currentBuild.previousBuild.result != "SUCCESS" && currentBuild.previousBuild.result != null
-					}
 				}
 			}
 			steps {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -82,79 +83,173 @@ spec:
...
+pipeline {
+	agent {
+		kubernetes {
+			inheritFrom 'ocp-agent'
+			defaultContainer 'exec'
+		}
+	}
+	stages {
+		stage('Clone Upstream') {
+			when {
```
csibbitt, line 95: Is there any way to avoid re-writing this condition in every stage?

csibbitt (commented):
In general this LGTM, but I'd like to see the logs output when it actually runs. So far I don't see this PR listed in Jenkins. I think something might be wrong with out polling or smee config.

Jenkinsfile
vyzigold: Unfortunately the "when" block must be inside a "stage" block, we can't move it outside and apply it for all stages at once. Possible solution might be to enclose all of our current stages into one top level stage and put the condition there. Something similar to the example there: https://www.jenkins.io/doc/book/pipeline/syntax/#sequential-stages

We would basically go from:
```
stages {
    stage('1') {
        when {
            condition
        }
        steps {
            ...
        }
    }
    stage('1') {
        when {
            condition
        }
        steps {
            ...
        }
    }
}
```
to:
```
stages {
    stage('top-stage') {
        when {
            condition
        }
        stages {
            stage('1') {
                steps {
                    ...
                }
            }
            stage('2') {
                steps {
                    ...
                }
            }
        }
    }
}
```
But I think the currentBuild.result == null might still be needed at the beginning of each stage.

vyzigold:
test

vyzigold:
Looks like the comment triggers don't work, I'll try it again and try to find the issue.

vyzigold:
test

vyzigold:
test

vyzigold:
test

vyzigold:
@csibbitt It passed the CI now, so take a look at the logs if you want. I also had to change the condition a little, it's a bit simpler now.

Jenkinsfile
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env groovy
 
 
-def tested_files = "build/.*|deploy/.*|roles/.*|tests/smoketest/.*|Makefile|watches.yaml|Jenkinsfile"
+def tested_files = "\"build/.\\*|deploy/.\\*|roles/.\\*|tests/smoketest/.\\*|Makefile|watches.yaml|Jenkinsfile\""
```
csibbitt, line 4: Did this actually fix it, or does it have the failure to detect changes on new branches as you described?

csibbitt (approved):

Jenkinsfile
vyzigold: That issue is fixed, but this isn't the line that fixed it. The condition before was just all wrong, it compared changes only with the previous build. That's why it worked for me when testing (I had some builds before I started working on it) and that's why it didn't work when creating a PR (there weren't any previous builds).

Now I made it actually look at the target branch of the PR and compare the changes with that, so it doesn't care about previous builds.

leifmadsen:
@vyzigold please make sure this is backported to stable-1.4, thanks!

PR #358, opened 2022-06-23
Sync the build/ directory from master

leifmadsen:


csibbitt (approved):
I don't know much about the loki changes, but if they work on master it's good enough for me!

PR #359, opened 2022-06-23
Sync roles/ and deploy/ directory from master

leifmadsen:


leifmadsen:
test

leifmadsen:
At the end of the day I think we should remove/export the Loki stuff entirely from this system, dropping it from the CRD etc. However it'll be much easier to manage all of that if everything is just in sync from this point forward.

csibbitt (approved):

PR #360, opened 2022-06-23
Add logging to perftest-dashboard (#310)

leifmadsen:
Cherry picked from commit f624687a6931ff939ff3cdb318384e5bd5386eb3


leifmadsen:
Another one of those changes that is likely to be short lived as we move the logging out of this repo, but to make the sync easier when that happens.

csibbitt (approved):

PR #368, opened 2022-09-13
Bundle changes to support deployment on OCP v4.10

leifmadsen:


leifmadsen:
Coordinated change with https://github.com/infrawatch/smart-gateway-operator/pull/123

I haven't done an import or change here yet. I need to do that as part of some documentation I plan to update when building the Operators and importing them into OCP 4.10 environments.

Currently marked as draft until I can complete full scope and testing to find out what I missed.

leifmadsen:
OK this should be ready for an initial review. I made enough changes here to get the Operators built and running on OCP 4.10 (also see the corresponding SGO repo linked in my previous comment).

Obviously this won't actually work right now, but the scope here is just to get the Operators and workloads building and loaded into OCP 4.10. Functional CI will definitely fail for now, but that is expected, and will be resolved once the certificate management is implemented (at least in theory until we find any potential issues I haven't previously accounted for).

leifmadsen:
test

leifmadsen:
Locally deployed using following environment setup. Got hints from `.jenkins/agent/Dockerfile` for extra dependencies.

```
ansible-galaxy collection install -f kubernetes.core community.general
ansible-playbook build/run-ci.yaml -e __service_telemetry_events_enabled=false
 -e __service_telemetry_storage_ephemeral_enabled=true -e sgo_branch=$(git branch --show-current)
 ```

That got me as far as have the Operators running and the CSVs deployed for STO and SGO on OCP 4.10.

Having some issues with Prometheus Operator starting that I'll have to look at next...

File: build/stf-run-ci/tasks/setup_base.yml [hunk 1]
```diff
@@ -26,20 +26,43 @@
...
+          - kubernetes
+
+  - name: Create openshift-cert-manager-operator OperatorGroup
+    k8s:
+      definition:
+        apiVersion: operators.coreos.com/v1
+        kind: OperatorGroup
+        metadata:
+          generateName: openshift-cert-manager-operator-
+          name: openshift-cert-manager-operator
```
csibbitt, line 48: I think you want either generateName OR name, not both.

https://docs.openshift.com/container-platform/4.10/rest_api/objects/index.html#io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta says "GenerateName is an optional prefix, used by the server, to generate a unique name ONLY IF the Name field has not been provided"

I tested this and found that the resource as shown is always called openshift-cert-manager-operator (generateName is ignored). If I remove the name field, a unique one is generated:

```
(ansible) [csibbitt@csibbitt rhos-qe-jenkins]$ oc get operatorgroup
NAME                                    AGE
openshift-cert-manager-operator         57s
openshift-cert-manager-operator-bql8b   4s
```
I'm not sure which you wanted (just the name is probably fine, I don't think we want multiple operatorGroups), but having both doesn't make sense.

csibbitt (approved):
Found a minor issue but it's not dangerous or worth blocking on.

build/stf-run-ci/tasks/setup_base.yml
leifmadsen: Yea what I wanted was the operatorgroup to be unique (not that it matters in our test rig) as that's how the OLM deployment would end up creating it. I was following this example from the official docs:

https://docs.openshift.com/container-platform/4.9/hardware_enablement/psap-special-resource-operator.html

Unfortunately I started with generateName only, and it doesn't work on a fresh deployment when the OperatorGroup doesn't already exist apparently (https://github.com/openshift/openshift-restclient-python/issues/279). So I added the name and it worked.

You're right though, a dynamically generated OperatorGroup name is unnecessary and not something we do in other parts of the playbook, so removal of the superfluous `generateName` is appropriate. I'll make the change since it's small and might as well be consistent.

I suspect it worked for you because you created the manifest directly using `oc create` via the `oc` CLI client? (or maybe because you already had the other operatorgroup?)

csibbitt: Yeah it was done with `oc create` so not affected by that openshift-restclient-python issue perhaps.

Just for science, I checked and it doesn't make a difference whether there is an existing one or not.

```
$ oc create -f -
apiVersion: operators.coreos.com/v1
kind: OperatorGroup
metadata:
    generateName: openshift-cert-manager-operator-
    namespace: default
spec: {}
operatorgroup.operators.coreos.com/openshift-cert-manager-operator-7dqbp created

$ oc get operatorgroup
NAME                                    AGE
openshift-cert-manager-operator-7dqbp   11s
$ 
```

elfiesmelfie (approved):
Looks okay to me, and seems consistent with the changes to SGO regarding versions. 
The additional ansible is what I expected i.e. adding cert  manager, and I assume that the config is correct.

PR #370, opened 2022-09-15
stf15/stf 1123 create certificates with cert manager

leifmadsen:
- Bundle changes to support deployment on OCP v4.10
- Switch Subscription from AMQ Cert Manager
- Bump the value of the OPERATOR_CSV_MAJOR_VERSION parameter
- Remove superflous generateName from OperatorGroup manifest
- Initial implementation of certificates for Interconnect


leifmadsen:
Going to close this and set a different base so that it's more clear what changes actually happened here. This is one of those situations where the `Depends-On` functionality in Gerrit would be really handy...

PR #371, opened 2022-09-15
Initial implementation of certificates for Interconnect

leifmadsen:
Provide the initial configuration for the certificate creation using
Cert Manager for OpenShift. Also provides the ConfigMap for the SASL
database which is no longer being generated in the deployment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,15 @@ The quickest way to start up Service Telemetry Framework for development is to
 run the `quickstart.sh` script located in the `deploy/` directory after starting
 up a [CodeReady Containers](https://github.com/code-ready/crc) environment.
 
+```shell
+crc setup
+crc config set memory 16384
+crc config set enable-cluster-monitoring true
+crc start
+crc console --credentials
+oc login -u kubeadmin https://api.crc.testing:6443
+```
+
 To deploy a local build of the Service Telemetry Operator itself, start by
 running `ansible-playbook build/run-ci.yaml`. If you have code to coordinate
 across the supporting InfraWatch repositories, you can pass the
--- a/README.md
+++ b/README.md
@@ -6,11 +6,11 @@ Framework.
 ## Getting Started
 
 You'll need to do the following steps in order to load the prerequisites for
-deploying to an OpenShift 4.x environment:
+deploying to an OpenShift 4.10 environment:
 
 * import catalog containing Service Telemetry and Smart Gateway Operators via
   OperatorSource file
-* install the AMQ Certificate Manager Operator before installing Service
+* install the Certificate Manager for OpenShift before installing Service
   Telemetry Operator
 * install the Service Telemetry Operator
 
@@ -60,14 +60,3 @@ ansible-playbook \
 
 Additional flags for overriding various branch and path names is documented in
 `build/stf-run-ci/README.md`.
-
-## CI
-
-### Travis
-
-* Runs OLM and Ansible linting
-
-### Hybrid DIY CI
-
-* We run an internal CI server that smoketests builds and publishes the results
-* WIP - Not fully implemented yet
--- a/deploy/olm-catalog/service-telemetry-operator/manifests/service-telemetry-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/service-telemetry-operator/manifests/service-telemetry-operator.clusterserviceversion.yaml
@@ -354,11 +354,17 @@ spec:
           verbs:
           - '*'
         - apiGroups:
+          - cert-manager.io
+          resources:
+          - issuers
+          - certificates
+          verbs:
+          - '*'
+        - apiGroups:
           - interconnectedcloud.github.io
           - smartgateway.infra.watch
           - monitoring.coreos.com
           - elasticsearch.k8s.elastic.co
-          - certmanager.k8s.io
           - integreatly.org
           - loki.grafana.com
           resources:
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -106,11 +106,17 @@ rules:
   verbs:
   - '*'
 - apiGroups:
+  - cert-manager.io
+  resources:
+  - issuers
+  - certificates
+  verbs:
+  - '*'
+- apiGroups:
   - interconnectedcloud.github.io
   - smartgateway.infra.watch
   - monitoring.coreos.com
   - elasticsearch.k8s.elastic.co
-  - certmanager.k8s.io
   - integreatly.org
   - loki.grafana.com
   resources:
--- a/roles/servicetelemetry/tasks/_local_signing_authority.yml
+++ b/roles/servicetelemetry/tasks/_local_signing_authority.yml
@@ -2,7 +2,7 @@
 - name: Create selfsigned Issuer
   k8s:
     definition:
-      apiVersion: certmanager.k8s.io/v1alpha1
+      apiVersion: cert-manager.io/v1
       kind: Issuer
       metadata:
         name: '{{ meta.namespace }}-selfsigned'
@@ -13,7 +13,7 @@
 - name: Create CA certificate
   k8s:
     definition:
-      apiVersion: certmanager.k8s.io/v1alpha1
+      apiVersion: cert-manager.io/v1
       kind: Certificate
       metadata:
         name: '{{ meta.namespace }}-ca'
@@ -28,7 +28,7 @@
 - name: Create namespace CA Issuer
   k8s:
     definition:
-      apiVersion: certmanager.k8s.io/v1alpha1
+      apiVersion: cert-manager.io/v1
       kind: Issuer
       metadata:
         name: '{{ meta.namespace }}-ca'
@@ -41,7 +41,7 @@
 - name: Create certificate for elasticsearch
   set_fact:
     elasticsearch_certificate_manifest:
-      apiVersion: certmanager.k8s.io/v1alpha1
+      apiVersion: cert-manager.io/v1
       kind: Certificate
       metadata:
         name: elasticsearch-es-http
--- a/roles/servicetelemetry/tasks/component_qdr.yml
+++ b/roles/servicetelemetry/tasks/component_qdr.yml
@@ -1,4 +1,147 @@
 ---
+- block:
+  - name: Create self-signed interconnect issuer
+    k8s:
+      state: '{{ "present" if servicetelemetry_vars.transports.qdr.enabled else "absent" }}'
+      definition:
+        apiVersion: certmanager.k8s.io/v1alpha1
+        kind: Issuer
+        metadata:
+          name: {{ meta.name }}-interconnect-selfsigned
+          namespace: {{ meta.namespace }}
+        spec:
+          selfSigned: {}
+
+  - name: Create self-signed interconnect certificate
+    k8s:
+      state: '{{ "present" if servicetelemetry_vars.transports.qdr.enabled else "absent" }}'
+      definition:
+        apiVersion: certmanager.k8s.io/v1alpha1
+        kind: Certificate
+        metadata:
+          name: {{ meta.name }}-interconnect-selfsigned
+          namespace: {{ meta.namespace }}
+        spec:
+          commonName: {{ meta.name }}-interconnect.{{ meta.namespace }}.svc.cluster.local
+          isCA: true
+          issuerRef:
+            name: {{ meta.name }}-interconnect-selfsigned
+          secretName: {{ meta.name }}-interconnect-selfsigned
+
+
+  - name: Create default CA interconnect issuer using self-signed interconnect certificate
+    k8s:
+      state: '{{ "present" if servicetelemetry_vars.transports.qdr.enabled else "absent" }}'
+      definition:
+        apiVersion: certmanager.k8s.io/v1alpha1
+        kind: Issuer
+        metadata:
+          name: {{ meta.name }}-interconnect-ca
+          namespace: {{ meta.namespace }}
+        spec:
+          ca:
+            secretName: {{ meta.name }}-interconnect-selfsigned
+
+  - name: Create OpenStack CA certificate using self-signed interconnect certificate
+    k8s:
+      state: '{{ "present" if servicetelemetry_vars.transports.qdr.enabled else "absent" }}'
+      definition:
+        apiVersion: cert-manager.io/v1
+        kind: Certificate
+        metadata:
+          name: {{ meta.name }}-interconnect-openstack-ca
+          namespace: {{ meta.namespace }}
+        spec:
+          commonName: {{ meta.name }}-interconnect-openstack-ca
+          isCA: true
+          issuerRef:
+            name: {{ meta.name }}-interconnect-selfsigned
+          secretName: {{ meta.name }}-interconnect-openstack-ca
+
+  - name: Create OpenStack credentials certificate using self-signed interconnect certificate
+    k8s:
+      state: '{{ "present" if servicetelemetry_vars.transports.qdr.enabled else "absent" }}'
+      definition:
+        apiVersion: cert-manager.io/v1
+        kind: Certificate
+        metadata:
+          name: {{ meta.name }}-interconnect-openstack-credentials
+          namespace: {{ meta.namespace }}
+        spec:
+          commonName: {{ meta.name }}-interconnect
+          dnsNames:
+          - {{ meta.name }}-interconnect.{{ meta.namespace }}.svc.cluster.local
+          issuerRef:
+            name: {{ meta.name }}-interconnect-ca
+          secretName: {{ meta.name }}-interconnect-openstack-credentials
+
+  - name: Create inter-router CA issuer
+    k8s:
+      state: '{{ "present" if servicetelemetry_vars.transports.qdr.enabled else "absent" }}'
+      definition:
+        apiVersion: cert-manager.io/v1
+        kind: Issuer
+        metadata:
+          name: {{ meta.name }}-interconnect-inter-router-ca
+          namespace: {{ meta.namespace }}
+        spec:
+          ca:
+            secretName: {{ meta.name }}-interconnect-inter-router-ca
+
+  - name: Create inter-router CA certificate
+    k8s:
+      state: '{{ "present" if servicetelemetry_vars.transports.qdr.enabled else "absent" }}'
+      definition:
+        apiVersion: cert-manager.io/v1
+        kind: Certificate
+        metadata:
+          name: {{ meta.name }}-interconnect-inter-router-ca
+          namespace: {{ meta.namespace }}
+        spec:
+          commonName: {{ meta.name }}-interconnect-inter-router-ca
+          isCA: true
+          issuerRef:
+            name: {{ meta.name }}-interconnect-selfsigned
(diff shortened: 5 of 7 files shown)


roles/servicetelemetry/tasks/component_qdr.yml
```diff
@@ -38,7 +181,7 @@
             linkCapacity: 25000
         interRouterListeners:
           - authenticatePeer: true
-            expose: true
+            expose: false
```
leifmadsen, line 184: I wanted to get this opened up and note early in this process that I want to change the behaviour here to simplify the certificate creation.

In STF 1.4 the `interRouterListeners` would be created and exposed as Route, meaning something like `default-interconnect-55671.apps.[external_host_URL]` would be exposed, and then references as a `dnsName` in the creation of the `default-interconnect-inter-router-credentials` certificate.

This complicates the values to be found in order to make sure the `dnsNames` list is updated with the external route URL. Because we don't have a use case where we expect to build inter-router mesh networks outside of the current namespace, I don't believe setting `expose: true` is appropriate, and in theory this could be considered a hardening tactic (although I did this to reduce the needed logic to find and populate the appropriate inter-router Route creation dynamically).

roles/servicetelemetry/tasks/component_qdr.yml
```diff
@@ -1,4 +1,147 @@
...
+        namespace: "{{ meta.namespace }}"
+        labels:
+          application: "{{ meta.name }}-interconnect"
+          interconnect_cr: "{{ meta-name }}-interconnect"
+      data:
+        qdrouterd.conf: |
+          pwcheck_method: auxprop
+          auxprop_plugin: sasldb
+          sasldb_path: /tmp/qdrouterd.sasldb
+  when: interconnect_manifest is not defined
```
leifmadsen, line 142: @csibbitt could I get another set of eyes to review if this makes? This is doing exactly what the Interconnect router results in through it's requests to AMQ Cert Manager, so this matches 1:1 what would have happened on STF 1.4 and OCP 4.8.

However now is our chance to adjust anything we think might look less than ideal. I keep looking at this sequence and something feels off to me. If I go through this, it seems like...

self-signed issuer references:
- selfsigned-ca cert
- openstack-ca cert
- inter-router-ca cert

interconnect-ca issuer references:
- ca-secret: selfsigned-ca cert
-- creates: openstack-credentials cert

inter-router-ca issuer references:
- ca-secret: inter-router-ca cert
-- creates: inter-router-credentials cert

So it seems that the `openstack-ca` certificate is never used. 

The openstack credentials certificate uses the interconnect-ca issuer.
The inter-router credentials certificate uses the inter-router-ca- issuer.

If we were going to clean this up, it seems like we'd create an openstack-ca issuer that used `ca.secretName` of `openstack-ca` certificate, then create the openstack-credentials certificate with an `issuerRef: ...-openstack-ca`.

How close am I? :)

build/metadata.sh
```diff
@@ -13,7 +13,7 @@ REQUIRED_OPERATOR_SDK_VERSION=${REQUIRED_OPERATOR_SDK_VERSION:-v0.19.4}
 SERVICE_TELEMETRY_SUBSCRIPTION=${SERVICE_TELEMETRY_SUBSCRIPTION:-service-telemetry-operator-stable-infrawatch-operators-openshift-marketplace}
 OPERATOR_IMAGE=${OPERATOR_IMAGE:-"quay.io/infrawatch/${OPERATOR_NAME}"}
 OPERATOR_TAG=${OPERATOR_TAG:-latest}
-OPERATOR_CSV_MAJOR_VERSION=${OPERATOR_CSV_MAJOR_VERSION:-1.4}
+OPERATOR_CSV_MAJOR_VERSION=${OPERATOR_CSV_MAJOR_VERSION:-1.5}
```
leifmadsen, line 16: As we've done in the past, bumping major version of the Operator being created

build/stf-run-ci/tasks/setup_base.yml
```diff
@@ -26,20 +26,42 @@
...
-      metadata:
-        name: amq7-cert-manager-operator
-        namespace: openshift-operators
-      spec:
-        channel: 1.x
-        installPlanApproval: Automatic
-        name: amq7-cert-manager-operator
-        source: redhat-operators
-        sourceNamespace: openshift-marketplace
+- block:
```
leifmadsen, line 29: I was unable to get the Operator to install without pre-creating the namespace and OperatorGroup, which is why this is more verbose than the previous subscription, which used a pre-created namespace.

leifmadsen:
At this point in the review this mostly seems to work, but I'm having an issue in my environment right now where it seems like STO skips the QDR creation steps, along with some other issues with launching other requested workloads from Operators like Prometheus. I'm going to assume it's my local environment for now which is old, and will rebuild this week to re-test. If I continue to see problems I will need to debug further as I wasn't getting those issues on OCP 4.8 with the same code (no changes were made in the areas that are failing). This is why I have the do-not-merge item on right now.

These are changes that would do well to run through the smoketests and other CI paths, as it may catch other changes required to be made to make sure testing is anticipating the appropriate values and objects.

leifmadsen:
test

leifmadsen:
test

deploy/olm-catalog/service-telemetry-operator/manifests/service-telemetry-operator.clusterserviceversion.yaml
```diff
@@ -353,12 +353,18 @@ spec:
...
+          - issuers
+          - certificates
+          verbs:
+          - '*'
         - apiGroups:
           - interconnectedcloud.github.io
           - smartgateway.infra.watch
           - monitoring.coreos.com
           - elasticsearch.k8s.elastic.co
-          - certmanager.k8s.io
```
leifmadsen, line 361: Removed access to old AMQ Cert Manager API interface.

deploy/role.yaml
```diff
@@ -105,12 +105,18 @@ rules:
   - statefulsets
   verbs:
   - '*'
+- apiGroups:
```
leifmadsen, line 108: I broke this out into it's own section and limited access to the 2 current `cert-manager.io` resources we need access to, which is `Certificates` and `Issuers`. We'll need to extend this if we ever do more elaborate certificate management.

deploy/olm-catalog/service-telemetry-operator/manifests/service-telemetry-operator.clusterserviceversion.yaml
```diff
@@ -353,12 +353,18 @@ spec:
           - statefulsets
           verbs:
           - '*'
+        - apiGroups:
+          - cert-manager.io
+          resources:
+          - issuers
+          - certificates
+          verbs:
+          - '*'
```
leifmadsen, line 362: This block is generated / updated in the CSV as sourced from `deploy/role.yaml` via `operator-sdk`. This is not something that should be manually applied by hand.

```
operator-sdk generate bundle --channels unstable --default-channel --manifests --metadata
```

roles/servicetelemetry/tasks/_local_signing_authority.yml
```diff
@@ -2,7 +2,7 @@
 - name: Create selfsigned Issuer
   k8s:
     definition:
-      apiVersion: certmanager.k8s.io/v1alpha1
+      apiVersion: cert-manager.io/v1
```
leifmadsen, line 5: Also updating for certificate creation for the Elasticsearch deployment. Honestly this should probably just be moved to the `component_elasticsearch.yml` and put into a `block` much like I did for the `component_qdr.yml`. It's unnecessary though, and I would only do this as a follow up later since it's not required to provide minimal changes.

roles/servicetelemetry/tasks/component_qdr.yml
```diff
@@ -1,4 +1,147 @@
...
+        namespace: "{{ meta.namespace }}"
+        labels:
+          application: "{{ meta.name }}-interconnect"
+          interconnect_cr: "{{ meta.name }}-interconnect"
+      data:
+        qdrouterd.conf: |
+          pwcheck_method: auxprop
+          auxprop_plugin: sasldb
+          sasldb_path: /tmp/qdrouterd.sasldb
+  when: interconnect_manifest is not defined
```
leifmadsen, line 142: I guess this is part of the `generate_...` parameters normally provided to `Interconnect` manifests so we need to generate this now.

roles/servicetelemetry/tasks/main.yml
```diff
@@ -12,9 +12,9 @@
   set_fact:
     has_interconnect_api: "{{ True if 'interconnectedcloud.github.io' in api_groups else False }}"
 
-- name: Check if we have certmanager.k8s.io API
+- name: Check if we have cert-manager.io API
```
leifmadsen, line 15: I missed this the first time around and was wondering why my Interconnects weren't be created. This fixed it :)

leifmadsen:
Local testing verification (waiting on update of CI agent to provide deployment dependencies):

```
./build/validate_deployment.sh 

* [info] Waiting for QDR deployment to complete

deployment "default-interconnect" successfully rolled out

* [info] Waiting for prometheus deployment to complete

statefulset rolling update complete 1 pods at revision prometheus-default-d49b8757...

* [info] Waiting for elasticsearch deployment to complete 


* [info] Waiting for alertmanager deployment to complete

statefulset rolling update complete 1 pods at revision alertmanager-default-77546f7cdd...

* [info] Waiting for smart-gateway deployment to complete

deployment "default-cloud1-coll-meter-smartgateway" successfully rolled out
deployment "default-cloud1-coll-event-smartgateway" successfully rolled out
deployment "default-cloud1-ceil-event-smartgateway" successfully rolled out
deployment "default-cloud1-ceil-meter-smartgateway" successfully rolled out

* [info] Waiting for all pods to show Ready/Complete


* [info] CI Build complete. You can now run tests.
```

```
./tests/smoketest/smoketest.sh

[...snip...]

*** [INFO] Cleanup resources...
job.batch "stf-smoketest-smoke1" deleted

*** [SUCCESS] Smoke test job completed successfully
```

```
oc get stf default -oyaml
apiVersion: infra.watch/v1beta1
kind: ServiceTelemetry
metadata:
  creationTimestamp: "2022-09-20T16:01:07Z"
  generation: 2
  name: default
  namespace: service-telemetry
  resourceVersion: "76543"
  uid: 6d43f227-72ab-471f-8782-ee7e139d76ca
spec:
  alerting:
    alertmanager:
      receivers:
        snmpTraps:
          enabled: true
      storage:
        strategy: ephemeral
  backends:
    events:
      elasticsearch:
        enabled: true
        storage:
          strategy: ephemeral
    logs:
      loki:
        enabled: false
        flavor: 1x.extra-small
        replicationFactor: 1
        storage:
          objectStorageSecret: test
    metrics:
      prometheus:
        enabled: true
        storage:
          strategy: ephemeral
  highAvailability:
    enabled: false
  observabilityStrategy: use_community
status:
  conditions:
  - lastTransitionTime: "2022-09-20T17:24:31Z"
    reason: Successful
    status: "True"
    type: Running
```

I also checked the `oc logs --selector=name=service-telemetry-operator -f` to make sure there were no errors (that's what I did to find the various bugs that I fixed in the most recent set of commits) and that the reconciliation loop eventually did reconcile (no infinite loops).

I have _not_ done a fully functional set of tests and verified that dashboarding works, that data consumption works, or that our commands for exporting the CA certificate and importing into OSP works as designed an e2e test. This only covers the deployment of STF and certificates in order to get a fully operational system.

csibbitt:
test

roles/servicetelemetry/tasks/component_qdr.yml
csibbitt: I endorse this change. I'm pretty sure this route was for the Cloudforms-compatible OSP-controller-mesh-mode feature, so that the OSP-side mesh would connect to STF in inter-router mode and not require additional edge nodes. AFAIK this is no longer supported.

roles/servicetelemetry/tasks/component_qdr.yml
csibbitt: The interconnect-openstack-ca cert is "used" by the edge-listener L36 / 179 `sslProfile: openstack` which is defined on L51 / 194 and uses that cert. We'd need to experiment to figure out whether operationally this is "really used for anything", but AFAICT it's really just a "dummy" in this case (more below...)

I followed through your breakdown and it otherwise reads correctly, though I don't agree with your suggestion about the openstack-ca Issuer. I don't think it would make anything cleaner or confer any other advantage. Hopefully I can clarify more below...

TLDR; I don't think we really need to make any changes here, but some cleanup is likely possible.

To help me get this straight in my head I've cut down the code and arranged each certificate under it's issuer, similar to what you did above but I find it slightly easier to read.

```
kind: Issuer + Certificate (uses cert issued by self)
    name: "{{ meta.name }}-interconnect-selfsigned" 
kind: Certificate
    name: "{{ meta.name }}-interconnect-openstack-ca"
kind: Certificate
    name: "{{ meta.name }}-interconnect-inter-router-ca"

kind: Issuer (uses the same self-signed cert as above)
    name: "{{ meta.name }}-interconnect-ca" 
kind: Certificate
    name: "{{ meta.name }}-interconnect-openstack-credentials"

kind: Issuer (uses own cert issued by interconnect-selfsigned)
    name: "{{ meta.name }}-interconnect-inter-router-ca" 
kind: Certificate
    name: "{{ meta.name }}-interconnect-inter-router-credentials"      
```

There is one self-signed certificate in the system, `interconnect-selfsigned` which serves as the root of all trust. For some reason it's used to sign certs under two different "Issuers", itself and the `interconnect-ca` Issuer, possibly meant to be an "Intermediate CA", but usually an Intermediate CA has it's own cert signed by the root (like the inter-router one does). I'm not exactly sure what happened here, but it doesn't seem like a serious problem, and I think it MAY have been done this way to help simplify the OSP setup, see more below...

The `interconnect-ca` Isuer issues an `openstack-credentials` certificate which is used in the edge-listener's SSL profile. This cert is presented to clients that connect. You can check by comparing the outputs: 
* `oc get secret default-interconnect-openstack-credentials -o json | jq -r '.data["tls.crt"]' | base64 -d | openssl x509 -in - -text | less`
* `openssl s_client -servername default-interconnect-5671-stf15.apps.yourcluster -connect default-interconnect-5671-stf15.apps.yourcluster:443 2>/dev/null | openssl x509 -text | less`

The self-signed cert is also used to issue a cert to the more conventional looking (IMO) Intermediate CA called `interconnect-inter-router-ca` which is then used to generate and validate credentials for TLS mutual-auth connections between mesh members. Easy enough, very normal looking. Self-contained mesh-mode stuff. This should not change.

The self-signed cert is ALSO used to sign the `interconnect-openstack-ca` cert, which seems a bit weird to me. This is used by the edge-listener to validate client certificates presented during TLS mutual-auth, which is not something we currently do. I expect it might not work out of the box with a generated cert like this, anyways. I think in the TLS mutual-auth scenario we'd import this certificate from OSP, or from whatever CA issued the certs that OSP clients will present, rather than asking for it to be generated. AFAICT (haven't experimented yet), this cert looks like filler (since we don't use mutual-auth). We need to give the listener "something", but it's possible we could remove this cert and tell it to use `interconnect-openstack-credentials` as both the "caCert" and the "credentials" instead. I'm not sure if that's any less confusing, but it matches an example from the QDR docs so maybe it's not that crazy? https://access.redhat.com/documentation/en-us/red_hat_amq/7.6/html-single/deploying_amq_interconnect_on_openshift/index#creating-inter-cluster-router-network-router-ocp (step 6)

Finally, we export that self-signed certificate to OSP so it can verify the edge-listener's ceritificate (`openstack-credentials`) when it connects. This should not change.

So based on this, the only "Improvements" I can think to make are the following:

* Create an additional `interconnect-ca` Certificate issued by `interconnect-selfsigned`, to be used by the `interconnect-ca` Issuer so that it looks more like a normal intermediate CA 
  * This would likely alter our OSP connection procedures, as we'd either need to provide this cert to OSP at config-time, or have it bundled into the `interconnect-openstack-credentials` cert that the listener presents (either one of these steps would let openstack validate the credentials cert).
  * It's possible there is some complication here that lead to the same cert intentionally being used by the both the root and "intermediate" CAs
* (Or, better) Eliminate `interconnect-ca` entirely and issue `openstack-credentials` from the root. Everything should continue to work "as is" in this case AFAICT, just with one less Issuer to confuse us
* Rename `interconnect-openstack-ca` to `dummy-interconnect-openstack-ca`
  * (Or, better?) eliminate it entirely

Since this stuff is quite confusing to look at, eliminating the superfluous parts to help simplify things might be worth a try.

csibbitt (approved):
AFAICT This code is good to merge as-is, though I wrote a whole book explaining how we might be able to eliminate one Issuer and one Certificate.

csibbitt:
I've built and deployed from this branch (as of 6e82569) to my quicklab and it passed smoke tests. No surprises. I have not yet connected an Openstack to it, though I'm setting up to do it from a Wallaby cloud. Based on what I see, I doubt there will be any changes at all to the procedure.

README.md
```diff
@@ -47,6 +47,15 @@ The quickest way to start up Service Telemetry Framework for development is to
 run the `quickstart.sh` script located in the `deploy/` directory after starting
 up a [CodeReady Containers](https://github.com/code-ready/crc) environment.
 
+```shell
+crc setup
+crc config set memory 16384
+crc config set enable-cluster-monitoring true
```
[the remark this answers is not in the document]
leifmadsen: I did test without it, and the Prometheus deployment fails as we request access to the `prometheus-k8s` service account, which is created as part of the openshift-monitoring deployment. Without cluster monitoring enabled, the Prometheus instance we deploy fails to start due to missing volume.

roles/servicetelemetry/tasks/component_qdr.yml
leifmadsen: > I can't comment on L53 / 196 below, but I'm thinking we need to remove those generate* properties?

Oops! Did I not disable them? Definitely need to do that (although the Interconnect Operator knows that the `certmanager.k8s.io` API doesn't exist, and will say that it has the inability to create certificates, so it seems to handle the Operator / CRD missing gracefully, but we should definitely be explicit in our `generate...` configuration. Nice catch, will fix that tomorrow.

For the SASL stuff, I believe it comes from the `interRouterListeners` configuration, which in `component_qdr.yml` on L186 has `saslMechanisms: EXTERNAL`. As a result the Interconnect workload would fail to launch due to missing ConfigMap, so I just made sure it was available as it was previously in STF 1.4.

I haven't looked at the code, but my theory is that since the AMQ Cert Manager (certmanager.k8s.io) isn't available, the code path that would have generated this ConfigMap in Interconnect Operator is not run, and results in the workload unable to load a volume it is expecting. Adding the ConfigMap here results in the router workloads coming up as intended.

leifmadsen:
test

roles/servicetelemetry/tasks/component_qdr.yml
leifmadsen: I have changed the `generate_...` commands to be `false`.

leifmadsen:
test

PR #374, opened 2022-09-20
Fix casc template

csibbitt:
Repos were not scanning when deploying from this template because the `$` in `java.util.Arrays$ArrayList` was triggering an empty expansion of `$ArrayList`. This uses the `$$` sequence and a non-existent envvar to escape the `$`  (https://stackoverflow.com/a/61259844)


leifmadsen (approved):
groovy is so weird...

PR #376, opened 2022-09-26
Allow forward slashes in working branch names

leifmadsen:
When a branch is named something like 'enhance/stf-1234' the 'working_branch' value will be invalid as only the first field delimited by a forward-slash will be used. This change increases the range of fields from 3 and onward.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -103,11 +103,11 @@ pipeline {
 					catchError(buildResult: 'FAILURE', stageResult: 'FAILURE') {
 						checkout scm
 						script {
-							working_branch = sh(script: 'git ls-remote --heads origin | grep $(git rev-parse HEAD) | cut -d / -f 3', returnStdout: true).toString().trim()
+							working_branch = sh(script: 'git ls-remote --heads origin | grep $(git rev-parse HEAD) | cut -d / -f 3-', returnStdout: true).toString().trim()
 							if (!working_branch) {
 								// in this case, a merge with the base branch was required thus we use the second to last commit
 								// to find the original topic branch name
-								working_branch = sh(script: 'git ls-remote --heads origin | grep $(git rev-parse HEAD~1) | cut -d / -f 3', returnStdout: true).toString().trim()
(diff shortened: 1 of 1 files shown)


leifmadsen:
test

leifmadsen:
I believe this is working now. Previously I would have just seen `ci` in the base branch name while doing the checkout/rescue of the same-named branches. In this case I now see the full name of the branch:

```
fatal: [localhost]: FAILED! => {"changed": false, "cmd": "/usr/bin/git checkout --force ci/bug_in_working_branch", "msg": "Failed to checkout ci/bug_in_working_branch", "rc": 1, "stderr": "error: pathspec 'ci/bug_in_working_branch' did not match any file(s) known to git\n", "stderr_lines": ["error: pathspec 'ci/bug_in_working_branch' did not match any file(s) known to git"], "stdout": "", "stdout_lines": []}
```

csibbitt (approved):
lgtm

leifmadsen:
test

leifmadsen:
Not really sure why the CI is failing. Seems to not be getting SNMP deployed which fails the smoketest. Will have to try and reproduce locally.

leifmadsen:
This really has nothing to do with the smoketests so not really sure how much debugging to do here, but I suppose we gotta get this passing, so I'll take a look locally and see if this branch results in a failure?

leifmadsen:
I can't reproduce this error locally. I went through the `Jenkinsfile` while on this branch locally, working against a quicklab instance, and the smoketests pass locally. This seems like something in the CI environment then? I worked through the Jenkinsfile and ran commands as the server should be testing this (separate `ServiceTelemetry` deployment using the same inline manifest as found in the `Jenkinsfile`.

Tempted to just merge this as the changes are really at the Jenkins level and should have no affect on the actual deployment.

leifmadsen:
I'm just merging this because it shouldn't be affecting the smoketests at all, and I've done a full reproduction with quicklab using the same process the Jenkinsfile provides, and all smoketests etc pass. I'll have to continue debugging this if we keep running into it, but I don't see how this is failing the deployment so late into the game, and with external scripts. 🤠

PR #378, opened 2022-09-28
Infrared test script for OSP17.0

csibbitt:
This is a copy of the 16.2 directory with changes required to make it work.

This diff tells a better story about what exactly changed:  https://paste.opendev.org/show/bdgvwGJHuaCSqe8CVChy/

- Updated VM_IMAGE to RHEL 9.0
- OSP_VERSION to 17
- Certificate handling like in OSP 13
- Bootmode UEFI [1]
- Network option changes - dvr, bgpvpn, and l2gw all forced off [1]
- VBMC option changes [1]
- NeutronBridemappings [1]
- `extra-vars osp_version=17.0` [1]

[1] These changes were copied from our downstream Jenkins jobs. I'm not 100% sure which of them are strictly required, but at least one of them is.

### DIFF
--- a/tests/infrared/17.0/enable-stf.yaml.template
+++ b/tests/infrared/17.0/enable-stf.yaml.template
@@ -1,0 +1,85 @@
+# enable data collection that is compatible with the STF data model
+# enablement of the collectors and transport are done with separate enviroment files
+# recommended:
+# - environments/metrics/ceilometer-qdr-write.yaml
+# - environments/metrics/collectd-qdr-write.yaml
+# - environments/metrics/qdr-edge-only.yaml
+
+---
+tripleo_heat_templates:
+    []
+
+custom_templates:
+    # matches the documentation for enable-stf.yaml in stable-1.3 documentation
+    parameter_defaults:
+        # only send to STF, not other publishers
+        EventPipelinePublishers: []
+        PipelinePublishers: []
+
+        # manage the polling and pipeline configuration files for Ceilometer agents
+        ManagePolling: true
+        ManagePipeline: true
+
+        # enable Ceilometer metrics and events
+        CeilometerQdrPublishMetrics: true
+        CeilometerQdrPublishEvents: true
+
+        # enable collection of API status
+        CollectdEnableSensubility: true
+        CollectdSensubilityTransport: amqp1
+
+        # enable collection of containerized service metrics
+        CollectdEnableLibpodstats: true
+
+        # set collectd overrides for higher telemetry resolution and extra plugins
+        # to load
+        CollectdConnectionType: amqp1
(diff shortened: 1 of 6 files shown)


tests/infrared/17.0/infrared-openstack.sh
```diff
@@ -0,0 +1,211 @@
...
+      --network-dvr no \
+      --network-l2gw no \
+      --storage-backend ceph \
+      --storage-external no \
+      --overcloud-ssl no \
+      --introspect yes \
+      --tagging yes \
+      --deploy yes \
+      --ntp-server "${NTP_SERVER}" \
+      --overcloud-templates ceilometer-write-qdr-edge-only,outputs/enable-stf.yaml,outputs/stf-connectors.yaml,outputs/gnocchi-connectors.yaml \
```
paramite, line 128: I guess it probably works, but just out of curiosity, the "ceilometer-write-qdr-edge-only" will be created where / when?

csibbitt: https://github.com/redhat-openstack/infrared/blob/master/plugins/tripleo-overcloud/vars/overcloud/templates/ceilometer-write-qdr-edge-only.yml

tests/infrared/17.0/enable-stf.yaml.template
```diff
@@ -0,0 +1,85 @@
+# enable data collection that is compatible with the STF data model
+# enablement of the collectors and transport are done with separate enviroment files
+# recommended:
+# - environments/metrics/ceilometer-qdr-write.yaml
+# - environments/metrics/collectd-qdr-write.yaml
+# - environments/metrics/qdr-edge-only.yaml
+
+---
+tripleo_heat_templates:
+    []
```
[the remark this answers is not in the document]
leifmadsen: I can explain [3]: reason is that the default template strategy is "merge" and so if you include the collectd environment file and then deploy for multi-cloud (or a differently named cloud configuration) you get problems because both the default connector and your newly added connector are configured, so collectd will try to send to an address without any QDR listeners.

leifmadsen (approved):
Won't have time to independently verify this, but looking through this I don't spot anything that looks off.

tests/infrared/17.0/enable-stf.yaml.template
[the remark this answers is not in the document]
paramite: Thanks guys. Yes Chris, it answers my question.

tests/infrared/17.0/infrared-openstack.sh
paramite: But isn't ".yaml" missing there?

PR #379, opened 2022-10-13
[DNM] Testing only

leifmadsen:
Test PR to help validate https://github.com/infrawatch/prometheus-webhook-snmp/pull/10

PR #381, opened 2022-10-18
Move the SNMP trap delivery checks

leifmadsen:
Move the SNMP trap delivery checks as where they are situated now seems
to cause false positives. Moves the checks closer to the end of the
smoketest run seems to result in a better change that the logs the check
is looking for have been provided.

### DIFF
--- a/tests/smoketest/smoketest.sh
+++ b/tests/smoketest/smoketest.sh
@@ -64,7 +64,6 @@ oc run curl --restart='Never' --image=quay.io/infrawatch/busyboxplus:curl -- sh 
 # it takes some time to get the alert delivered, continuing with other tests
 
 
-
 # Trying to find a less brittle test than a timeout
 JOB_TIMEOUT=300s
 for NAME in "${CLOUDNAMES[@]}"; do
@@ -73,10 +72,8 @@ for NAME in "${CLOUDNAMES[@]}"; do
     RET=$((RET || $?)) # Accumulate exit codes
 done
 
+echo "*** [INFO] Clean up the curl pod used for delivery of test alert"
 oc delete pod curl
-SNMP_WEBHOOK_POD=$(oc get pod -l "app=default-snmp-webhook" -ojsonpath='{.items[0].metadata.name}')
-oc logs "$SNMP_WEBHOOK_POD" | grep 'Sending SNMP trap'
-SNMP_WEBHOOK_STATUS=$?
 
 echo "*** [INFO] Showing oc get all..."
 oc get all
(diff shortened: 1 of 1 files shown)


leifmadsen:
test

tests/smoketest/smoketest.sh
```diff
@@ -130,17 +127,24 @@ echo "*** [INFO] Logs from alertmanager..."
...
+oc logs "$SNMP_WEBHOOK_POD" | grep 'Sending SNMP trap'
+SNMP_WEBHOOK_STATUS=$?
 echo
 
 if [ $SNMP_WEBHOOK_STATUS -ne 0 ]; then
     echo "*** [FAILURE] SNMP Webhook failed"
     exit 1
 fi
 
+echo "*** [INFO] Cleanup resources..."
```
csibbitt, line 142: Why did this move? The jobs it cleans up aren't related to the webhook pod, and now I believe that if the tests fail for basically any reason that would break SNMP, the script will `exit 1` and not do the cleanup.

csibbitt (approved):
This is still going to be brittle, but since it will improve the situation without incurring new large waits, I think it's worth merging.  Suspicious about moving the cleanup; I don't think it's dangerous since CI tests start from a clean namespace anyways, but might be a PITA for local testing?

tests/smoketest/smoketest.sh
```diff
@@ -130,17 +127,24 @@ echo "*** [INFO] Logs from alertmanager..."
...
+echo "*** [INFO] Check status of SNMP webhook trap delivery"
+SNMP_WEBHOOK_POD=$(oc get pod -l "app=default-snmp-webhook" -ojsonpath='{.items[0].metadata.name}')
+oc logs "$SNMP_WEBHOOK_POD" | grep 'Sending SNMP trap'
+SNMP_WEBHOOK_STATUS=$?
 echo
 
 if [ $SNMP_WEBHOOK_STATUS -ne 0 ]; then
     echo "*** [FAILURE] SNMP Webhook failed"
     exit 1
 fi
```
leifmadsen, line 140: This might not even still be enough. In my other testing (https://github.com/infrawatch/service-telemetry-operator/pull/379/files#diff-16335c82c44304ee49855ecebf854c2a8db384a1240f299c5f023b4b9845e407R135-R151) I placed the checks after the clean up, which itself probably has some extra delay built in. It might make more sense to loop this with set of timeouts and max loop count before deciding to ultimately fail this.

leifmadsen:
I've created a loop check and tried to avoid moving anything really around. The exit will still happen at the end even though we will know it's marked as a failure earlier, but this keeps most of the logic and flow the same. I don't have a local system to test this against so I'll rely on the CI to test for me. If it fails with syntax problems I'll have to stand up an environment and test this rather than hacking multiple runs here.

csibbitt (approved):

PR #384, opened 2022-10-20
Move the SNMP trap delivery checks (#381)

leifmadsen:
* Move the SNMP trap delivery checks

Move the SNMP trap delivery checks as where they are situated now seems
to cause false positives. Moves the checks closer to the end of the
smoketest run seems to result in a better change that the logs the check
is looking for have been provided.

* Use a loop to check for SNMP status with break and max time

(cherry picked from commit 1dc0808ace835f2410c35b1b523398ed35f797f8)

### DIFF
--- a/tests/smoketest/smoketest.sh
+++ b/tests/smoketest/smoketest.sh
@@ -75,8 +75,18 @@ done
 
 oc delete pod curl
 SNMP_WEBHOOK_POD=$(oc get pod -l "app=default-snmp-webhook" -ojsonpath='{.items[0].metadata.name}')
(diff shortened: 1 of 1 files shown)


leifmadsen:
test

leifmadsen:
As far as I can tell this timed out after 300 seconds. It looks like everything was on its way to passing, but the job basically exited after taking too long. Going to try this one more time, then I might have to try increasing the timeout to see if that's what I'm running into, or if there is really a problem. Going through the logs it doesn't appear at first glance there is an actual issue though other than it taking a long time.

leifmadsen:
test

PR #385, opened 2022-10-20
Lock the bundle to OCP v4.10

leifmadsen:


leifmadsen:
Unfortunately I'm not really sure exactly how I can test this other than to make sure things don't result in failure here. This label is primarily for downstream product release pipeline. I tried to test this by making a direct comment, but that was reverted by the pipeline during the build process. Looks like I need to land this, cherry pick it back to stable-1.5, and then tag a new beta build, followed by a poll-upstream-sources to result in a check-patch and check-merge job to land the changes in dist-git where they won't be reverted.

A bit of a long run for a test unfortunately. I'm open to other ideas on how I can pre-test this before landing and tagging which unfortunately is a bit of a shot in the dark.

I looked through [this](https://docs.openshift.com/container-platform/4.10/operators/operator_sdk/osdk-working-bundle-images.html#osdk-control-compat_osdk-working-bundle-images) documentation, but unfortunately it doesn't explicitly call out a locked version example. I'm going by some internal communication that stated we should use a leading `=` on the label.

leifmadsen:
test

leifmadsen:
Failure seemed to be on the host itself vs error with the test. Going to try another run.

leifmadsen:
test

leifmadsen:
test

leifmadsen:
test

leifmadsen:
A bit of an older guide, but shows the syntax is correct. Going to merge and backport this so it can be imported for testing.

PR #388, opened 2022-11-10
Force SGO checkout during build

csibbitt:
* Replacing the placeholder namespace during the build results in a "there are local changes" error on next build
* This forces the checkout to discard that (and other!?) local changes
* Quicker dev/test loop


leifmadsen (approved):

csibbitt:
test

PR #392, opened 2022-11-18
Expand support for OCP 4.11 (#391)

leifmadsen:
* Expand support for OCP 4.11

Allow installation to be done on OCP 4.11 while updating the smoketest
jobs to support later versions of the client. Also migrate to using
community-operators CatalogSource instead of OperatorHub.io. Only enable
community-operators when the use_community strategy is enabled.

Update the token request syntax when requesting a service account token.
Add checks to look for oc client version and fail if we're using a
version that's too old.

* Make passwords safer in smoketest job template

Encapsulate the password values with double quotes to help make them
safer for consumption in the template. I had an odd situation where the
password contained a bunch of extended characters and caused the
smoketest to report an error on the template having an issue with yaml
to json.

The password contained several characters such as . and : which confused
the template. Wrapping the contents in the double quotes allowed the
smoketest to apply the job.batch template and result in a working
smoketest run.

(cherry picked from commit 729f8411663d840d21a697bb965dcaf9eddb7c18)

### DIFF
--- a/build/stf-run-ci/tasks/setup_base.yml
+++ b/build/stf-run-ci/tasks/setup_base.yml
@@ -13,6 +13,8 @@
           name: certified-operators
         - disabled: false
           name: redhat-operators
+        - disabled: "{{ false if __service_telemetry_observability_strategy == 'use_community' else true }}"
+          name: community-operators
 
 - name: Create OperatorGroup
   k8s:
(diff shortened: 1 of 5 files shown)


csibbitt (approved):

leifmadsen:
Per discussion, do not expand anything at this time. Changes will remain on `master` only for now. Closing.

PR #395, opened 2022-11-29
Migrate to dependencies.yaml

leifmadsen:
Migrate the requires list out of the CSV and into
metadata/dependencies.yaml. Primarily this is for product release
reasons and dependency resolution with OSBS when performing builds for
product release. These changes are the result of following through the
documentation written at
https://olm.operatorframework.io/docs/concepts/olm-architecture/dependency-resolution/#declaring-dependencies
and
https://github.com/operator-framework/operator-registry/blob/master/docs/design/operator-bundle.md

### DIFF
--- a/deploy/olm-catalog/service-telemetry-operator/manifests/service-telemetry-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/service-telemetry-operator/manifests/service-telemetry-operator.clusterserviceversion.yaml
@@ -203,12 +203,6 @@ spec:
         name: servicemonitors.monitoring.coreos.com
         version: v1
       version: v1beta1
-    required:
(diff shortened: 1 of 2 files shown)


leifmadsen:
Do not merge this yet. I need to create an index image with the bundles and determine if this will work as expected and not cause any regressions. I just wanted to get something stage while I was investigating this during a CVP job failure that is non-gating (does not block releases).

leifmadsen:
Per a conversation we had today, there might be some further dependency changes that could live in here other than just the reference to the CRD. Will need to investigate further and make sure we're landing the expected changes that allows the release pipeline to work more efficiently.

leifmadsen:
I tested this previously and I don't think this works with operator-sdk 0.19.4. Abandoning.

PR #397, opened 2022-11-29
Update metadata to reference stable-1.5

leifmadsen:
Update the metadata to reference stable-1.5 so that we can perform
nightly builds against upstream stable-1.5 repository and result in an
index image that can be referenced by a CatalogSource manifest to allow
installation of STF 1.5 content.

Depends-On: https://github.com/infrawatch/smart-gateway-operator/pull/132


leifmadsen:
test

leifmadsen:
These changes are primarily to allow us the ability to ship nightly stable-1.5 builds upstream using the release-automation tooling. I have tested these changes via https://github.com/infrawatch/release-automation/pull/6 and performed a mild regression test by doing a manual installation following the current upstream documentation, but replacing the `nightly` tags with `nightly-1.5` tags on the index image for the CatalogSource.

I then did an installation following documentation to result in a working STF cluster which is currently receiving data from a standalone Zed installation.

I am now working through the process of verifying that the builds are all proper. Initial investigation is failing that but those are changes that I need to review in the release-automation set up.

The parent changes that will test and verify this functionality is https://github.com/infrawatch/release-automation/pull/6

The bundles appear to be correct because I'm getting separate tags in the bundle images and I'm seeing the appropriate channel names showing up telling me these metadata changes are correct. Once I confirm that it's an issue in the release-automation scripts or GitHub Actions job, then I will ask for reviewers. I trust this is safe and ready to merge down though, but want to verify.

leifmadsen:
So I was thinking there was a problem with the sha256 because they were the same between master and stable-1.5, but I think the code being pulled into the base image (which is the same release version on 4.10) is just resulting in the same contents and thus the same layered image stack. I ran separate builds triggering against different branches in quay.io, and the result is the same, so this isn't coming from release-automation pushing images from the same build.

Also walking through the releaser.sh script there, the only `docker push` is around the built bundles and index image after importing bundles via `opm`, so it doesn't seem there is any problem / bug there.

I went through all the hashes showing in the CSVs installed from the bundle, and everything aligns properly, so it looks like everything is good to go here, no issues!

csibbitt (approved):

PR #399, opened 2022-12-06
Enable Prometheus RemoteWriteReceiver

csibbitt:


leifmadsen (approved):
Works for me. Hopefully at some point we can just add the ObO installation from the upstream index image or something and leverage that going forward?

I do think we should make this configurable though at some point, perhaps when we add the ObO support which would provide a set of CRD changes under the `backends.metrics` configuration and/or `observabilityStrategy` setting.

For testing and development purposes though, I don't think any of that needs to happen before landing this.

leifmadsen:
I don't think this will auto-merge due to lack of CI trigger. Perhaps we should just disable that in the `master` repo for now? If I have time Friday maybe I'll try creating another cluster and trigger?

csibbitt:
Closed because we are not planning to use remote-write functionality anymore

PR #401, opened 2023-01-24
Changes for 4.12

csibbitt:
* Catalog changes
* CI change to pre-clean cert-manager-operator
  * not 100% sure this is 4.12 related, but it's new and first seen during testing 4.12

### DIFF
--- a/build/stf-run-ci/tasks/pre-clean.yml
+++ b/build/stf-run-ci/tasks/pre-clean.yml
@@ -48,3 +48,14 @@
         image: quay.io/operatorhubio/catalog:latest
         displayName: OperatorHub.io Operators
         publisher: OperatorHub.io
+
+# Remove the cert manager since we install it as part of the CI/documented pre-install process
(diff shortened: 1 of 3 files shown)


elfiesmelfie:
I've manually triggered a build of this PR on OCP 4.12.

csibbitt:
Thanks, @elfiesmelfie  - looks like it passed, too! :tada:

leifmadsen (approved):

elfiesmelfie:
> Thanks, @elfiesmelfie - looks like it passed, too! tada

I think we're still missing something to have both jenkinses reporting back. I think they're both using the same channel to report back i.e. continuous-integration/jenkins/pr-merge. 
I just triggered a build on OCP 4.10 and the check reverted to pending

elfiesmelfie:
I made some changes to Jenkins, and have a custom status check label.

I need to figure out how to add this to the casc config for future deployments

PR #405, opened 2023-02-16
Remove Loki from stf-run-ci

vyzigold:
This PR removes everything Loki related from stf-run-ci and hardcodes the logs backend as disabled when deploying with stf-run-ci.


build/stf-run-ci/tasks/main.yml
```diff
@@ -42,72 +40,13 @@
     base_dir: "{{ playbook_dir }}"
   when: base_dir | length == 0
 
-- name: Get new operator sdk
-  when: __local_build_enabled | bool or __deploy_from_bundles_enabled | bool or __deploy_loki_enabled | bool
```
leifmadsen, line 46: Seems like this might have actually been a bit more open than it should have. I don't think we wanted to change the golang version just because we enabled local builds or bundle deployment? Either way, I guess it's going away now :)

leifmadsen (approved):
Looks good to me. Thanks!

csibbitt (changes requested):
pre-emptive block while I write comments

csibbitt (changes requested):
Please just remove the loki condition on the "Get new operator sdk" task, but leave the task in place for other uses. The rest LGTM.

build/stf-run-ci/tasks/main.yml
leifmadsen: Ah ok, that makes sense and might be why some of the `when` stuff got overloaded here. It might sound like we have to only soft remove some of this and leave the ability for the `__deploy_from_bundles_enabled` to get the newer operator-sdk along with the script that pulls that newer version down.

vyzigold: I thought, that the loki-operator was the only thing, that uses the newer operator-sdk. Thanks for finding this.

csibbitt: Once you added it, I found additional uses :)

elfiesmelfie:
I'll run the tests manually

PR #406, opened 2023-02-27
Expose ability to set certificate renewal target times

vkmc:
Expose the duration value in STO to allow better control of the default certificate renewal times.


deploy/crds/infra.watch_v1beta1_servicetelemetry_cr.yaml
```diff
@@ -33,6 +33,8 @@ spec:
           strategy: persistent
           persistent:
             pvcStorageRequest: 20Gi
+      endpoint_cert_duration: 70080h
```
leifmadsen, line 36: The use of the snake_case is part of the conversion from camelCase to the Ansible workflow. The CRD should be in camelCase, not snake_case.

deploy/crds/infra.watch_servicetelemetrys_crd.yaml
```diff
@@ -175,6 +175,12 @@ spec:
                                     type: string
                                 type: object
                             type: object
+                          endpoint_cert_duration:
```
leifmadsen, line 178: We might want to add a pattern expression here as well like they do for Prometheus CRD for the retention value:

https://github.com/prometheus-operator/prometheus-operator/blob/main/example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml#L5538-L5544

roles/servicetelemetry/tasks/_local_signing_authority.yml
```diff
@@ -8,7 +8,6 @@
         name: '{{ ansible_operator_meta.namespace }}-selfsigned'
         namespace: '{{ ansible_operator_meta.namespace }}'
       spec:
-        duration: '{{ certificate_duration }}'
```
leifmadsen, line 11: With this removal does this mean that the self-signed issuer will need to regenerate every 90(?) days? The certificate_duration was recently added I believe, so I'm not sure the implications of removing it here on the overall schema.

leifmadsen (changes requested):
Couple of requested changes to the casing scheme in the files. Also there should be changes to the stf-run-ci interface to allow passing in values for testing purposes.

roles/servicetelemetry/tasks/_local_signing_authority.yml
csibbitt: Victoria asked me about this out-of-band and we agree that my addition of the duration property to the Issuer object was a mistake. That property does not appear to exist[1], and we expect this change to have no effect.

That said, since we have no unit-test coverage, I would appreciate seeing the certificates that are generated by an altered duration in order to prove that this change (and really the whole PR) is behaving as expected.

[1] https://cert-manager.io/docs/reference/api-docs/#cert-manager.io/v1.IssuerConfig

csibbitt: This idea feels like it might be better suited to https://github.com/infrawatch/functional-tests/ where we do a series of `oc patch` commands to exercise all the documented features, rather than the smoketest which is more focused on ensuring the default config stands up and appears to basically operate end to end.

vkmc: +1 I was performing manual testing and also looking to see if there is a good suite I can add testing for this. I'll try adding some coverage in the functional tests suite in my follow up patch.

deploy/crds/infra.watch_servicetelemetrys_crd.yaml
vkmc: For what I understand, the duration can be only expressed in hours (based on the code here https://github.com/cert-manager/cert-manager/blob/v1.2.0/pkg/apis/certmanager/v1/const.go#L26). So I'll use a regex only for that.

csibbitt: I got an email referencing a comment here about "can be only expressed in hours", but maybe it was deleted? In case there is doubt about the type, it is here: https://github.com/cert-manager/cert-manager/blob/master/deploy/crds/crd-certificates.yaml#L91 

"Value must be in units accepted by Go time.ParseDuration https://golang.org/pkg/time/#ParseDuration"

deploy/crds/infra.watch_servicetelemetrys_crd.yaml
```diff
@@ -175,11 +175,15 @@ spec:
...
+                          endpointCertDuration:
                             description: The requested 'duration' (i.e. lifetime) of the ElasticSearch endpoint Certificate.
+                              Minimum accepted duration is 1 hour. Value must be in units accepted by Go time.ParseDuration https://golang.org/pkg/time/#ParseDuration
+                            pattern: ^((([0-9]+)h)?(([0-9]+)m)?(([0-9]+)s)?(([0-9]+)ms)?)$
                             type: string
-                          ca_cert_duration:
+                          caCertDuration:
                             description: The requested 'duration' (i.e. lifetime) of the ElasticSearch CA Certificate.
+                              Minimum accepted duration is 1 hour. Value must be in units accepted by Go time.ParseDuration https://golang.org/pkg/time/#ParseDuration
+                            pattern: ^((([0-9]+)h)?(([0-9]+)m)?(([0-9]+)s)?(([0-9]+)ms)?)$
```
leifmadsen, line 186: Looking at this some more, I feel like because we're adding 2 parameters here, that we should actually nest them like we do for the `storage` parameters. Perhaps add these under a top level `security` object. This would help in the future if we need to add more items (we probably won't of course), and gives a more natural setup of these parameters IMO.

leifmadsen: `security` or `certificates` ? Might be some other items that make more sense.

deploy/crds/infra.watch_servicetelemetrys_crd.yaml
vkmc: Yes Chris, thanks for checking... I added a pattern to match that type.

deploy/crds/infra.watch_servicetelemetrys_crd.yaml
vkmc: +1 fixed that in the latest PR

build/stf-run-ci/README.md
```diff
@@ -35,6 +35,8 @@ choose to override:
...
 | `prometheus_webhook_snmp_repository`                   | <git_repository>         | https://github.com/infrawatch/prometheus-webhook-snmp | Which Prometheus webhook snmp git repository to clone                                                               |
 | `loki_operator_repository`                             | <git_repository>         | https://github.com/viaq/loki-operator                 | Which Loki-operator git repository to clone                                                                         |
+| `__service_telemetry_events_endpoint_cert_duration`    | [ParseDuration](https://golang.org/pkg/time/#ParseDuration) | 2160h              | Lifetime of the ElasticSearch endpoint certificate (minimum duration is 1h)   |
+| `__service_telemetry_events_ca_cert_duration`          | [ParseDuration](https://golang.org/pkg/time/#ParseDuration) | 70080h             | Lifetime of the ElasticSearch CA certificate (minimum duration is 1h)         |
```
leifmadsen, line 39: These generally follow the naming pattern in the CRD, so you'll want to update these to `__service_telemetry_events_certificates_endpoint_cert_duration` etc.

leifmadsen (commented):
This is looking really close now. I think if you just update those stf-run-ci variable names to include `certificates` (following the existing patterns) then this looks pretty much ready to go.

leifmadsen:
Looking good! Will do a final pass through after lunch but I think you've hit everything here.

build/stf-run-ci/defaults/main.yml
```diff
@@ -9,6 +9,8 @@ __local_build_enabled: true
 __deploy_from_bundles_enabled: false
 __deploy_stf: true
 
+__service_telemetry_events_certificates_endpoint_cert_duration: 70080h
+__service_telemetry_events_certificates_ca_cert_duration: 2160h
```
vkmc, line 13: I'd expect this value to be selected when running ansible-playbook --extra-vars working_branch="vkmc-add-cert-duration-param" build/run-ci.yaml with no params, but it's defaulting to 70080h. I need to investigate why is that.

build/stf-run-ci/tasks/deploy_stf.yml
```diff
@@ -40,6 +40,9 @@
                 persistent:
                   storageClass: {{ __service_telemetry_storage_persistent_storage_class }}
       {% endif %}
+              certificates:
+                endpoint_cert_duration: {{ __service_telemetry_events_certificates_endpoint_cert_duration }}
+                ca_cert_duration: {{ __service_telemetry_events_certificates_ca_cert_duration }}
```
leifmadsen, line 45: Ah I think I figured out the problem you mentioned earlier.

snake_case vs camelCase here (should use camelCase for the ServiceTelemetry manifest)

leifmadsen, line 45:
```suggestion
                endpointCertDuration: {{ __service_telemetry_events_certificates_endpoint_cert_duration }}
                caCertDuration: {{ __service_telemetry_events_certificates_ca_cert_duration }}
```

build/stf-run-ci/tasks/deploy_stf.yml
```diff
@@ -59,6 +62,12 @@
       {% if __service_telemetry_storage_persistent_storage_class is defined %}
                 storageClass: {{ __service_telemetry_storage_persistent_storage_class }}
       {% endif %}
+          transports:
+            qdr:
+              enabled: true
+              certificates:
+                endpoint_cert_duration: {{ __service_telemetry_transports_certificates_endpoint_cert_duration }}
+                ca_cert_duration: {{ __service_telemetry_transports_certificates_ca_cert_duration }}
```
leifmadsen, line 70:
```suggestion
                endpointCertDuration: {{ __service_telemetry_transports_certificates_endpoint_cert_duration }}
                caCertDuration: {{ __service_telemetry_transports_certificates_ca_cert_duration }}
```

roles/servicetelemetry/defaults/main.yml
```diff
@@ -56,6 +54,9 @@ servicetelemetry_defaults:
           persistent:
             storage_class: ""
             pvc_storage_request: 20Gi
+        certificates:
+          endpoint_cert_duration: 70080h
+          ca_cert_duration: 70080h
```
leifmadsen, line 59: You were using the defaults here instead of overriding them because of the casing issue earlier (untested).

vkmc: Gotcha, fixing it

leifmadsen (approved):
Looks good to me now. Thanks!

vkmc:
Performing some more manual testing since I noticed an odd behavior. Please do not merge this yet.

vkmc:
Find the issue (thanks Leif!) and performed manual testing again. It works as expected and it's ready to be merged.

csibbitt:
test

csibbitt:
Well the 4.10 test passed, but the 4.12 CI server died sometime mid-way through the job and is giving the "Application is not available" page.

I see that the commit before the indent change passed all tests, and the indent change looks good, so I'm going to trust the reports that this passes manual testing.

PR #409, opened 2023-03-01
CI change to pre-clean cert-manager-operator

leifmadsen:
(cherry picked from commit 37dceed7e55856820c10fb812da0ed9cd6551a3b with
modifications)

### DIFF
--- a/build/stf-run-ci/tasks/pre-clean.yml
+++ b/build/stf-run-ci/tasks/pre-clean.yml
@@ -29,3 +29,33 @@
     kind: clusterroles
     label_selectors:
       - "olm.owner.namespace = {{ namespace }}"
+
+# Clean the environment if it has OperatorHub.io CatalogSource still enabled as
+# environment is using community-operators CatalogSource when use_community has
+# been enabled. This avoids installing an additional CatalogSource which is no
+# longer required.
+- name: Remove OperatorHub.io CatalogSource if it installed
+  k8s:
+    state: absent
+    definition:
+      apiVersion: operators.coreos.com/v1alpha1
+      kind: CatalogSource
+      metadata:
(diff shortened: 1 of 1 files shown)


csibbitt (approved):

leifmadsen:
I can't seem to get the 4.12 CI system to work, so if this passes OCP 4.10 testing I'm going to merge it down instead of getting blocked by 4.12 CI.

leifmadsen:
Oh duh.... this is stable-1.5 branch which doesn't have the OCP 4.12 support in the CSV, so the 4.12 CI system will never pass.

I think there are other issues with launching the test workload, but I also just realized only OCP 4.10 CI passing is required to have landed this anyways.

elfiesmelfie:
> 
> I think there are other issues with launching the test workload, but I also just realized only OCP 4.10 CI passing is required to have landed this anyways.

I think we can edit that on a per-branch basis in the repo settings

leifmadsen:
> > I think there are other issues with launching the test workload, but I also just realized only OCP 4.10 CI passing is required to have landed this anyways.
> 
> I think we can edit that on a per-branch basis in the repo settings/branches and you'll need to edit branch protection rules to require a certain status before merging. It'll autocomplete and you can search using "jenkins"

Sounds good. I would just leave this for now because next week we'll be merging down changes from master, at which point we _do_ want to test against 4.12 on this branch. At some point in the next few weeks we _might_ want to add OCP 4.13 support to master (but not stable-1.5) until OCP 4.14 is available (at which point we'd drop the OCP 4.13 testing).

It might be important to have OCP 4.13 once it comes out so that we can develop our roadmap items and integrate some testing/validation for them.

PR #410, opened 2023-03-01
Add olm.substitutesFor annotation

leifmadsen:
When building the bundle, add the olm.substitutesFor annotation so that
rebuilds (via Freshmaker or otherwise) that have different build
versions appended to the end of the main version (e.g.
service-telemetry-operator-v1.5.15000000001-0.1234.p) can allow the
bundle to be upgraded without changing the skipRange, allowing for
rebuilds due to changes in the resulting base image, but no changes in
the actual operator itself can result in the latest bundle version being
deployed.

Related: RHOSOR-576

### DIFF
--- a/deploy/olm-catalog/service-telemetry-operator/manifests/service-telemetry-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/service-telemetry-operator/manifests/service-telemetry-operator.clusterserviceversion.yaml
@@ -171,7 +171,7 @@ metadata:
       Service Telemetry platform for telco grade monitoring.
     olm.properties: '[{"type": "olm.maxOpenShiftVersion", "value": "4.12"}]'
     olm.skipRange: '>=<<BUNDLE_OLM_SKIP_RANGE_LOWER_BOUND>> <<<OPERATOR_BUNDLE_VERSION>>'
-    olm.substitutesFor: '<<OPERATOR_BUNDLE_VERSION>>'
+    olm.substitutesFor: 'service-telemetry-operator.v<<OPERATOR_BUNDLE_VERSION>>'
     operatorframework.io/suggested-namespace: service-telemetry
     operators.openshift.io/valid-subscription: '["OpenStack Platform", "Cloud Infrastructure",
       "Cloud Suite"]'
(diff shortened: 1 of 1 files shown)


leifmadsen:
Following an example via https://github.com/operator-framework/operator-registry/pull/612/files#diff-bafb1f09f1d5bf3045200f3c2dc3e41f111d2682971d45a64104d5d5d7821f05R8

leifmadsen:
I think the easiest thing here is going to be to generate bundle and upload it to an index image, create a custom CatalogSource with that index image, then modify the CSV to append some extra string to the end of the version, generate a new bundle, update the index image with the new bundle version, and then push that index image to the testing location. Force update the index image for the CatalogSource on the test bench, and then see if an upgrade from `1.5.<unixdate>` to `1.5.<unixdate>-0.1234.rebuild` works as intended. The intention is that the bundle is still installable, and that the bundle is updated when the index image is refreshed (always running latest version).

csibbitt (approved):

compi-migui:
This is just for testing how a build with `olm.substitutesFor` would behave, right? We're not actually trying to merge this?

That was my impression but now I see Chris' approval and I'm not sure anymore :)

leifmadsen:
> This is just for testing how a build with `olm.substitutesFor` would behave, right? We're not actually trying to merge this?
> 
> That was my impression but now I see Chris' approval and I'm not sure anymore :)

It's approved, but until I finish testing per my comment, I won't merge it (hence the do-not-merge label :)). I do plan to merge this at some point though.

compi-migui:
Oh then I think we don't need this at all. Freshmaker already adds olm.substitutesFor when doing rebuilds:
https://github.com/redhat-exd-rebuilds/freshmaker/blob/f111fa37c7ee4d5dc6b2face5e46c88ffd3e4065/freshmaker/handlers/botas/botas_shipped_advisory.py#L300-L312
So that the rebuilt image can substitute for the original image we build through usual channels.

Our original builds don't need to substitute for previous ones at all, skipRange takes care of that.

It would make sense to do a test build, to confirm that skipRange and substitutesFor play nicely with each other and Freshmaker's builds don't break our upgrade path.

leifmadsen:
> Oh then I think we don't need this at all. Freshmaker already adds olm.substitutesFor when doing rebuilds: https://github.com/redhat-exd-rebuilds/freshmaker/blob/f111fa37c7ee4d5dc6b2face5e46c88ffd3e4065/freshmaker/handlers/botas/botas_shipped_advisory.py#L300-L312 So that the rebuilt image can substitute for the original image we build through usual channels.
> 
> Our original builds don't need to substitute for previous ones at all, skipRange takes care of that.
> 
> It would make sense to do a test build, to confirm that skipRange and substitutesFor play nicely with each other and Freshmaker's builds don't break our upgrade path.

Then I'm confused what RHOSOR-576 is asking for then? If there are no changes required, is the only reason we're not shipping freshmaker builds because we've never tested them for OLM update workflow?

leifmadsen:
Closing this PR as it seems this isn't necessary.

compi-migui:
> Then I'm confused what RHOSOR-576 is asking for then? If there are no changes required, is the only reason we're not shipping freshmaker builds because we've never tested them for OLM update workflow?

That is exactly right.

PR #411, opened 2023-03-02
Implement changes for operator-sdk-1.26.0 testing

leifmadsen:
Implement changes that allow testing validation via operator-sdk-1.26.0
without bumping the entire bundle generation process from
operator-sdk-0.19.4 to post-operator-sdk-1.x.

These are the same tests run for validation during product pipeline
verification.

* Adds test to verify building of the bundle image works.
* Adds KinD deployment to allow executing scorecard checks.

Related: STF-1252

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,15 +20,9 @@ jobs:
 
       - name: Lint Ansible roles/servicetelemetry/ directory
         run: ${HOME}/.local/bin/ansible-lint roles/servicetelemetry
-# TODO: requires a bunch of work on our bash scripts, or finesse
-#      - name: Run Super-Linter
-#        uses: github/super-linter@v3
-#        env:
-#          DEFAULT_BRANCH: master
-#          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-  build-check:
-    name: Build check
+  build-operator-check:
+    name: Build Operator check
     runs-on: ubuntu-20.04
 
     steps:
@@ -38,15 +32,15 @@ jobs:
       - name: Verify image builds
         run: docker build --tag infrawatch/service-telemetry-operator:latest --file build/Dockerfile .
 
-  bundle-check:
-    name: Bundle check
+  build-bundle-check:
+    name: Build bundle check
     runs-on: ubuntu-20.04
 
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
 
-      - name: Get operator-sdk image
+      - name: Get operator-sdk image 0.19.4
         run: curl --output operator-sdk -JL https://github.com/operator-framework/operator-sdk/releases/download/$RELEASE_VERSION/operator-sdk-$RELEASE_VERSION-x86_64-linux-gnu
         env:
           RELEASE_VERSION: v0.19.4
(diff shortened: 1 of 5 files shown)


build/generate_bundle.sh
```diff
@@ -38,6 +38,13 @@ generate_bundle() {
     echo "---- Generated bundle complete at ${WORKING_DIR}/manifests/${OPERATOR_NAME}.clusterserviceversion.yaml"
 }
 
+copy_extra_metadata() {
+    echo "-- Copy extra metadata in"
+    pushd "${REL}/../"
+    cp -r ./deploy/olm-catalog/service-telemetry-operator/tests/ "${WORKING_DIR}"
+    cp ./deploy/olm-catalog/service-telemetry-operator/metadata/properties.yaml "${WORKING_DIR}/metadata/"
+}
+
```
leifmadsen, line 47: We add this because our version of operator-sdk for building doesn't understand these files, but newer versions of operator-sdk (for testing purposes) does, and newer versions of `opm` (as used in both downstream and upstream index image builds) also understands these files. Just copy them into the bundle directory during building.

deploy/olm-catalog/service-telemetry-operator/manifests/service-telemetry-operator.clusterserviceversion.yaml
```diff
@@ -169,7 +169,6 @@ metadata:
     description: Service Telemetry Framework. Umbrella Operator for instantiating
       the required dependencies and configuration of various components to build a
       Service Telemetry platform for telco grade monitoring.
-    olm.properties: '[{"type": "olm.maxOpenShiftVersion", "value": "4.12"}]'
```
leifmadsen, line 172: This gets moved out of the CSV directly now in favour of it being loaded from the `properties.yaml`. If you run the `operator-sdk-1.26.0 bundle validation` it will WARN you about having these contents inline within the CSV itself.

deploy/olm-catalog/service-telemetry-operator/metadata/properties.yaml
```diff
@@ -0,0 +1,4 @@
+    olm.properties: '[{"type": "olm.maxOpenShiftVersion", "value": "4.10"}]'
+properties:
+  - type: olm.maxOpenShiftVersion
+    value: "4.10"
```
leifmadsen, line 4: Looks like this was wrong for a bit. It should be 4.12 in master. Will make a suggested change and commit it.

deploy/olm-catalog/service-telemetry-operator/tests/scorecard/config.yaml
```diff
@@ -0,0 +1,21 @@
...
+    labels:
+      suite: basic
+      test: basic-check-spec-test
+  - image: quay.io/operator-framework/scorecard-test:latest
+    entrypoint:
+    - scorecard-test
+    - olm-bundle-validation
+    labels:
+      suite: olm
+      test: olm-bundle-validation-test
```
leifmadsen, line 21: The contents of this scorecard file came from https://sdk.operatorframework.io/docs/testing-operators/scorecard/#config-file and matches the 2 tests we were encouraged to pass in STF-1256.

deploy/olm-catalog/service-telemetry-operator/metadata/properties.yaml
leifmadsen: Ummmm weird... it looked like it was wrong wherever I was looking at this before, but that's not true. Fixing my typo. Also that `olm.properties` line was for reference when editing this file, and is invalid.

leifmadsen, line 4:
```suggestion
properties:
  - type: olm.maxOpenShiftVersion
    value: "4.12"
```

.github/workflows/main.yml
```diff
@@ -20,33 +20,56 @@ jobs:
...
     runs-on: ubuntu-20.04
 
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
 
+      - name: Get operator-sdk image 0.19.4
+        run: curl --output operator-sdk -JL https://github.com/operator-framework/operator-sdk/releases/download/$RELEASE_VERSION/operator-sdk-$RELEASE_VERSION-x86_64-linux-gnu
+        env:
+          RELEASE_VERSION: v0.19.4
```
paramite, line 46: Is there a reason to not have this env var noted in the workflow set instead? I'm asking because you are using the same env var with same value below.

.github/workflows/main.yml
```diff
@@ -60,8 +83,37 @@ jobs:
...
 
+      # generate the bundle using operator-sdk-0.19.4
       - name: Generate bundle
         run: WORKING_DIR=/tmp/bundle ./build/generate_bundle.sh
 
+      # prepare the environment to run bundle validation and bundle scorecard checks
+      - name: Get operator-sdk image 1.26.0
+        run: curl --output operator-sdk-$RELEASE_VERSION -JL https://github.com/operator-framework/operator-sdk/releases/download/$RELEASE_VERSION/operator-sdk_linux_amd64
+        env:
+          RELEASE_VERSION: v1.26.0
```
paramite, line 94: Same here. Why not to do two env vars instead and have it in global env for the workflow?

build/generate_bundle.sh
paramite: Thanks for explanation. Can we please put this comment in the code instead, so that when somebody else than you will need to read the code in the future, it won't end with scratching head and lots of WTH? :).

paramite (approved):
Seems fine to me. Just few nitpicks.

.github/workflows/main.yml
```diff
@@ -20,33 +20,56 @@ jobs:
...
+  build-operator-check:
+    name: Build Operator check
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Verify image builds
+        run: docker build --tag infrawatch/service-telemetry-operator:latest --file build/Dockerfile .
```
vkmc, line 33: NIT This is for my learning mostly... shouldn't we prioritize using podman instead of docker? Not sure if the ubuntu-20.04 image that is being used as base includes the required repos

build/generate_bundle.sh
vkmc: +1

.github/workflows/main.yml
leifmadsen: For the purposes of what we're doing here, I'm not sure there is a huge difference. We're simply checking that a build is possible. I agree we should use podman, but I'm not sure what version is on ubuntu-20.04, what repos are necessary, and I'd prefer not to debug problems with podman here. I'm going to defer on this one for now.

csibbitt (approved):
LGTM, my only question is whether there is a specific trigger we are waiting for to bring the build itself up to SDK >= 1.26+?

leifmadsen:
> LGTM, my only question is whether there is a specific trigger we are waiting for to bring the build itself up to SDK >= 1.26+?

I was talking about this yesterday, and until there is a reason to change to operator-sdk-1.x (which I think is going to be relatively invasive from a build/CI/workflow/documentation perspective) I'm kind of holding off. The newer version of operator-sdk requires a different build method via Makefiles and kustomize. Last time I looked at this I realize how many extra little systems we'd have to touch and kind of abandoned it since the risk/reward didn't seem worth it.

We could evaluate again though, because last time the big issue was around the CI system which is much more modular and existing than it was when I looked at an operator-sdk version bump last time.

Also since we don't really have a roadmap for versions of RHOSP beyond 17.1 and understanding what we're doing there Operator-wise, I wasn't sure if the work was worth it since we'll go basically into maintenance mode at the end of this year (pending changes to roadmap, approach etc).

leifmadsen:
Merging this as the latest changes were only for GHA and a comment in the generate_bundle.sh, so I'm not going to re-run the OCP test suite.

.github/workflows/main.yml
vkmc: +1, thanks

PR #414, opened 2023-03-08
Fix default CA cert lifetime values in stf-run-ci

csibbitt:


vkmc (changes requested):
Also here, 70080h

vkmc (approved):
Lgtm

PR #415, opened 2023-03-08
fix/client version mismatch

leifmadsen:
- Fix PROMETHEUS_K8S_TOKEN to account for oc version mismatch
- Fix to use correct variable name in test


csibbitt (approved):

leifmadsen:
Everything passed CI so I'm going to merge this down and cherry pick it back to stable-1.5. The missing tests are expected due to recent rename and will never trigger.

PR #402, opened 2023-01-25
[jenkins] Add custom context labels for github build status notification

elfiesmelfie:
Multiple jenkins deployment can now be run and report their build status separately instead of both reporting to the same ``continuous-integration/jenkins/pr-merge`` and overriding each other.

There is now ``continuous-integration/jenkins/ocp-<OCP_VERSION>/pr-merge`` NOTE: the OCP_VERSION is hardcoded at the moment

https://github.com/jenkinsci/github-scm-trait-notification-context-plugin

PR #421, opened 2023-04-25
Observability Operator for Prometheus & Alertmanager

csibbitt:
- Adds two options to observability_strategy:
  * use_redhat (OBO Only)
  * use_hybrid (OBO + grafana + elasticsearch + loki)
- Default to use_redhat for new installs
- Use upstream source for OBO in CI
- Added sensubility to deployment validation
- Adds  RBAC definitions required for OBO
  - These used to be handled by prometheus operator for us
  - If in the future we use the MonitoringStacks interface in OBO, it will manage them for us


csibbitt:
test

leifmadsen:
I don't think the comment triggered testing has been working for several weeks. There are open issues about migrating away from smee as a trigger, but so far I haven't seen any replacement. (I always test it anyways, but pretty sure the expectation is that everything is manually executed at this point.)

build/stf-run-ci/tasks/setup_base.yml
```diff
@@ -103,13 +103,55 @@
...
         installPlanApproval: Automatic
-        name: rhods-prometheus-operator
-        source: redhat-operators
+        name: prometheus
+        source: community-operators
         sourceNamespace: openshift-marketplace
   when:
     - __service_telemetry_observability_strategy == "use_community"
+
+- block:
```
leifmadsen, line 117: 😮

csibbitt:
> I don't think the comment triggered testing has been working for several weeks. There are open issues about migrating away from smee as a trigger, but so far I haven't seen any replacement. (I always test it anyways, but pretty sure the expectation is that everything is manually executed at this point.)

Except that these machines were just rebuilt and I don't know where to find the password, so I couldn't manually trigger it if I wanted too, so figured it was worth a shot.... and, hey it's working!

csibbitt:
Currently this does CI tests with "use_hybrid" instead of "use_redhat" because the tests depend on ElasticSearch for events testing. We have work scheduled to replace the built-in ES orcheastration with config that lets you point to an external one, and I figure that means moving the orcheastration from the operator role to the CI role so that we can still validate that functionality. As such, I didn't feel like it was a good idea to remove ES from the smoketests with this change, though I'm slightly uncomfortable that we're not testing with "use_redhat" (which will be the default for deployments). I think I will try adding some sort of ES bypass into the smoketest to allow us to test with "use_redhat" in the meantime.

I've also not tested moving between strategies yet. use_community <> use_redhat/hybrid seems tricky because you'll end up with two copies of things in under two namespdifferent APIs. I expect that use_redhat <> use_hybrid would be smooth.

roles/servicetelemetry/templates/manifest_alertmanager.j2
```diff
@@ -1,8 +1,9 @@
-apiVersion: monitoring.coreos.com/v1
+apiVersion: {{ prometheus_operator_api_string }}/v1
 kind: Alertmanager
 metadata:
   labels:
     alertmanager: '{{ ansible_operator_meta.name }}'
+    {{ prometheus_operator_label }}
```
leifmadsen, line 6: Does this leave an empty line (and does it even matter?) when the label is undefined in a use_community setup? I'm guessing you've run this locally and it didn't care though.

leifmadsen (commented):
Overall this looks really good. This is one of those PRs though where I wish we had an update/upgrade CI test that verified deployments across z-streams and y-streams.

A few questions and I think the only change I requested was around a couple of descriptions. This makes a lot of sense to me though and I approve of this direction.

leifmadsen:
> > I don't think the comment triggered testing has been working for several weeks. There are open issues about migrating away from smee as a trigger, but so far I haven't seen any replacement. (I always test it anyways, but pretty sure the expectation is that everything is manually executed at this point.)
> 
> Except that these machines were just rebuilt and I don't know where to find the password, so I couldn't manually trigger it if I wanted too, so figured it was worth a shot.... and, hey it's working!

Oh right! Well that's a nice find.

If you login to the `openshift-install` VM (which I think is where these machines are triggered from), the MOTD has a command/path for dumping the login password to the machines from the setup directory.

leifmadsen:
> Currently this does CI tests with "use_hybrid" instead of "use_redhat" because the tests depend on ElasticSearch for events testing. We have work scheduled to replace the built-in ES orcheastration with config that lets you point to an external one, and I figure that means moving the orcheastration from the operator role to the CI role so that we can still validate that functionality. As such, I didn't feel like it was a good idea to remove ES from the smoketests with this change, though I'm slightly uncomfortable that we're not testing with "use_redhat" (which will be the default for deployments). I think I will try adding some sort of ES bypass into the smoketest to allow us to test with "use_redhat" in the meantime.
> 
> I've also not tested moving between strategies yet. use_community <> use_redhat/hybrid seems tricky because you'll end up with two copies of things in under two namespdifferent APIs. I expect that use_redhat <> use_hybrid would be smooth.

Thanks for this. You were probably typing this as I was going through the code, so I didn't see it yet. It addresses a few of my comments.

roles/servicetelemetry/defaults/main.yml
```diff
@@ -3,8 +3,7 @@
 #
 # remove SmartGateway object when cloud no longer in current `clouds` object list.
 clouds_remove_on_missing: false
-# default observability strategy (compatible with STF 1.3)
-observability_strategy: use_community
+observability_strategy: use_redhat
```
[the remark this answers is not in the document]
csibbitt: Oh that's a good point - I'd assumed all deployed instances would have `observability_strategy: use_community` since that's the previous default, and that they wouldn't change when you upgrade. If this will change an existing STF deployment from use_community to use_redhat, that's bad news and was not the intention. I will test this, thanks.

roles/servicetelemetry/tasks/component_servicemonitor.yml
```diff
@@ -1,7 +1,7 @@
 - name: Create SG-specific Service Monitor manifest
   set_fact:
     sg_specific_servicemonitor_manifest: |
-      apiVersion: monitoring.coreos.com/v1
+      apiVersion: {{ prometheus_operator_api_string }}/v1
```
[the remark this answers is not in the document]
csibbitt: Yeah okay, that's a good point. Will do.

roles/servicetelemetry/templates/manifest_alertmanager.j2
csibbitt: I think the answer is that yes it does leave a line, but no it doesn't matter. I'm working on testing all three modes now.

build/stf-run-ci/tasks/setup_base.yml
```diff
@@ -103,13 +103,55 @@
...
+          displayName: Observability Operator - Test
+          icon:
+            base64data: ""
+            mediatype: ""
+          image: quay.io/rhobs/observability-operator-catalog:latest
+          publisher: Sunil Thaha
+          sourceType: grpc
+          updateStrategy:
+            registryPoll:
+              interval: 1m0s
```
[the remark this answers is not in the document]
leifmadsen: Yea that's fine with me. I also misread which file this was going into, and thought you were adding CatalogSource and Subscription management via STO :)

build/stf-run-ci/tasks/setup_base.yml
```diff
@@ -103,13 +103,55 @@
...
+        metadata:
+          labels:
+            operators.coreos.com/observability-operator.openshift-operators: ""
+          name: observability-operator
+          namespace: openshift-operators
+        spec:
+          channel: development
+          installPlanApproval: Automatic
+          name: observability-operator
+          source: observability-operator
```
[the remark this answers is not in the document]
leifmadsen: See comment above where I thought this was landing in STO roles no CI roles :)

deploy/alerts/alerts.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: monitoring.coreos.com/v1
+apiVersion: monitoring.rhobs/v1
```
leifmadsen, line 1: This is probably going to require documentation to address migration to another Operator in the future?

csibbitt: Ultimately, now might be a good time to start managing these in the operator instead of a documented `oc apply -f`

roles/servicetelemetry/tasks/component_prometheus.yml
```diff
@@ -18,11 +18,69 @@
...
+          namespace: '{{ ansible_operator_meta.namespace }}'
+  when:
+    - observability_strategy in ['use_redhat', 'use_hybrid']
+
+- name: Bind the local prometheus SA to prometheus cluster role (for oauth perms)
   k8s:
     definition:
       apiVersion: rbac.authorization.k8s.io/v1
-      kind: ClusterRoleBinding
+      kind: RoleBinding
```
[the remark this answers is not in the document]
csibbitt: In order to remove it, I'd need to grant the operator additional permissions, which is itself more legacy that's even harder to clean up. This will best be addressed with a release note.

roles/servicetelemetry/defaults/main.yml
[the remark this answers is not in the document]
csibbitt: I tested this and it's true that it will attempt to convert existing installs. We either need to make that a smooth experience, or figure out another strategy for the migration. For the time being I'm going to revert the default while we think this through.

csibbitt:
I'm inviting more reviewers to increase overall exposure and maybe get some ideas about how we could provide a smooth upgrade experience. The basic problem is that existing deployments are all _implicitly_ "use_community" because they don't specify an observability_strategy. If we change the default to be "use_redhat" in future versions, it will cause unexpected changes (trying to use OBO suddently) that may require (as-yet-unspecified) manual interactions to bring things back to a settled state. 

One idea I have in mind is to detect existing "use_community" installs that don't explicitly set observability_strategy and set it for them so that they will remain untouched until deliberately edited.

deploy/crds/infra.watch_v1beta1_servicetelemetry_cr.yaml
```diff
@@ -3,7 +3,7 @@ kind: ServiceTelemetry
 metadata:
   name: default
 spec:
-  observabilityStrategy: use_community
+  observabilityStrategy: use_redhat
```
csibbitt, line 6: I believe this is still safe to use as a default here because it's the CR used if you're deploying from the UI, so it's only used if it's a brand new install.

roles/servicetelemetry/tasks/component_prometheus.yml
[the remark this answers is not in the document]
csibbitt: I've found another instance of this. I thin I'm going to revert it and handle this in a separate PR

csibbitt:
test

csibbitt:
## TESTING - Changing from observability_strategy: use_community to use_redhat (and back)

The latest commit adds the ability to smoothly transition from use_community to use_redhat and back, assuming both the community prometheus operator and OBO are installed.

```
$ oc get stf -o yaml | grep strategy
    observability_strategy: use_community

$ oc get po | grep default-0
alertmanager-default-0         3/3     Running     0               100m
prometheus-default-0           3/3     Running     1 (101m ago)    101m

$ oc get prometheus.monitoring.coreos.com
default             1                                          102m

No resources found in service-telemetry namespace.
```

```
$ oc edit stf
[...]
observability_strategy: use_redhat

$ oc get po | grep default-0
alertmanager-default-0                                    3/3     Running     0               46s
prometheus-default-0                                      3/3     Running     0               51s

$ oc get prometheus.monitoring.coreos.com
No resources found in service-telemetry namespace.

default             1         1       True         True        75s
```

I confirmed that the scrape targets are UP and everything looks to be fine.

I can even change it back the other way.

```
$ oc edit stf
[...]
observability_strategy: use_community

$ oc get po | grep default-0
alertmanager-default-0                                    0/3     Terminating   0               40s
prometheus-default-0                                      3/3     Running       1 (30s ago)     38s

$ oc get prometheus.monitoring.coreos.com
default             1                                          2m9s

No resources found in service-telemetry namespace.
```

csibbitt:
I believe the latest patch solves the upgrade problem. I've set the default to use_redhat, and added code for the following:

* Existing STF objects get an explicit "use_community" added
* New STF objects get the default ("use_redhat") explicitly added

## TESTING - Upgrades (kinda)

I'm starting with an existing empty (all defaults) STF object in a system where use_community is the default. All containers up and healthy.

```
$ oc get stf -o yaml | grep -A 1 spec:
  spec: {}
  status:

$ oc get prometheus.monitoring.coreos.com
default             1                                          61m
```

Now I upgrade to the new code where use_redhat is the default, but it honours existing community deploymemnts.

```
$ git reflog | head -1
0838da7 HEAD@{0}: commit (amend): Enforce explicit observability_strategy

$ ./build.sh 
[...]
Successfully pushed image-registry.openshift-image-registry.svc:5000/service-telemetry/service-telemetry-operator@sha256:7aea4ac3942403c4494158d3e9564fefb5e414db15bda604ae028a3231972564
Push successful

$ oc delete po service-telemetry-operator-6dc87b7bbd-92vd4
pod "service-telemetry-operator-6dc87b7bbd-92vd4" deleted

$ oc get stf -o yaml | grep -A 1 spec:
  spec:
    observability_strategy: use_community
```

I create a second STF, also empty (defaults), which should default to redhat components.

```
$ oc apply -f - <<EOF
apiVersion: infra.watch/v1beta1
kind: ServiceTelemetry
metadata:
  name: default2redhat
spec: {}
EOF

$ oc get stf default2redhat -o yaml | grep -A 1 spec:
spec:
  observability_strategy: use_redhat
```

They even live happily side by side

```
$ oc get prometheus.monitoring.coreos.com
default             1                                          34s

NAME             VERSION   DESIRED   READY   RECONCILED   AVAILABLE   AGE
default2redhat             1         1       True         True        39s

$ oc get po | egrep default.\*-0
alertmanager-default-0                                           3/3     Running     0             92s
alertmanager-default2redhat-0                                    3/3     Running     0             3m4s
prometheus-default-0                                             3/3     Running     0             97s
prometheus-default2redhat-0                                      3/3     Running     0             2m23s
```

csibbitt:
test

roles/servicetelemetry/defaults/main.yml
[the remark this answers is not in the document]
csibbitt: Update on this thread is that I took care of these issues with [0838da7](https://github.com/infrawatch/service-telemetry-operator/pull/421/commits/0838da71b79afd51b7bd245397abb34e3621ba85) and show some testing below https://github.com/infrawatch/service-telemetry-operator/pull/421#issuecomment-1526167871

csibbitt:
I'm going to keep working on an ES bypass for the smoketests so we can test with use_redhat instead of use_hybrid, but I'll post that to another PR.

I consider this ready to merge once it passes CI but will keep it open for additional feedback.

leifmadsen:
> I'm going to keep working on an ES bypass for the smoketests so we can test with use_redhat instead of use_hybrid, but I'll post that to another PR. (Edit: #422)
> 
> I consider this ready to merge once it passes CI but will keep it open for additional feedback.

Going to do another code walk through today and will provide feedback. Once this is all complete I'd like to talk next week about how we approach the Elasticsearch stuff that will follow these changes and build upon it.

build/stf-run-ci/README.md
```diff
@@ -53,7 +53,7 @@ choose to override:
...
 | `__service_telemetry_trap_default_severity`            | <default_severity>       | <undefined>                                           | The trap severity if none is found in the Prometheus alert labels.                                                  |
 | `__service_telemetry_logs_enabled`                     | {true,false}             | false                                                 | Whether to enable logs support in ServiceTelemetry                                                                  |
-| `__service_telemetry_observability_strategy`           | <observability_strategy> | `use_community`                                       | Which observability strategy to use for deployment. Default deployment is 'use_community'. Also supported is 'none' |
+| `__service_telemetry_observability_strategy`           | <observability_strategy> | `use_community`                                       | Which observability strategy to use for deployment. Default deployment is 'use_community'. Also supported are 'use_redhat', 'use_hybrid', and 'none' |
```
leifmadsen, line 56: You could update this documentation to match the new defaults which seem to be use_hybrid going forward right?

roles/servicetelemetry/tasks/component_alertmanager.yml
```diff
@@ -28,6 +28,26 @@
...
+  k8s:
+    state: absent
+    api_version: monitoring.coreos.com/v1
+    kind: alertmanager
+    name: '{{ ansible_operator_meta.name }}'
+    namespace: '{{ ansible_operator_meta.namespace }}'
+  when:
+    - observability_strategy != "use_community"
+
+- name: Ensure no rhbos Alertmanager is installed if not using it
```
leifmadsen, line 41:
```suggestion
- name: Ensure no rhobs Alertmanager is installed if not using it
```

build/stf-run-ci/README.md
leifmadsen: Default in stf-run-ci is defaulting to use_hybrid (mid-term solution I guess to make sure we're covering our non-Red Hat scoped areas which will be migrated ideally in OCP 4.14+), so I think that does look to be the value here vs use_redhat which is the default for new STF installations going forward.

build/stf-run-ci/defaults/main.yml
```diff
@@ -26,7 +26,7 @@ __service_telemetry_snmptraps_trap_oid_prefix: "1.3.6.1.4.1.50495.15"
 __service_telemetry_snmptraps_trap_default_oid: "1.3.6.1.4.1.50495.15.1.2.1"
 __service_telemetry_snmptraps_trap_default_severity: ""
 __service_telemetry_logs_enabled: false
-__service_telemetry_observability_strategy: use_community
+__service_telemetry_observability_strategy: use_hybrid
```
leifmadsen, line 29: So stf-run-ci defaults to use_hybrid (keeps our existing test structure currently, but uses OBO for metrics storage vs community catalogsource for Prometheus), but then I noticed the default value in roles/servicetelemetry/defaults/main.yml is targeting use_redhat (for new installations).

I was initially a bit confused, but I get it now, and just wanted to write this down to verify.

New STF installations will have use_redhat enabled because that's what the default/main.yml will have, along with the implicit value set in the CR (which in turn also populates the OCP UI interface when you create an STF object from the console).

We will test via use_hybrid for now to keep our larger test coverage, but it will default to using OBO instead of community prometheus as the metrics data store.

I went through the code here, and it looks like use_hybrid is basically use_community (components wise) but OBO vs Prometheus Community.

There is the other PR #422 to narrow the ES scope, but if stf-run-ci is using use_hybrid, doesn't that mean we have all the necessary components to test events storage? I feel like I'm missing something here and I can't see it.

roles/servicetelemetry/tasks/pre.yml
```diff
@@ -79,6 +79,54 @@
...
+  - name: Apply community observability_strategy if missing on an STF object with an existing community prometheus
+    k8s:
+      definition:
+        apiVersion: infra.watch/v1beta1
+        kind: ServiceTelemetry
+        metadata:
+          name: "{{ ansible_operator_meta.name }}"
+          namespace: "{{ ansible_operator_meta.namespace }}"
+        spec:
+          observability_strategy: use_community
```
leifmadsen, line 108: this works because `k8s` module uses a `merge` strategy by default?

leifmadsen (approved):
This all looks good to me. I did leave a couple of comments with questions just to make sure I understood what was going on here correctly. Everything looks consistent here, and I appreciate the clean up task that makes sure the existing ServiceTelemetry objects have their observability strategy set so we're not causing changes on existing deployments.

build/stf-run-ci/defaults/main.yml
csibbitt: Your explanations are all correct. 

As for PR #422 , yes, stf-run-ci does default to use_hybrid, but #422 asks Jenkins to deploy with use_redhat instead and limits the smoke test to only that coverage by default. On THIS PR, I noted that testing with use_hybrid and deploying with use_redhat by default seems like a bad idea. #422 is the solution to that problem, but comes at the cost of removing all ES related testing (ES testing should return when we add the supported forwarding model). I proposed it separately to try to keep this focused and prevent altering the smoketests alongside a large change.

roles/servicetelemetry/tasks/pre.yml
csibbitt: :heavy_check_mark:

csibbitt:
I'm going to merge #422 into this and then try to get it to pass CI tests before ultimately merging this to maste.

csibbitt:
test

csibbitt:
4.12 builds are failing to push to the internal registry in the OCP cluster. I will investigate.

csibbitt:
I found cluster-image-registry-operator spitting errors so I restarted it and it seemed to come up healthy. Shortly thereafter the cluster stopped responding to API calls and I could no longer log in. It came back, and I found the oauth container restarting.

```
I0508 20:26:52.862099       1 controller.go:450] Starting Controller
[...]
ditions.1.message={"Progressing: The registry is ready\nNodeCADaemonProgressing: The daemon set node-ca is deploying node pods" -> "Progressing: The registry is ready\nNodeCADaemonProgr
essing: The daemon set node-ca is deployed"}, changed:status.conditions.1.reason={"NodeCADaemonUnavailable::Ready" -> "Ready"}, changed:status.conditions.1.status={"True" -> "False"}
W0508 20:28:15.831346       1 reflector.go:347] github.com/openshift/client-go/route/informers/externalversions/factory.go:101: watch of *v1.Route ended with: an error on the server ("u
nable to decode an event from the watch stream: stream error: stream ID 111; INTERNAL_ERROR; received from peer") has prevented the request from succeeding
```

```
$ oc get po -A | egrep -v Running\|Completed
NAMESPACE                   NAME                                  READY   STATUS        RESTARTS         AGE
openshift-authentication    oauth-openshift-5fc7b6968c-nh8wq      0/1     Pending       0                4s
openshift-authentication    oauth-openshift-6b456c6dbb-8r8ds      1/1     Terminating   0                125m
openshift-etcd              installer-6-elf-412-5vmkz-master-2    0/1     Error         0                14d
```

It seems to have stabilized and builds are succeeding! Still, something bad going on here.

csibbitt:
@elfiesmelfie I switched the setting here https://github.com/infrawatch/service-telemetry-operator/pull/402#issuecomment-1538891071  but now I think I need to switch it back?

PR #424, opened 2023-05-03
Use stable-v1 cert-manager in CI for OCP >= 4.13

csibbitt:
- cert-manager goes GA in 4.13 and the stable-v1 channel is available
- Older versions must still use the tech-preview channel

### DIFF
--- a/build/stf-run-ci/tasks/setup_base.yml
+++ b/build/stf-run-ci/tasks/setup_base.yml
@@ -50,6 +50,15 @@
           namespace: openshift-cert-manager-operator
         spec: {}
 
+  - name: Get OCP version
+    shell: oc version -o yaml | grep openshiftVersion | awk '{print $2}'
+    register: ocp_ver
+
+  - name: Use tech-preview channge for cert_manager in older OCP versions
+    set_fact:
+      cert_manager_channel: tech-preview
+    when: ocp_ver.stdout is version('4.13', '<')
+
   - name: Subscribe to Cert Manager for OpenShift Operator
     k8s:
       definition:
@@ -59,7 +68,7 @@
           name: openshift-cert-manager-operator
           namespace: openshift-cert-manager-operator
         spec:
-          channel: tech-preview
+          channel: "{{ cert_manager_channel }}"
           installPlanApproval: Automatic
           name: openshift-cert-manager-operator
           source: redhat-operators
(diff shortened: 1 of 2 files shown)


leifmadsen:
I talked with Chris because I wasn't sure if this was necessary due to another PR I have up for properties.yaml. However, the way stf-run-ci works right now, dependency resolution will not happen since there is no Subscription or index image to pull from.

I have some plans to extend the testing to allow for validation from the properties.yaml, but until that is done, this change is required for testing against various OCP versions. At some point in the future I'm hoping to be able to pull a bunch of this logic out, or at least skip in favour of an index/CatalogSource based deployment.

leifmadsen (approved):

csibbitt:
I think that cert-manager is already being made available in 4.12 as well, based on this https://github.com/openshift/openshift-docs/pull/58962

I'll try it and adjust the number in this PR if so.

csibbitt:
test

leifmadsen:
Looks like the scorecard check timed out. GHA now has the new fancy feature where you can just re-run the specific test.

(I see that it got triggered as I was typing this lol)

build/stf-run-ci/tasks/setup_base.yml
```diff
@@ -50,6 +50,15 @@
           namespace: openshift-cert-manager-operator
         spec: {}
 
+  - name: Get OCP version
+    shell: oc version -o yaml | grep openshiftVersion | awk '{print $2}'
+    register: ocp_ver
+
+  - name: Use tech-preview channge for cert_manager in older OCP versions
```
vkmc, line 57: NIT change

build/stf-run-ci/tasks/setup_base.yml
```diff
@@ -59,7 +68,7 @@
           name: openshift-cert-manager-operator
           namespace: openshift-cert-manager-operator
         spec:
-          channel: tech-preview
+          channel: "{{ cert_manager_channel }}"
```
vkmc, line 71: Shouldn't we have a value for OCP >= 4.12? Not sure what is the behavior if this string is empty

csibbitt: The default value is set in defaults/main.yml L61 (see above)

build/stf-run-ci/tasks/setup_base.yml
csibbitt:
```suggestion
  - name: Use tech-preview change for cert_manager in older OCP versions
```

csibbitt, line 57:
```suggestion
  - name: Use tech-preview channel for cert_manager in older OCP versions
```

PR #425, opened 2023-05-04
Pinning prometheus version to ease migration to OBO

csibbitt:
* Currently we are installing :latest
* We and OBO currently install v2.43.0
* This will mitigate any potential for a prometheus roll-back
* We will remove the pin in STF 1.5.5 after migration to OBO is complete


csibbitt:
# Testing

I editted the pin to say v2.42.0 and deployed a local build so that I could simulate OBO wanting to deploy a higher version than our pin (we know it won't be lower).

1. Community prometheus pinned to v2.42.0
    ```
    $ oc get stf -o yaml | grep Strategy:
        observabilityStrategy: use_community

    $ oc get pod prometheus-default-0 -o yaml | grep prometheus:v
        image: quay.io/prometheus/prometheus:v2.42.0
    ```
1. Migrate to OBO which defaults to v2.43.0
    ```
    oc patch stf default --patch '{"spec":{"observabilityStrategy":"use_redhat"}}' --type merge
    ```
1. Confirm we migrated and are still pinned
    ```
    default   v2.42.0   1         1       True         True        30s

    $ oc get pod prometheus-default-0 -o yaml | grep prometheus:v
        image: quay.io/prometheus/prometheus:v2.42.0
    ```
1. Remove the pin and deploy local build (simulate future STF 1.5.5 behavior)
    ```
    $ git diff
    diff --git a/roles/servicetelemetry/templates/manifest_prometheus.j2 b/roles/servicetelemetry/templates/manifest_prometheus.j2
    index 4e8651d..94ceecb 100644
    --- a/roles/servicetelemetry/templates/manifest_prometheus.j2
    +++ b/roles/servicetelemetry/templates/manifest_prometheus.j2
    @@ -7,7 +7,7 @@ metadata:
      name: '{{ ansible_operator_meta.name }}'
    spec:
    -  version: v2.42.0
    +  version:
    ```
    * **NOTE**: The version key needs to remain and be set blank because the manifest will merge with what's in the cluster
1. Observe that the upgrade has occurred
    ```
    $ oc get pod prometheus-default-0 -o yaml | grep prometheus:v
        image: quay.io/prometheus/prometheus:v2.43.0
    ```

I'm also going to need to do the same for Alertmanager....

csibbitt:
It runs out I *don't* need to do this for Alertmanager, because it is not using the latest image like prometheus is, so we are not at risk of getting ahead of the image that OBO supports. It deploys v0.21.0 as seen here: https://github.com/prometheus-operator/prometheus-operator/blob/release-0.47/pkg/operator/defaults.go#L20

I have still been unable to figure out why/how we are defaulting to latest for prometheus to begin with instead of `v2.26.0`: https://github.com/prometheus-operator/prometheus-operator/blob/release-0.47/pkg/operator/defaults.go#L75

I haven't managed to find anywhere in STO, in my STF object, or in the Prometheus object that's deployed that alters the image, tag, or version for prometheus. It seems to be like it SHOULD be requesting v2.26.0, but it's not. I'm noticing that it's not even `:latest`, it's just blank, which suggests this is an image override rather than a tag or version override. I just can't explain where it comes from!

```
$ oc get po prometheus-default-0 -o yaml | grep  image:
    image: quay.io/prometheus/prometheus

$ oc get prometheus -o yaml | grep image:
      image: quay.io/openshift/origin-oauth-proxy:latest

$ oc get prometheus -o yaml | grep tag:
$ oc get prometheus -o yaml | grep version:
```

csibbitt:
The `:latest` (actually untagged) mystery is solved! As I traced through the code further I found that the DefaultPrometheusVersion was previously used for several checks but not actually the tag itself.

We are not the first to be hit by this: https://github.com/prometheus-operator/prometheus-operator/issues/4893 

It looks like the operator behavior changed in operator version 0.61.0.

PR #432, opened 2023-06-02
[zuul] Add jobs to test with different versions of OCP

elfiesmelfie:


ci/prepare.yml
```diff
@@ -4,7 +4,20 @@
     - name: Install pre-reqs from pip
       ansible.builtin.pip:
         name:
+          - openshift
           - kubernetes
+          - ansible-core>=1.12
```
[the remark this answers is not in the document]
elfiesmelfie: ansible and ansible-core have different versions; I will double check that, but I think this is right for ansible-core.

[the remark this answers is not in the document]
leifmadsen: I feel like I've asked this before and was told it was right :) Just that if I run `ansible --version` I get this:

```
ansible [core 2.12.10]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/leif/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.10/site-packages/ansible
  ansible collection location = /home/leif/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.10.11 (main, Apr  5 2023, 00:00:00) [GCC 12.2.1 20221121 (Red Hat 12.2.1-4)]
  jinja version = 3.0.3
  libyaml = True
```

I get this in my `pip list` as well:

```
pip list | grep ansible
ansible-core                 2.12.10
ansible-lint                 6.14.2
```

I guess it says here >= 1.12, so I suppose that means it could result in 2.x series as well? Just looks really odd to me.

.zuul.yaml
```diff
@@ -42,9 +42,55 @@
...
+    parent: stf-base
+    # Little hacky-hack so it won't complain about scenario not being defined.
+    vars:
+      scenario: "zuul-common"
+    run:
+      - ci/check_ocp_version.yml
+    post-run:
+      - ci/post-collect_logs.yml
+
+- job:
```
csibbitt, line 59: Things move pretty fast. All our testing now should be 4.12 and 4.14 only

csibbitt (commented):
Approved in principle, but this will need a merge/rebase and adjustments to cover just 4.12 and 4.14

.zuul.yaml
leifmadsen: In theory if there is the resourcing available, it'd be nice to see this on 4.11 through 4.14, but that's feeling perhaps a bit excessive on the testing ranges, especially upstream.

I also vote for 4.12 and 4.14 only.

elfiesmelfie: Thank you for the feedback.

I'm using multiple versions here so that I can make sure this works across versions :)
I will clean things up. I want to get the local-build_index-deploy job merged first. Then start adding different versions of OCP for testing these different scenarios. 

I would recommend using the versions we officially support, as well as the latest version (to pre-empt problems in the next supported release i.e. test on 4.13 until there's a 4.14 release)

elfiesmelfie: This is still pretty much a PoC to prove that this is do-able. I plan to work on this in the next week or so

elfiesmelfie:
recheck

leifmadsen (approved):
Approving as-is. Doesn't appear there is a 4.14 CRC yet.

elfiesmelfie:
recheck

PR #435, opened 2023-06-06
Use the proper parameter for SNMP timeout

leifmadsen:
The SNMP_TIMEOUT environment variable in the manifest_snmp_traps.j2
template for Service Telemetry Operator uses the wrong parameter,
passing port instead of timeout value to the template.

This change updates the template to use the appropriate parameter as
input.

Closes: rhbz#2213016

### DIFF
--- a/roles/servicetelemetry/templates/manifest_snmp_traps.j2
+++ b/roles/servicetelemetry/templates/manifest_snmp_traps.j2
@@ -28,7 +28,7 @@ spec:
             - name: SNMP_PORT
               value: "{{ servicetelemetry_vars.alerting.alertmanager.receivers.snmp_traps.port }}"
             - name: SNMP_TIMEOUT
-              value: "{{ servicetelemetry_vars.alerting.alertmanager.receivers.snmp_traps.port }}"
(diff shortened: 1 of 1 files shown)


leifmadsen:
I haven't tested this yet. I made this change based on looking at the contents when I was looking at something else to close out an issue. Oops!

csibbitt (approved):

leifmadsen:
test

leifmadsen:
I should probably do a real test here before merging to make sure I have all the proper parameters aligned and that changing the CR results in what I expect. I have that on my list for this sprint and should be able to get this merged soon. Thanks for the review!

leifmadsen:
I was testing spin up of OCP on PSI and it resulted in a working cluster.

I deployed this branch against that cluster and performed some verification.

* deployed from index image
* checked current values for timeout in ServiceTelemetry object and in default-snmp-webhook pod. Noted value for TIMEOUT was 1
* changed ServiceTelemetry object `alerting.alertmanager.receivers.snmpTraps.timeout` from a value of `1` to `5`
* watch STO logs waiting for changes to happen, which they did
* verified default-snmp-webhook pod restarted after changes. Confirmed.
* checked default-snmp-webhook TIMEOUT value was updated to a value of `5`

This issue is verified. I will merge it now.

PR #428, opened 2023-05-19
Add index-based deployment method to stf-run-ci

leifmadsen:
Depends-On: https://github.com/infrawatch/smart-gateway-operator/pull/141

Provides an implementation into stf-run-ci that allows you to locally generate and build bundle images for Service Telemetry Operator and Smart Gateway Operator, then create an index image in your OCP cluster. Creates a CatalogSource so that you can load the custom packagemanifests, and then Subscribe to Service Telemetry Operator without having to pre-Subscribe to other dependent Operators.

Builds on the properties.yaml changes to Service Telemetry Operator which describes the Operator dependencies (other than Elasticsearch), making it so the pre-deployed Operators are no longer necessary.

PR #436, opened 2023-06-07
Allow bundle deploy from upstream container repo

leifmadsen:
Allow deployment from upstream container repo. Found this while testing
with --skip-tags per documentation, but fails to deploy bundles due to
hard-coded flags for authorized registries.


build/stf-run-ci/tasks/pre-clean.yml
```diff
@@ -62,6 +62,26 @@
         publisher: CloudOps
         sourceType: grpc
 
+- name: Remove service-telemetry-operator-bundle CatalogSource (bundle deploy)
```
leifmadsen, line 65: Clean up to allow iterative testing.

build/stf-run-ci/tasks/setup_stf_from_bundles.yml
```diff
@@ -82,8 +82,8 @@
 
 - name: Deploy SGO via OLM bundle
   shell:
-    cmd: "{{ base_dir }}/working/operator-sdk run bundle {{__smart_gateway_bundle_image_path}} --pull-secret-name=pull-secret --ca-secret-name=registry-tls-ca --namespace={{ namespace }} --timeout 600s"
+    cmd: "{{ base_dir }}/working/operator-sdk run bundle {{__smart_gateway_bundle_image_path}} {% if pull_secret is defined %}--pull-secret-name=pull-secret --ca-secret-name=registry-tls-ca{% endif %} --namespace={{ namespace }} --timeout 600s"
```
leifmadsen, line 85: Not exactly happy with this. Considering putting contents above this into two `block` items with tags added to them. When run, have it `set_fact` that I can use to define `pull-secret-name` and `ca-secret-name` separately.

leifmadsen (commented):
Kind of a POC that allows deployment via bundles upstream (didn't want to go through the whole process of doing a downstream bundle test).

Needs additional logic when deploying from bundles directly to pre-deploy, although this might only need to wait for `properties.yaml` to land in a bundle, because it looks like bundles are deployed via a Subscription since bundles are created via a CatalogSource.

I didn't actually test this logic before, thinking it would have just kind of read the bundles and loaded the CSVs directly. The fact this runs via a CatalogSource and Subscription makes me think if `properties.yaml` was in the STO bundle, it would have requested the other Operator dependencies.

leifmadsen:
@csibbitt here is a kind of draft PR that allows testing bundle-based deployments from the upstream bundles since I didn't want to do the full setup for a downstream bundle.

From what I'm looking at, everything seems to be working as it should. The only issue I can see is around no longer installing the pre-provisioned items. I _think_ this is fine once everything merges, because at that point you'd have properties.yaml in the bundle and it should do all the resolving (not verified, but see comment above about why I think that).

In the meantime, everything remains the same in stable-1.5, so for CVEs or other async releases in the meantime, the downstream release pipeline should be able to use the contents fron stable-1.5 branch. Once we import to stable-1.5, we'll be imminently looking to do a release and tagging/importing the properties.yaml changes, meaning I think this can all land now.

I can also do a full verification once this lands. I can trigger the release-automation branch GHA to create bundles, and then I can re-run this test to verify that properties.yaml does what I think it should. In the meantime, I can see if I can make this work by running from a bundle generated in the local environment through the local_build system.

csibbitt (approved):
I think this patch is fine and in fact I would rather see it merge directly to master since it isn't actually dependent on any of the other work in this chain.

csibbitt:
> Kind of a POC that allows deployment via bundles upstream (didn't want to go through the whole process of doing a downstream bundle test).
> 
> Needs additional logic when deploying from bundles directly to pre-deploy, although this might only need to wait for `properties.yaml` to land in a bundle, because it looks like bundles are deployed via a Subscription since bundles are created via a CatalogSource.
> 
> I didn't actually test this logic before, thinking it would have just kind of read the bundles and loaded the CSVs directly. The fact this runs via a CatalogSource and Subscription makes me think if `properties.yaml` was in the STO bundle, it would have requested the other Operator dependencies.

The logic here seems sound to me. What you describe matches my experience - seeing CatalogSource and Subscription objects that are handled by OLM, rather than evidence that operator-sdk itself rips open the bundle to do it's own installation.

leifmadsen:
> I think this patch is fine and in fact I would rather see it merge directly to master since it isn't actually dependent on any of the other work in this chain.

Since I modified this against my other branch, it seems swapping to master does what I should have expected. I am going to open a separate PR with just the diff and have it land against master, and then I'll merge master back into #428 

Expect to see this one close out shortly since I'm not sure it's worth doing an interactive rebase to get this back against master.

leifmadsen:
Closing this in favour of a different PR targeting master.

PR #439, opened 2023-06-12
Support External Elasticsearch

csibbitt:
* Deprecate existing STF-deployed ES
* New CRD properties for external ES
* Adjust SG and Grafana to use new vars


csibbitt:
This passes smoke tests with OBSERVABILITY_STRATEGY=use_hybrid and an STF deployment to match. I can see from the smoketest logs that events were flowing:

```
*** [INFO] Get documents for this test from ElasticSearch...
[2023-06-12 19:12:05] Notification: severity = OKAY, host = stf-smoketest-smoke1-qc9rx, plugin = interface, plugin_instance = lo, type = if_octets, message = Host stf-smoketest-smoke1-qc9rx, plugin interface (instance lo) type if_octets: All data sources are within range again. Current value of "rx" is 0.000000.
*** [INFO] Found 34 documents
```

This means that all the new ext* usage in the templates and the default values are all aligned and haven't broken anything. I haven't actually tested connecting to a non-STF deployed ES yet, though.

csibbitt:
I just tested grafana (same use_hybrid setup as above) and it works after some fixes. Both ES data sources pass the "Save and Test" in Grafana.

Working on a true test of a separately deployed ES next.

deploy/crds/infra.watch_servicetelemetrys_crd.yaml
```diff
@@ -172,44 +172,62 @@ spec:
...
+                            type: string
+                          extUseBasicAuth:
+                            description: Whether to provide HTTP Basic Auth headers
+                            type: boolean
+                          extUseTls:
+                            description: Whether to enable TLS
+                            type: boolean
                           version:
-                            description: Version of ElasticSearch to deploy. Elasticsearch licensing has changed as of version 7.11. See https://www.elastic.co/pricing/faq/licensing for details.
+                            description: [DEPRECATED - Use ext* after STF 1.5.3] Version of ElasticSearch to deploy. Elasticsearch licensing has changed as of version 7.11. See https://www.elastic.co/pricing/faq/licensing for details.
```
leifmadsen, line 194: I know we have a bunch of ElasticSearch items here, but it should actually be Elasticsearch. I'm not overly worried about it. I'm not sure if the branding changed at some point or I just always did this wrong in the past.

roles/servicetelemetry/defaults/main.yml
```diff
@@ -46,6 +46,12 @@ servicetelemetry_defaults:
     events:
       elasticsearch:
         enabled: false
+        extHostUrl: https://elasticsearch-es-http:9200
+        extTlsServerName: elasticsearch-es-http.{{ ansible_operator_meta.namespace }}.svc.cluster.local
+        extTlsSecretName: elasticsearch-es-cert
+        extUserSecretName: elasticsearch-es-elastic-user
+        extUseBasicAuth: true
+        extUseTls: true
```
leifmadsen, line 54: Haha, this is the case where you want to use the snake_case (it's in the Ansible now and the conversion has happened.

ext_host_url
ext_tls_server_name
etc...

csibbitt: Awww, damnit! I'll need to switch the templates too. Gets me every time. You'd think I'd learn, but I guess not.

roles/servicetelemetry/tasks/component_clouds.yml
```diff
@@ -19,20 +19,51 @@
...
 - name: Events Smart Gateway deployment
+  when: servicetelemetry_vars.backends.events.elasticsearch.enabled
   block:
     - name: Lookup ElasticSearch BasicAuth
       k8s_info:
         api_version: v1
         kind: Secret
         namespace: '{{ ansible_operator_meta.namespace }}'
-        name: 'elasticsearch-es-elastic-user'
+        name: '{{ servicetelemetry_vars.backends.events.elasticsearch.extUserSecretName }}'
```
leifmadsen, line 29: 🐍

roles/servicetelemetry/vars/dummy_user_certs.yml
```diff
@@ -0,0 +1,56 @@
+# These are required because sg-core always expects there to be a user cert + key, whether we need it for auth or not
+# CN = STF DUMMY USER CERT - DO NOT USE
+elastic_user_cert_dummy: |
```
leifmadsen, line 3: Just because I wasn't sure, I looked up if there was a chomp indicator needed here (e.g. `|-` to strip any newlines at the end) and I don't think that's necessary. Noting because I thought of it.

roles/servicetelemetry/defaults/main.yml
leifmadsen: It's definitely a tricky rule!

roles/servicetelemetry/tasks/component_clouds.yml
```diff
@@ -19,20 +19,51 @@
...
+    - when: es_certs[0].data[user.key] is not defined or es_certs[0].data[user.crt] is not defined
+      block:
+        - name: Load dummy certs
+          include_vars:
+            file: dummy_user_certs.yml
+
+        - name: Augment the secret with dummy TLS cert/key if no TLS user auth material provided
+          k8s:
+            definition:
+              api_version: v1
```
leifmadsen, line 58: When you use `definition` you use the actual k8s manifest format, so in this case it's `apiVersion` and not the Ansible parameter `api_version`.

roles/servicetelemetry/tasks/component_clouds.yml
```diff
@@ -19,20 +19,51 @@
...
+        kind: Secret
+        namespace: '{{ ansible_operator_meta.namespace }}'
+        name: '{{ servicetelemetry_vars.backends.events.elasticsearch.extTlsSecretName }}'
+      register: es_certs
+
+    - when: es_certs[0].data[user.key] is not defined or es_certs[0].data[user.crt] is not defined
+      block:
+        - name: Load dummy certs
+          include_vars:
+            file: dummy_user_certs.yml
```
leifmadsen, line 53: Do you want to maybe add a check here to see if only one of these is set? If that's the case I would think this include_vars overrides the user-passed data, which I think makes sense because you should probably be passing in both values not just one.

csibbitt: The 'when' clause is already such that both must be supplied, otherwise both will be replaced by the dummies. I think this is what you're suggesting, but I might have misunderstood.

csibbitt:
This has now been tested with an independently deployed Elasticsearch and `observabilityStrategy: use_redhat`. I deployed Elasticsearch via ECK in another namespace and created a route that leads to it so I could connect the smartgateway via that "external" route.

Here is my ES config in the STF object:

```
$ oc get servicetelemetry default -o jsonpath='{.spec.backends.events.elasticsearch}' | jq
{
  "enabled": true,
  "extHostUrl": "https://es-route-es.apps.ocp413-csibbs.<...>.redhat.com",
  "extTlsSecretName": "es-ca",
  "extUseBasicAuth": true,
  "extUseTls": true,
  "extUserSecretName": "es-user"
}
```

Construct the TLS secret like this (just the required CA cert here, but it can also contain an mTLS cert + key):

```
$ openssl s_client -servername es-route-es.apps.ocp413-csibbs.<...>.redhat.com -connect es-route-es.apps.ocp413-csibbs.<...>.redhat.com:443 -showcerts

[ COPY/PASTE THE SIGNING CERT INTO ES-CA.pem ]

$ cat ES-CA.pem | openssl x509 -text | grep CN
        Issuer: CN = ingress-operator@1680638728
        Subject: CN = ingress-operator@1680638728

$ oc create secret generic es-ca --from-file=ca.crt=ES-CA.pem
secret/es-ca created
```

Construct the User Secret for BasicAuth like this:

```
 oc create secret generic es-user --from-literal=elastic=$ESPASS
```

I get a working system with two events SGs both connected to Elasticsearch:

```
$ oc get po | grep event
default-cloud1-ceil-event-smartgateway-7f9986698c-t6js9   2/2     Running     0          5m
default-cloud1-coll-event-smartgateway-656dd6b875-2fpdn   2/2     Running     0          5m

$ oc logs default-cloud1-ceil-event-smartgateway-7f9986698c-t6js9 -c sg-core
2023-06-20 03:37:36 [INFO] initialized handler [transport pair: socket0, handler: events]
2023-06-20 03:37:36 [INFO] loaded transport [transport: socket0]
2023-06-20 03:37:36 [INFO] loaded application plugin [application: elasticsearch]
2023-06-20 03:37:36 [INFO] storing events and(or) logs to Elasticsearch. [plugin: elasticsearch, url: https://es-route-es.apps.ocp413-csibbs.<...>.redhat.com]
2023-06-20 03:37:36 [INFO] socket listening on /tmp/smartgateway [plugin: socket]
```

Unfortunately the smoketest won't pass against my external ES setup because it has hard-coded endpoint and secret name values. (Existing automated smoketests in use_hybrid mode should continue to work fine with the defaults)

csibbitt:
test

csibbitt:
Depends-On: https://github.com/infrawatch/smart-gateway-operator/pull/142

roles/servicetelemetry/templates/manifest_smartgateway_events.j2
```diff
@@ -1,27 +1,28 @@
...
+      hostURL: {{ elasticsearch.ext_host_url }}
+      useTLS: {{ elasticsearch.ext_use_tls }}
+      tlsClientCert: /config/certs/user.crt
+      tlsClientKey: /config/certs/user.key
       tlsCaCert: /config/certs/ca.crt
-      tlsServerName: 'elasticsearch-es-http.{{ ansible_operator_meta.namespace }}.svc.cluster.local'
-      user: '{{ elastic_user | default('elastic') }}'
-      password: '{{ elastic_pass | default('') }}'
-      useBasicAuth: true
+      tlsServerName: {{ (elasticsearch.ext_tls_server_name != "") | ternary(elasticsearch.ext_tls_server_name, elasticsearch.ext_host_url.rpartition("//")[-1].partition(":")[0]) }}
```
leifmadsen, line 21: I wasn't sure if `rpartition` was a typo, but it does not appear to be. See https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html#referencing-key-value-dictionary-variables

I also saw this note which was interesting:

> Use bracket notation if you use keys which start and end with two underscores (which are reserved for special meanings in python) or are any of the known public attributes:

It might be worth a note here just to document exactly what is being parsed here for future reference. It's complex enough that my brain hurts (although it's almost lunch).

leifmadsen:
Going to have to spend some time going through the changes and looking at them, but I did another quick pass for anything syntactically weird, and all of that looks good to me.

roles/servicetelemetry/templates/manifest_smartgateway_events.j2
csibbitt: Hmm, I can't add a comment inside a template, but could move this logic into a pair of tasks to make them clearer and comment there? It probably won't be today, but I can come back and add that.

To help others, this is the logic spelled out:

- If ext_tls_server_name is not empty, use it
- Else, use the ext_host_url, returning the substring to the right of '//' and the left of ':'

deploy/crds/infra.watch_servicetelemetrys_crd.yaml
```diff
@@ -167,49 +167,67 @@ spec:
...
                     properties:
                       elasticsearch:
-                        description: Events storage backend ElasticSearch
+                        description: Events storage backend Elasticsearch
                         properties:
                           enabled:
-                            description: Enable ElasticSearch as a storage backend for events
+                            description: Enable Elasticsearch as a storage backend for events
+                            type: boolean
+                          extHostUrl:
```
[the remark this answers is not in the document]
leifmadsen: I see your points and logic of how you got here :)

I'm going to think about it some and will get back to you.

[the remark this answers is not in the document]
leifmadsen: Looking at the CRD there are two suggestions I came up with. Simply noting as a discussion point.

I was looking at the `storage` section and thinking about adding a new strategy called `- external` and then adding an `external` section (like we have a `persistent` section). Thinking some more about that I thought it might be awkward and not very intuitive.

I then thought about the `external` subsection again, and was wondering if maybe the naming is the awkward bit? What about instead of external we use a section called `forwarding`? It doesn't really change anything about the syntax, but would give an opportunity to add another `description` section that would explain what the purpose of the configurations would be.

There is also the option where we can remove the old syntax from the CRD, but still allow it to be passed in and consumed for the lifecycle of STF 1.5 so that we don't totally break any existing systems. That would at least help clean up the CRD itself and the UI interface. It might even be something to consider for the first round where the old CRD parameters are removed, but still accepted by the Ansible roles.

Ideally we can accept the old syntax but then have the code basically accept the existing managed ES that we are deploying as the new forwarded endpoint without requiring an administrator to remove it, but at which point we'd drop all management responsibility and simply have the `forwarding` (`external`, `ext<param>`, ...) section be populated from the old params.

Hope that makes sense. Maybe when you're back we can all get together during one of backlog grooming meetings or otherwise and discuss?

csibbitt:
test

deploy/crds/infra.watch_servicetelemetrys_crd.yaml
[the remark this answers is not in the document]
csibbitt: Implementing this as a `forwarding` section

csibbitt:
test

csibbitt:
I think I'm done with this now. I would appreciate additional review so I can merge it. It's passing CI and it passes my local testing as shown above (I redid basically the same thing with the newly moved parameters). I will be following up this work with adjustments to CI testing that actually exercise the events system again based on this work (currently we are just testing metrics)  and any additional problems should get shaken out during that work.

leifmadsen:
> I think I'm done with this now. I would appreciate additional review so I can merge it. It's passing CI and it passes my local testing as shown above (I redid basically the same thing with the newly moved parameters). I will be following up this work with adjustments to CI testing that actually exercise the events system again based on this work (currently we are just testing metrics) and any additional problems should get shaken out during that work.

I have this on my list of tasks for tomorrow. I will be sure to go through this thoroughly before I am out on Friday.

deploy/crds/infra.watch_servicetelemetrys_crd.yaml
```diff
@@ -167,49 +167,71 @@ spec:
...
+                                type: string
+                              useBasicAuth:
+                                description: Whether to provide HTTP Basic Auth headers
+                                type: boolean
+                              useTls:
+                                description: Whether to enable TLS
+                                type: boolean
                           version:
-                            description: Version of ElasticSearch to deploy. Elasticsearch licensing has changed as of version 7.11. See https://www.elastic.co/pricing/faq/licensing for details.
+                            description: (DEPRECATED - Use ext* after STF 1.5.3) Version of Elasticsearch to deploy. Elasticsearch licensing has changed as of version 7.11. See https://www.elastic.co/pricing/faq/licensing for details.
```
leifmadsen, line 198: I think all the descriptions need to be updated here. They refer to `ext` prefix which you recently removed.

deploy/crds/infra.watch_servicetelemetrys_crd.yaml
```diff
@@ -167,49 +167,71 @@ spec:
...
                         properties:
                           enabled:
-                            description: Enable ElasticSearch as a storage backend for events
+                            description: Enable Elasticsearch as a storage backend for events
                             type: boolean
+                          forwarding:
+                            description: Configuration for where to forward events
+                            type: object
+                            properties:
+                              hostUrl:
```
leifmadsen, line 179: Note: following results due to code walk through, so some questions and review path happens below, and I think I answer most of my own questions. I think commenting inline at each of the paths would have been more confusing.

~~~

Could the `hostUrl` just be the Service name here and resolve internally to the existing managed ES? If you had a previously deployed ES and removed the `version`, `storage`, etc fields, and inserted the Service name (or one of the resolved-locally DNS names?) could you do an in-place migration to orphan the existing ES from management?

That was one of the items I wanted to test and look for in the code base as I'm not exactly sure what will trigger an `absent` value (looks)... so just setting enabled: true will create the ES manifest: https://github.com/infrawatch/service-telemetry-operator/blob/9eacaf4000b9bf20d21f87ebe48e3ba7613eb2ea/roles/servicetelemetry/tasks/component_elasticsearch.yml#L12

OK, so the creation and other items only happens when strategy is `use_redhat`. OK that makes sense. If you change to `use_redhat`, I don't see anything that will trigger the `state: absent`, so ES will remain it seems.

In this section for grafana (https://github.com/infrawatch/service-telemetry-operator/pull/439/files#diff-720dce1475e3129001153c472f39b8f203f25ebfff3e61cc92c503ba4f5a854cL64-L77) I think it might make sense to put this into a block to still exist for `use_community, use_hybrid`?

I think it might make sense to have the hostUrl (https://github.com/infrawatch/service-telemetry-operator/pull/439/files#diff-468bf2d79dac29401233b72f3937e9f68cb9e7049af7bec4d3969c8eb69affe3R48) default to https://elasticsearch-es-http:9200 which I noticed in the `defaults/main.yml` you set to, so that makes sense. You have a couple other example spots, and it's probably worth to just set the value to that everywhere.

https://github.com/infrawatch/service-telemetry-operator/pull/439/files#diff-10e9e317481eb8f22725d09cf3555482d690ff87e2b60861ea51467669b4da9fR32 would then get the proper value if the CR was created with defaults everywhere.

Only last thing I think to look at is https://github.com/infrawatch/service-telemetry-operator/pull/439/files#diff-b661f3e2be8a3856711ae0ed99adda16e30a66a972ec5dc5118f57bfbf3ffe2dR16-R25 which uses user.key and user.crt instead of the former tls.crt and tls.key. Does the certificates generated by our existing deployment contain those fields and would their existing values work against the pre-deployed system?

Now that I've walked through this I want to do a quick deploy and switch from use_community to use_redhat. If it doesn't work, it looks really close to working.

leifmadsen: Last thought here. There could potentially be a KCS for anyone who wanted to stand up an ES in a similar format to how we were doing it via managed. That gets us off the hook for managing, but still gives a way to get the system operational how it was before, and removing from product docs, but still have a sort of example configuration that the defaults work against. Possibly reduces some maintenance burden and gives someone a place to start, or at least easier for us to build into CI.

csibbitt: "I think commenting inline at each of the paths would have been more confusing." Oh yeah? You won't mind if I meatloaf my responses, then, too, right?

> Could the hostUrl just be the Service name here and resolve internally to the existing managed ES?

Yes

> In this section for grafana ...  a block to still exist for use_community, use_hybrid?

I'm not sure what your suggesting exactly. Currently the grafana tasks are only included `in ['use_community', 'use_hybrid']` https://github.com/infrawatch/service-telemetry-operator/blob/master/roles/servicetelemetry/tasks/main.yml#L101  and the block you call out only run if there is an elasticsearch to configure. https://github.com/infrawatch/service-telemetry-operator/pull/439/files#diff-720dce1475e3129001153c472f39b8f203f25ebfff3e61cc92c503ba4f5a854cL64 Are you suggesting something else?

> hostUrl default to https://elasticsearch-es-http:9200/ [but CRD defaults it to https://external-elastic-http.domain:9200"]

Maybe I need to take a second look at this in the context of the "in place migration" like you mention. The example in the CSV is to emphasize that it is an EXTERNAL ES. If you are creating a fresh STF instance with default values, that internal service name is not going to be what you want, either, you absolutely must fill it in. I will test what happens when you upgrade from an old STF version - I think you might be right that the migration will be smoother if the default value matches the legacy value.

> Does the certificates generated by our existing deployment contain those fields and would their existing values work against the pre-deployed system?

No and no. However the dummy cert/key that is included when these fields are missing takes care of that. To clarify - our current setup does not rely on certificate authentication, but the SG has it hard-coded as being required. Previously it supplied the cert/key used by ES itself, which is nonsensical. So now we just supply the dummy cert instead.

> There could potentially be a KCS for anyone who wanted to stand up an ES in a similar format to how we were doing it via managed

Sure. I've added a note to my ticket about CI work, since I'll have all the artifacts in front of me when I do that.

Bottom line:

- I will test in-place migration from use_hybrid (w/ ES) -> use_redhat (w/ existing but now unmanaged ES) 
- A KCS article will cover how to set up a community ES and configure STF to point at it

deploy/crds/infra.watch_servicetelemetrys_crd.yaml
csibbitt: Good catch, I'll send a patch soon

deploy/crds/infra.watch_servicetelemetrys_crd.yaml
leifmadsen: > "I think commenting inline at each of the paths would have been more confusing." Oh yeah? You won't mind if I meatloaf my responses, then, too, right?

Not at all.

> > Could the hostUrl just be the Service name here and resolve internally to the existing managed ES?
> 
> Yes

✅ 

> 
> > In this section for grafana ...  a block to still exist for use_community, use_hybrid?
> 
> I'm not sure what your suggesting exactly. Currently the grafana tasks are only included `in ['use_community', 'use_hybrid']` https://github.com/infrawatch/service-telemetry-operator/blob/master/roles/servicetelemetry/tasks/main.yml#L101 and the block you call out only run if there is an elasticsearch to configure. https://github.com/infrawatch/service-telemetry-operator/pull/439/files#diff-720dce1475e3129001153c472f39b8f203f25ebfff3e61cc92c503ba4f5a854cL64 Are you suggesting something else?

Ah I missed that part. I know the events datasource is only used in the virtual machine view dashboard for VM create/delete annotations. We disabled the annotations by default, but if someone was using use_community, and then enabled the annotations in that dashboard, would they still work without this block?

> > hostUrl default to https://elasticsearch-es-http:9200/ [but CRD defaults it to https://external-elastic-http.domain:9200"]
> 
> Maybe I need to take a second look at this in the context of the "in place migration" like you mention. The example in the CSV is to emphasize that it is an EXTERNAL ES. If you are creating a fresh STF instance with default values, that internal service name is not going to be what you want, either, you absolutely must fill it in. I will test what happens when you upgrade from an old STF version - I think you might be right that the migration will be smoother if the default value matches the legacy value.

I didn't noticed anything that would necessarily cause an issue, but it seems like having everything match at least makes things more consistent, and if you needed to change the value anyways, I see no reason to make it different.
 
> > Does the certificates generated by our existing deployment contain those fields and would their existing values work against the pre-deployed system?
> 
> No, the fields do not exist in the tls material generated by our existing deployments. However the dummy cert/key that is included when these fields are missing takes care of that. To clarify - our current setup does not rely on certificate authentication, but the SG has it hard-coded as being required. Previously we supplied the cert/key used by ES itself, which is nonsensical. So now we just supply the dummy cert instead.

OK thanks for clarifying that.

> > There could potentially be a KCS for anyone who wanted to stand up an ES in a similar format to how we were doing it via managed
> 
> Sure. I've added a note to my ticket about CI work, since I'll have all the artifacts in front of me when I do that.

✅ 

> Bottom line:
> 
> * I will test in-place migration from use_hybrid (w/ ES) -> use_redhat (w/ existing but now unmanaged ES)
> * A KCS article will cover how to set up a community ES and configure STF to point at it

I can create a separate issue in the backlog for the KCS if you want? I don't think that really fits into any of the existing scoped work.

deploy/olm-catalog/service-telemetry-operator/manifests/infra.watch_servicetelemetrys_crd.yaml
```diff
@@ -122,63 +122,97 @@ spec:
...
+                        description: Events storage backend Elasticsearch
                         properties:
                           certificates:
                             properties:
                               caCertDuration:
-                                description: The requested 'duration' (i.e. lifetime)
-                                  of the ElasticSearch CA Certificate. Minimum accepted
-                                  duration is 1 hour. Value must be in units accepted
-                                  by Go time.ParseDuration https://golang.org/pkg/time/#ParseDuration
+                                description: (DEPRECATED - Use ext* after STF 1.5.3)
```
leifmadsen, line 130: Just need to run `operator-sdk generate bundle --channel unstable --default-channel unstable` I believe to update the deploy/ directory with the updated contents from deploy/crds/

deploy/crds/infra.watch_servicetelemetrys_crd.yaml
csibbitt: >  if someone was using use_community, and then enabled the annotations in that dashboard, would they still work without this block?

If there is an ES configured, then the block _will_ be included and the annotations will work. I didn't test the annotations themselves, but the "Save and test" in the grafana datasource config worked.

> I see no reason to make it different

Okay, agreed.

> create a separate issue in the backlog

STF-1448

csibbitt:
Before I merge this, I'd like to do the following test:

1. Install STO from master and deploy w/ use_community (like existing STF 1.5 installs)
1. Update to the version from this branch
1. Change to use_redhat
1. Check that everything continues to work
1. Remove the ES config (other than "enabled") - Expect no effect, continues to work

leifmadsen (approved):
I plan on doing a round of testing of STF as it sits during our next sprint. Code looks good to me and should merge down. We can continue to test this and open issues if we find any. Thanks for working this so thoroughly!

csibbitt:
> * I understand that we are deprecating all the params related to STF managed ES. I assume that after some releases, we will remove those params from this CRD. Is there a way to mark those params as deprecated officially? 

Yes, the intention is probably to remove support for them in STF 1.5.4. I don't know of any way to mark them deprecated "offifically" but I'll do some more spec reading and maybe ask someone who is closer to OLM.

> Should we update versioning somewhere?

Haha, yes, probably. It looks like our CRD is still saying "v1beta1". I am personally a big fan of using semantic versioning to signal feature additions and backwards-incompatibility, but AFAICT we have never observed any sort of conventions here for our CRDs, and I feel like other operators are in the same boat. Maybe I've just been ignorant. @leifmadsen do you have any idea about this? Should I bump this to "v1beta2", for example, and then we make it "v2" when we remove support for managing the community ES deployment?

> * What would be the preferred upgrade path for STF users? Do we expect them to create an external instance for ES, migrate they data, and then point the STF object to that ES instance?

For existing users with existing event data, upgrading will cause STF to stop managing the ES object, but otherwise nothing else will change. Users will be able to modify the ES object and rely on the community ECK operator to deal with it instead of having it flow through STF.

> * I see Leif and you discussing the three scenarios of hybrid, community or redhat, why do we make that distinction? External is external, no matter what is the source, no?

We are talking about the `observability_strategy` parameter, which can be set to:
`none` - only deploy smart gateways, no PromOp, no ECK, no grafana
`use_community` - Community PromOP, ECK, and grafana - this is the current way that all STF 1.5.1 installs are done
`use_redhat` - PromOp from RHOBO, no ECK, no grafana - this is the "fully supported install" we are trying to achieve
`use_hybrid` - PromOp from RHOBO, plus Community ECK, and grafana - ECK integration remains deprecated and support for managing the ES object will be removed from a future release.

Forwarding to an external ES should work in all cases. The testing and discussion has been around ensuring that is the case.

leifmadsen:
> > * I understand that we are deprecating all the params related to STF managed ES. I assume that after some releases, we will remove those params from this CRD. Is there a way to mark those params as deprecated officially?
> 
> Yes, the intention is probably to remove support for them in STF 1.5.4. I don't know of any way to mark them deprecated "offifically" but I'll do some more spec reading and maybe ask someone who is closer to OLM.

I think this is something we can bring to the PM. I'm not sure if we should remove functionality in STF 1.5 or not. I would agree that STF 1.5.4 is probably the proper target if we communicate it correctly. Having the ability to "orphan" the existing managed deployment during the STF 1.5.3 cycle and updating the CR in-place (manually or otherwise) to switch to a forwarding mechanism should allow us to do this. We can then remove the related Ansible tasks and simplify the CRD.

I'm adding it to my list of discussion items for next week with PM.

> > Should we update versioning somewhere?
> 
> Haha, yes, probably. It looks like our CRD is still saying "v1beta1". I am personally a big fan of using semantic versioning to signal feature additions and backwards-incompatibility, but AFAICT we have never observed any sort of conventions here for our CRDs, and I feel like other operators are in the same boat. Maybe I've just been ignorant. @leifmadsen do you have any idea about this? Should I bump this to "v1beta2", for example, and then we make it "v2" when we remove support for managing the community ES deployment?

So I am also a fan of semantic versioning where possible, but I've found CRDs are a bit different. I swear I had found something at one point that somewhat discouraged changing CRD versions in certain ways, like simply extending the API interface, and only changing the version if the interface changed in such a way that existing CRs would no longer be compatible. I have yet to find such a document. Perhaps that was a Leif-ism...

At one point I did change the version of the CRD. It moved from v1alpha2 to v1beta1. Unfortunately that was a bit of a nightmare, because Ansible Operators don't have the ability to do webhook conversions, which is the interface that allows you to write a controller and do automatic in-place migrations of Custom Resources to the latest CRD.

See https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/

Also since we have to self-manage the CRD instead of having them generated with decorators in the code, along with some other nice things you get when running a Golang operator, it seemed to make the version changing more overhead than it seemed to be worth.

So _should_ we have better CRD versioning and not be afraid of increasing the versions when we make significant changes? Likely yes. Is it really worth the overhead to self-manage and make sure we don't break existing production deployments during a z-stream? Possibly not.

With that being said, I think STF 2.0 should be a fresh start to our CRD, and we should absolutely create a v2 CRD and remove items we consider legacy, and come up with a possibly cleaner interface. We'll need to have an architectural discussion about how we'll migrate from v1 to v2 CRD, since we're not going to have the benefits of having webhook conversions to do manage the migrations for us.

> > * What would be the preferred upgrade path for STF users? Do we expect them to create an external instance for ES, migrate they data, and then point the STF object to that ES instance?
> 
> For existing users with existing event data, upgrading will cause STF to stop managing the ES object, but otherwise nothing else will change. Users will be able to modify the ES object and rely on the community ECK operator to deal with it instead of having it flow through STF.

+1

New deployments would configure the `forwarding` section along with setting `observabilityStrategy: use_redhat` which gets them OBO + unmanaged ES (forwarding). As of STF 1.5.4 I think we can just remove the community bits from our code base and simply have use_redhat and none as the observability strategies.

This is something else I'll add to the PM discussion though to make sure there aren't any issues with that. The primarily problem I can think of is because the Operator will self-upgrade on publish to the index image and OLM pulling a new index image periodically, is some potential issue we cause on a product deployment. Having `use_redhat` as the default strategy and some well worded documentation around our support methods, but leaving the existing strategies seems possibly valid as well though. That might give customers a bit more time to plan a migration if necessary, even if we effectively EOL the use_hybrid and use_community strategies from downstream product.

If we plan to EOL STF 1.5 say in 2024-Q3 or Q4 in favour of an STF 2.0, it might not necessarily be worth dropping all the existing functionality, since we'd be doing that anyways shortly after the final STF 1.5 release and it goes into maintenance mode prior to STF 2.0 GA. Nothing is set in stone here though, and we have time to discuss the pros and cons of each of the approaches and then make a decision, and communicate that to the field.

> > * I see Leif and you discussing the three scenarios of hybrid, community or redhat, why do we make that distinction? External is external, no matter what is the source, no?
> 
> We are talking about the `observability_strategy` parameter, which can be set to: `none` - only deploy smart gateways, no PromOp, no ECK, no grafana `use_community` - Community PromOP, ECK, and grafana - this is the current way that all STF 1.5.1 installs are done `use_redhat` - PromOp from RHOBO, no ECK, no grafana - this is the "fully supported install" we are trying to achieve `use_hybrid` - PromOp from RHOBO, plus Community ECK, and grafana - ECK integration remains deprecated and support for managing the ES object will be removed from a future release.
> 
> Forwarding to an external ES should work in all cases. The testing and discussion has been around ensuring that is the case.

+1, nothing to add here.

csibbitt:
# (simulated) Upgrade testing

Start from super clean. Not sure where packageserver comes in, but nothing remotely STF-deps related here.

```
$ oc get csv -A
NAMESPACE                              NAME            DISPLAY          VERSION   REPLACES   PHASE
openshift-operator-lifecycle-manager   packageserver   Package Server   0.19.0               Succeeded

$ oc get sub -A
No resources found

$ ansible-playbook build/run-ci.yaml -e __local_build_enabled=true -e __service_telemetry_snmptraps_enabled=true -e __service_telemetry_storage_ephemeral_enabled=true
[...]
        "* [info] CI Build complete. You can now run tests."
```

ES and Event SGs are healthy.

```
$ oc get elasticsearch
NAME            HEALTH   NODES   VERSION   PHASE   AGE
elasticsearch   yellow   1       7.16.1    Ready   3h42m

$ oc get po | grep event
default-cloud1-ceil-event-smartgateway-7549f78d75-nkwz6   2/2     Running     2 (20m ago)    41m
default-cloud1-coll-event-smartgateway-85d4f5d484-4xr2x   2/2     Running     10 (19m ago)   41m
```

Rebuild and restart

```
$ git checkout csibbitt_STF-1385_external_es 

$ ./build/build.sh 

$ oc delete po service-telemetry-operator-5fc75d5986-s64d7
pod "service-telemetry-operator-5fc75d5986-s64d7" deleted

$ oc get po | grep service-telemetry-operator
service-telemetry-operator-2-build                        0/1     Completed   0                52m
service-telemetry-operator-3-build                        0/1     Completed   0                6m17s
service-telemetry-operator-5fc75d5986-qgngj               1/1     Running     0                40s
```

Switch to fully supported mode, observe the new prometheus container

```
$ oc patch stf default --patch '{"spec":{"observabilityStrategy":"use_redhat"}}' --type merge
servicetelemetry.infra.watch/default patched

$ oc get po | grep prometheus-def
prometheus-default-0                                      3/3     Running     0               38s

default   v2.43.0   1         1       True         True        14m
```

ES and Event SGs are healthy.

```
$ oc get elasticsearch
NAME            HEALTH   NODES   VERSION   PHASE   AGE
elasticsearch   yellow   1       7.16.1    Ready   58m

$ oc get po | grep event
default-cloud1-ceil-event-smartgateway-ccd57c969-m868p    2/2     Running     0               5m17s
default-cloud1-coll-event-smartgateway-86f6f76dcd-h6vwr   2/2     Running     0               5m18s

Delete all the ES config

$ oc edit stf default
[...]
  backends:
    events:
      elasticsearch:
        enabled: true
servicetelemetry.infra.watch/default edited
```

ES and Event SGs are healthy.

```
$ oc get elasticsearch
NAME            HEALTH   NODES   VERSION   PHASE   AGE
elasticsearch   yellow   1       7.16.1    Ready   7m

$ oc get po | grep event
default-cloud1-ceil-event-smartgateway-ccd57c969-m868p    2/2     Running     0               14m
default-cloud1-coll-event-smartgateway-86f6f76dcd-h6vwr   2/2     Running     0               14m
```

So I think it works the way we want. Note that I didn't have much luck getting it to pass smoke tests at the initial stage, but I must be doing something wrong since they pass in our actual CI.... but fair warning. :)

csibbitt:
test

vkmc (approved):
Thanks for clarifying, it looks good to me :+1:

PR #443, opened 2023-07-12
Updates stf-run-ci setup playbook to use cert-manager in stable-v1

vkmc:
For newer OCP versions (OCP >= 4.12), cert-manager should be consumed from the stable-v1 channel. This differs from previous versions of OCP, in which we were consuming cert-manager from the tech-preview channel.

The way of installing both operators is different (the tech-preview requires a specific namespace and operator group, whereas the stable-v1 version does not).

This change adds a conditional block checking the versions of OCP in which we are running this playbook and executing the installation steps that correspond which each version.

### DIFF
--- a/build/stf-run-ci/tasks/setup_base.yml
+++ b/build/stf-run-ci/tasks/setup_base.yml
@@ -77,7 +77,7 @@
             name: openshift-cert-manager-operator
             namespace: openshift-cert-manager-operator
           spec:
-            channel: stable-v1
+            channel: "{{ cert_manager_channel }}"
             installPlanApproval: Automatic
             name: openshift-cert-manager-operator
             source: redhat-operators
--- a/build/stf-run-ci/tasks/setup_base.yml
+++ b/build/stf-run-ci/tasks/setup_base.yml
@@ -30,31 +30,43 @@
 
 - when: not __deploy_from_index_enabled | bool
   block:
-    - name: Create openshift-cert-manager-operator namespace
-      k8s:
-        definition:
-          apiVersion: project.openshift.io/v1
-          kind: Project
-          metadata:
-            name: openshift-cert-manager-operator
-          spec:
-            finalizers:
-            - kubernetes
+    when: ocp_ver.stdout is version('4.12', '<')
+    block:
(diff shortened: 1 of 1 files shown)


vkmc:
Ran smoke tests on an environment with cert-manager from stable-v1 installed in the way that it is being installed in this patch set to verify sanity of the deployment, it is passing.

Some validation of the creation of the subscription should be added, is https://github.com/infrawatch/service-telemetry-operator/blob/master/build/validate_deployment.sh a good place for this? cc @csibbitt @leifmadsen @elfiesmelfie

csibbitt:
> Ran smoke tests on an environment with cert-manager from stable-v1 installed in the way that it is being installed in this patch set to verify sanity of the deployment, it is passing.

:+1: 

> Some validation of the creation of the subscription should be added, is https://github.com/infrawatch/service-telemetry-operator/blob/master/build/validate_deployment.sh a good place for this? cc @csibbitt @leifmadsen @elfiesmelfie

I think if the cert-manager doesn't work, then validate_deployment.sh will fail while waiting for the QDR to deploy.

csibbitt (commented):
LGTM, let's try CI on 4.10 and 4.12

csibbitt:
test

csibbitt:
There was some side discussion that we may want to pin testing to the tech-preview channel in all cases where it's available, since that's what we'd expect people to be using if they installed following the docs. If that's the case then we might want to keep this structure and just adjust the versions?

vkmc:
Considering we will stick to tech-preview for STF up to 1.5.2 we should

- Drop this change, since we still have cert-manager from tech-preview channel available in OCP 4.12
- Revert https://github.com/infrawatch/service-telemetry-operator/pull/424

For OCP 4.13 we will add dependency resolution and customers are no longer expected to pre-install cert-manager. We will need to reflect those changes in our CI. For that case, the steps followed will look like

- Create namespace
- Create operatorgroup
- Create subscription to STO
- Verify all dependencies are installed (SGO, cert-manager, AMQ Interconnect, OBO)

vkmc:
Related PR https://github.com/infrawatch/service-telemetry-operator/pull/444

Will close this PR once we agreed on the approach

PR #446, opened 2023-08-31
Remove registry access config tags

vkmc:
Remove "bundle_registry_tls_ca" and "bundle_registry_auth" tags in favor of "setup_bundle_registry_tls_ca" and
"setup_bundle_registry_auth" config options.

Values for these new options are set to true to keep backwards compatibility.


csibbitt:
I assume this simplifies the interfacing with another part of the test system?

vkmc:
yes Chris, this is part of the Zuul + CRC effort, perhaps @elfiesmelfie can provide further details

elfiesmelfie:
> yes Chris, this is part of the Zuul + CRC effort, perhaps @elfiesmelfie can provide further details

We skip certain functionality by using --skip-tags on the command line. This is fine when running locally, or using ansible-playbook as the interface to the CI, and configuring the job using args to ansible-playbook.
With zuul, we provide the playbooks to run in the CI. When including or importing the role in a platybook, we can't skip tags. By using bools instead, we can switch on and off specific functionality in a playbook for each scenario. 
For other users of stf-run-ci (e.g. some downstream testing), we can replace ``--skip-tags bundle_registry_auth`` with ``-e setup_bundle_registry_auth=false`` to get the same effect. Using the default value of true for these bools means we have the same behaviour when no tags are passed.

AFAICT, the only disadvantage that I can see is that we can't run `ansible-playbook --tags bundle_registry_auth ...` to selectively only run those tagged tasks, which is not something that we do anywhere anyway IINM.

elfiesmelfie:
I've rebased the efoley-add_zuul branch on top of this change.

The console output for that CI run is here: https://review.rdoproject.org/zuul/stream/aa9fc38b30ef49358dd3d150c0997a8f?logfile=console.log

If that change passes, then this works as expected.
The ci/deploy_stf.yml playbook sets the appropriate vars to false. If the job passes, it is because the registry setup was skipped and the behaviour is the same as using --skip-tags

elfiesmelfie (changes requested):
Please add the update to the table of role vars in README

csibbitt (approved):
Thanks for the explanation! I've always felt that tags are a bit weird anyways.

elfiesmelfie (approved):

elfiesmelfie:
test

elfiesmelfie:
I updated this branch to sync with master, since that was blocking merging.

I'll run the jenkins jobs against this once the latest tests are finished. Once that is done, we can merge this IMHO

PR #447, opened 2023-09-01
Add checks for not set sto and sgo bundle paths

vkmc:
This change adds a check for whether sto and sgo bundle paths are defined and properly inform the user if they are undefined.

Also sets the nightly builds available in Quay as defaults.


elfiesmelfie (commented):
This looks good to me, I'll rebase the efoley-add_zuul branch on top of this and see if it all passes CI. 
Once that happens, we can merge this.

elfiesmelfie:
I was seeing some issues with the gate, which have since been resolved.
I've rebased PR- 429 onto this branch, and we'll see the results soon.
While the job is running, you can follow it here:
https://review.rdoproject.org/zuul/status/change/429,a2bd6bef2dcd2619f598e16130be875e523236fc

I removed the bundle image paths from ci/deploy.yml. I expect that the job still passes, since this change takes care of that case where those values are unset.

elfiesmelfie (approved):

elfiesmelfie (approved):
Zuul passed: https://review.rdoproject.org/zuul/buildset/3d1acad5cfdf40248eb7b387039732b9

The commits in this review are included in that passing branch.

PR #449, opened 2023-09-06
Allow Grafana management in all observability modes

leifmadsen:
Allow management of Grafana in all observability modes. Grafana
continues to be installed from community operators, but can continue to
be used even when deploying STF in the default observability strategy of
use_redhat.

Closes STF-1500


leifmadsen:
Pretty sure this is the only change required to allow this to work. I am going to do a local build and test though just to verify everything.

csibbitt (approved):
Adding functionality by removing code. My favorite!

PR #429, opened 2023-06-02
DNM: [zuul] Test changes with zuul

elfiesmelfie:
This is a draft/experimenting branch. It will likely not be merged, but will work as a scratchpad for other changes.

Depends-On: https://github.com/infrawatch/smart-gateway-operator/pull/143
For index-deploy scenario:
Depends-On: https://github.com/infrawatch/service-telemetry-operator/pull/470
For local build scenario: 
Depends-On: https://github.com/infrawatch/service-telemetry-operator/pull/478

PR #452, opened 2023-09-12
Use vars for registry image container name

csibbitt:
Values are here: https://github.com/infrawatch/service-telemetry-operator/blob/627db785aa1c5746dad6a7274106511f38694b3b/build/stf-run-ci/defaults/main.yml#L37

Waiting on PR #429


build/stf-run-ci/tasks/create_catalog.yml
```diff
@@ -76,7 +76,7 @@
           dockerfile: |
             # The base image is expected to contain
             # /bin/opm (with a serve subcommand) and /bin/grpc_health_probe
-            FROM registry.redhat.io/openshift4/ose-operator-registry:v4.13
+            FROM {{default_operator_registry_image_base}}:{{default_operator_registry_image_tag}}
```
csibbitt, line 79: Despite the name "default", they never get used to create a "final" variable and are used as-is everywhere else in the code.

elfiesmelfie: Some WIP on this in: https://github.com/infrawatch/service-telemetry-operator/pull/429/commits/cf2923eb3ee25081777da231be1a28effe1acc46

leifmadsen (approved):
I'm approving this, but I'm not sure if it should land or we should wait for the zuul changes to be reviewed and landed. Approving to unblock if that makes sense.

build/stf-run-ci/tasks/create_catalog.yml
csibbitt: I'm in favor of any of these solutions in order of preference.

1. Start breaking your PR into digestible pieces so we can merge the improvements to this code independently
   * This likely has other benefits, that PR is getting very long lived and far-ranging
2. Merge this and adjust your PR accordingly
3. Abandon this and just assume we won't care to change this value until after that PR merges

I won't bother to apply any additional pressure here since I'm not fixing a problem, just noticed this and thought to be proactive.

csibbitt: Hi there. I'm back after 3 weeks and I can't find any other PR that contains this change or an equivalent/replacement/conflicting change.  The original link to https://github.com/infrawatch/service-telemetry-operator/commit/cf2923eb3ee25081777da231be1a28effe1acc46 shows this (due to force-push I assume?)

image

I'm going to remove the do-not-merge label as I think this change can safely merge.

leifmadsen: You'll get the "this commit does not believe to any branch on this repository" generally because when we merge branches, it squashes all the commits into a single commit rather than performing a merge commit. I noticed that as well in the past.

In this case though, you might be correct, because that commit refers to a parent commit in https://github.com/infrawatch/service-telemetry-operator/pull/470 so yea... I don't know where it went either :)

I agree this can likely merge without issue.

csibbitt:
recheck

csibbitt:
Two Jenkins' and a Zuul have approved this.

@elfiesmelfie do you have any objection to merging this? If you still have changes that conflict, I think they will be trivial enough to resolve.

elfiesmelfie:
> Two Jenkins' and a Zuul have approved this.
> 
> @elfiesmelfie do you have any objection to merging this? If you still have changes that conflict, I think they will be trivial enough to resolve.

I can see where I'll need to make updates, it'll conflict, but I won't be thrown off by it

elfiesmelfie (approved):

PR #453, opened 2023-09-12
Use fully qualified apiVersion on Routes

csibbitt:


csibbitt:
Correct apiVersion string according to https://docs.openshift.com/container-platform/4.10/rest_api/index.html and https://docs.openshift.com/container-platform/4.13/rest_api/index.html

csibbitt:
test

leifmadsen (approved):
Looks right :)

leifmadsen:
I did notice a few other references for `*_oauth_redir_ref` values which pass a `to_json` converted output of the OAuthRedirectReference, which I'm guess is part of the login interface for oauth-proxy. For example:

```yaml
- name: Construct oauth redirect reference                                                                                                                                                                     
  set_fact:
    alertmanager_oauth_redir_ref:
      kind: OAuthRedirectReference
      apiVersion: v1
      reference:
        kind: Route
        name: '{{ ansible_operator_meta.name }}-alertmanager-proxy'
```

Would it be worth to update these as well? I assume this is used when clicking on the Login button on the UI.

It seems to be referenced in the ServiceAccount:

```yaml
oc get serviceaccount alertmanager-stf -oyaml
apiVersion: v1
imagePullSecrets:
- name: alertmanager-stf-dockercfg-gf4xz
kind: ServiceAccount
metadata:
  annotations:
    serviceaccounts.openshift.io/oauth-redirectreference.alertmanager: '{"kind": "OAuthRedirectReference",
      "apiVersion": "v1", "reference": {"kind": "Route", "name": "default-alertmanager-proxy"}}'
```

And then references the internal Route name:

```bash
oc get routes
NAME                         HOST/PORT                                                                                               PATH   SERVICES                     PORT   TERMINATION          WILDCARD
default-alertmanager-proxy   default-alertmanager-proxy-service-telemetry.apps.stf15comm.rhos-d.infra.prod.upshift.rdu2.redhat.com          default-alertmanager-proxy   web    reencrypt/Redirect   None
default-interconnect-5671    default-interconnect-5671-service-telemetry.apps.stf15comm.rhos-d.infra.prod.upshift.rdu2.redhat.com           default-interconnect         5671   passthrough/None     None
default-prometheus-proxy     default-prometheus-proxy-service-telemetry.apps.stf15comm.rhos-d.infra.prod.upshift.rdu2.redhat.com            default-prometheus-proxy     web    reencrypt/Redirect   None
```

Since these are really just annotations I don't think we've run into the same issues as making new Route objects like we were seeing in the failed deployment. I'm not sure if these would be affected after the fact though in that environment?

csibbitt:
> I did notice a few other references for `*_oauth_redir_ref` values which pass a `to_json` converted output of the OAuthRedirectReference, which I'm guess is part of the login interface for oauth-proxy. For example:
> 
> ```yaml
> - name: Construct oauth redirect reference                                                                                                                                                                     
>   set_fact:
>     alertmanager_oauth_redir_ref:
>       kind: OAuthRedirectReference
>       apiVersion: v1
>       reference:
>         kind: Route
>         name: '{{ ansible_operator_meta.name }}-alertmanager-proxy'
> ```

I noticed these too. The "apiVersion: v1" here is not for the Route, it's for the OAuthRedirectReference. This API is not listed in the API List documentation I linked in my first comment, but I wondered if maybe it should be "oauth.openshift.io/v1". I tried looking it up further and found the docs showing it as "v1"[1] as well as lots of unofficial random examples. I could likely track down the ultimate canonical string in source code somewhere, but in any case I don't think this is related to the bug I'm currently fixing.

[1] https://access.redhat.com/documentation/en-us/openshift_container_platform/4.13/html/authentication_and_authorization/using-service-accounts-as-oauth-client#redirect-uris-for-service-accounts_using-service-accounts-as-oauth-client

leifmadsen:
> I noticed these too. The "apiVersion: v1" here is not for the Route, it's for the OAuthRedirectReference. This API is not listed in the API List documentation I linked in my first comment, but I wondered if maybe it should be "oauth.openshift.io/v1". I tried looking it up further and found the docs showing it as "v1"[1] as well as lots of unofficial random examples. I could likely track down the ultimate canonical string in source code somewhere, but in any case I don't think this is related to the bug I'm currently fixing.

Ahhhh yes now that you say that and I look at it again, it's all so obvious :)

Thanks for clarifying!

leifmadsen:
:shipit:

PR #454, opened 2023-09-12
Adjust the default scrape interval

leifmadsen:
Adjust the default scrape interval to be 15s in alignment with default
configuration changes setting the polling rate in RHOSP to 30 seconds.

Related STF-1512

### DIFF
--- a/roles/servicetelemetry/defaults/main.yml
+++ b/roles/servicetelemetry/defaults/main.yml
@@ -40,7 +40,7 @@ servicetelemetry_defaults:
       prometheus:
         enabled: true
         deployment_size: 1
-        scrape_interval: 15s
+        scrape_interval: 30s
         storage:
           strategy: persistent
           retention: 24h
--- a/roles/servicetelemetry/defaults/main.yml
+++ b/roles/servicetelemetry/defaults/main.yml
@@ -40,7 +40,7 @@ servicetelemetry_defaults:
       prometheus:
         enabled: true
         deployment_size: 1
-        scrape_interval: 10s
+        scrape_interval: 15s
         storage:
           strategy: persistent
           retention: 24h
--- a/tests/infrared/17.1/enable-stf.yaml.template
+++ b/tests/infrared/17.1/enable-stf.yaml.template
@@ -6,16 +6,14 @@ custom_templates:
     # matches the documentation for enable-stf.yaml in stable-1.3 documentation
     parameter_defaults:
         # only send to STF, not other publishers
-        EventPipelinePublishers: []
         PipelinePublishers: []
 
         # manage the polling and pipeline configuration files for Ceilometer agents
         ManagePolling: true
         ManagePipeline: true
 
-        # enable Ceilometer metrics and events
+        # enable Ceilometer metrics
         CeilometerQdrPublishMetrics: true
-        CeilometerQdrPublishEvents: true
 
         # enable collection of API status
         CollectdEnableSensubility: true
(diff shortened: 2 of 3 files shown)


tests/infrared/17.1/enable-stf.yaml.template
```diff
@@ -78,4 +76,4 @@ custom_templates:
               local:
                 host: "%{hiera('fqdn_canonical')}"
                 port: 11211
-
+# vim: ft=yaml
```
vkmc, line 79: NIT drop this

leifmadsen:
Moving this to ready for review since I've done a full deployment now, and everything seems to have updated to what it should be. Intervals on collectd are 30 seconds, the ceilometer configuration is 30 seconds, and then pipeline.yaml only includes the metering publisher.

Dashboards mostly working. I need to check that sensubility not working is a known issue. I believe there were items open still in 17.1 GA where sensubility is not currently fully operational.

vkmc (changes requested):
Looks good, just a minor nit. Thanks Leif!

tests/infrared/17.1/enable-stf.yaml.template
leifmadsen, line 79:
```suggestion
```

leifmadsen:
```yaml
c get servicemonitors.monitoring.rhobs default-cloud1-coll-meter -oyaml
apiVersion: monitoring.rhobs/v1
kind: ServiceMonitor
metadata:
  creationTimestamp: "2023-09-21T03:23:42Z"
  generation: 1
  labels:
    app: smart-gateway
  name: default-cloud1-coll-meter
  namespace: service-telemetry
  ownerReferences:
  - apiVersion: infra.watch/v1beta1
    kind: ServiceTelemetry
    name: default
    uid: c0cf35a3-eee6-4afc-81c3-9b0f450ae90e
  resourceVersion: "60991874"
  uid: d70d3d60-820b-4030-8e18-fdbe6a799a92
spec:
  endpoints:
  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
    interval: 15s
[...]
```

```yaml
[root@controller-0 ceilometer]# cat polling.yaml 
---
sources:
    - name: some_pollsters
      interval: 30
[...]
```

```xml
[root@controller-0 collectd.d]# cat 10-amqp1.conf 
# Generated by Puppet
<LoadPlugin amqp1>
  Globals false
  Interval 30
</LoadPlugin>

<Plugin amqp1>
[...]
```

```xml
[root@controller-0 etc]# cat collectd.conf 
# Generated by Puppet

#Hostname localhost
FQDNLookup true

AutoLoadPlugin false
#BaseDir "/var/lib/collectd"
#PluginDir "/usr/lib/collectd"
TypesDB "/usr/share/collectd/types.db" "/usr/share/collectd/types.db.libpodstats"
Interval 30
Timeout 2
ReadThreads 5
WriteThreads 5
Include "/etc/collectd.d/*.conf"
```

leifmadsen:
Oops, while going through and getting example configuration files, I realized I forgot to remove the events coordination from stf-connectors.yaml.template which resulted in the event_pipeline.conf being created for Ceilometer. Updated the configuration to match the landed documentation.

leifmadsen:
Going to re-run my deployment just to really get this sorted. See you in a couple of hours :)

leifmadsen:
Starting local test

roles/servicetelemetry/defaults/main.yml
```diff
@@ -40,7 +40,7 @@ servicetelemetry_defaults:
       prometheus:
         enabled: true
         deployment_size: 1
-        scrape_interval: 10s
+        scrape_interval: 15s
```
csibbitt, line 43: I wonder if this should really just be 30s.

I found a record of myself saying "I have no problem with those values" 9 days ago, and also the comment about "in _theory_ I'd rather sample twice as fast as a source, not half as fast", which is true, but was a bit of a joke meant to emphasize the very old SG behavior (and origin of the 5s/10s values in the code) where we were forced to sample half as fast because polling too fast would cause errors. We're not doing analog sampling, so Nyquist's Theorem doesn't apply here.  I think setting it to 30s to match everything else might be the simpler and better option.

leifmadsen, line 43:
```suggestion
        scrape_interval: 30s
```

leifmadsen: Done!

vkmc (approved):
LGTM

leifmadsen:
> Starting local test

For whatever reason the helper scripts which were working fine are no longer resulting in a deployment of RHOSP 17.1. Seems to be related to some image sync issue. Will try again today and hope it was just an internal infrastructure issue last week that has since been resolved 🤷🏻

PR #458, opened 2023-09-13
Update Dockerfile to create epel repo config when NO_PROXY is undefined

elfiesmelfie:


leifmadsen:
I saw this change in the other PR, but my eyes can't spot what the change actually is here other than a newline at the end? Also a small typo fix in the comment?

elfiesmelfie:
> I saw this change in the other PR, but my eyes can't spot what the change actually is here other than a newline at the end? Also a small typo fix in the comment?

```
- "${NO_PROXY:__ZZZZZ}" == "__ZZZZZ"
+ "${NO_PROXY:-__ZZZZZ}" == "__ZZZZZ"
```

csibbitt (approved):

csibbitt:
> ```
> - "${NO_PROXY:__ZZZZZ}" == "__ZZZZZ"
> + "${NO_PROXY:-__ZZZZZ}" == "__ZZZZZ"
> ```

The original is `${NO_PROXY-__ZZZZZ}`, it's the `:` that is missing, not the `-`

csibbitt:
I'm submitting this as an interesting curiosity, then I'm forcing the merge since we're having CI issues and this is trivial.

```
$ echo "${NO_PROXY:-__ZZZZZ}"
__ZZZZZ

$ echo "${NO_PROXY-__ZZZZZ}"
__ZZZZZ

$ NO_PROXY=true

$ echo "${NO_PROXY:-__ZZZZZ}"
true

$ echo "${NO_PROXY-__ZZZZZ}"
true
```

I don't see anything in the bash manual[1] that suggests the `:` is optional, but okay. I still like this version better.

https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

PR #461, opened 2023-09-14
[stf-run-ci] Update pre-flight checks to be time-limited

elfiesmelfie:
If the CSV doesn't succeed, the task continues indefinitely. This will cause an infinite loop in CI if there is no timeout on the job.

Adding in a 10 minute timeout allows the CI job to return a failure instead of running indefinitely.


csibbitt (approved):

vkmc (approved):
Looks good, thanks!

PR #465, opened 2023-09-14
[issue#306] Add missing ClusterRoles

### LINKED ISSUE #306: STF won't install without cluster-monitoring-operator installed (workaround within)
Not usually a problem since it's installed by default, but not on CRC.

```
TASK [Bind the local prometheus SA to prometheus cluster role] ******************************** 
fatal: [localhost]: FAILED! => {"changed": false, "error": 404, "msg": "Failed to create object: b'{\"kind\":\"Status\",\"apiVersion\":\"v1\",\"metadata\":{},\"status\":\"Failure\",\"message\":\"clusterroles.rbac.authorization.k8s.io \\\\\"prometheus-k8s\\\\\" not found\",\"reason\":\"NotFound\",\"details\":{\"name\":\"prometheus-k8s\",\"group\":\"rbac.authorization.k8s.io\",\"kind\":\"clusterroles\"},\"code\":404}\\n'", "reason": "Not Found", "status": 404}
```
needs https://paste.opendev.org/show/812354/

```
 TASK [Bind role] ******************************** 
fatal: [localhost]: FAILED! => {"changed": false, "error": 404, "msg": "Failed to create object: b'{\"kind\":\"Status\",\"apiVersion\":\"v1\",\"metadata\":{},\"status\":\"Failure\",\"message\":\"clusterroles.rbac.authorization.k8s.io \\\\\"alertmanager-main\\\\\" not found\",\"reason\":\"NotFound\",\"details\":{\"name\":\"alertmanager-main\",\"group\":\"rbac.authorization.k8s.io\",\"kind\":\"clusterroles\"},\"code\":404}\\n'", "reason": "Not Found", "status": 404}
```
needs https://paste.opendev.org/show/812355/

elfiesmelfie:
The cluster-monitoring-operator is required for STF to install. It creates the required alertmanager-main and prometheus-k8s. ClusterRoles, and STF relies on these being present. These are not present when using CRC, so ClusterRoles need to be explicitly created.

The names of the ClusterRoles have been updated, in case there is some conflict when cluster-monitoring-operator is installed after STF.

This is a workaround for not having cluster-monitoring-operator installed: https://github.com/infrawatch/service-telemetry-operator/issues/306

resolves #306

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -22,6 +22,7 @@ rules:
   - watch
   - update
   - patch
+  - delete
 - apiGroups:
   - authorization.k8s.io
   resources:
--- a/roles/servicetelemetry/tasks/component_alertmanager.yml
+++ b/roles/servicetelemetry/tasks/component_alertmanager.yml
@@ -66,7 +66,7 @@
         kind: Route
         name: '{{ ansible_operator_meta.name }}-alertmanager-proxy'
 
-- name: Add a service account to used by Alertmanager
+- name: Create ServiceAccount/alertmanager-stf with oauth redirect annotation
   k8s:
     definition:
       apiVersion: v1
@@ -97,6 +97,53 @@
           - subjectaccessreviews
           verbs:
           - create
+
+- name: Setup ClusterRoleBinding for Alertmanager
+  block:
+    - name: Define ClusterRoleBinding/alertmanager-stf
+      set_fact:
+        def_alertmanager_stf_crb: |
+          apiVersion: rbac.authorization.k8s.io/v1
+          kind: ClusterRoleBinding
+          metadata:
+            name: alertmanager-stf
+          roleRef:
+            apiGroup: rbac.authorization.k8s.io
+            kind: ClusterRole
+            name: alertmanager-stf
+          subjects:
+          - kind: ServiceAccount
+            name: alertmanager-stf
+            namespace: '{{ ansible_operator_meta.namespace }}'
+
+    - name: Create ClusterRoleBinding/alertmanager-stf
+      k8s:
+        definition:
+          "{{ def_alertmanager_stf_crb }}"
+  rescue:
+    - name: Remove ClusterRoleBinding/alertmanager-stf when fail to update
+      k8s:
+        state: absent
+        definition:
+          apiVersion: rbac.authorization.k8s.io/v1
+          kind: ClusterRoleBinding
+          metadata:
+            name: alertmanager-stf
+
+    - name: Create ClusterRoleBinding/alertmanager-stf
+      k8s:
+        definition:
+          "{{ def_alertmanager_stf_crb }}"
+
+- name: Create Role/alertmanager-stf
+  k8s:
+    definition:
+      apiVersion: rbac.authorization.k8s.io/v1
+      kind: Role
+      metadata:
+        name: alertmanager-stf
+        namespace: '{{ ansible_operator_meta.namespace }}'
+      rules:
         - apiGroups:
           - security.openshift.io
           resourceNames:
@@ -106,23 +153,45 @@
           verbs:
           - use
 
-- name: Bind role
-  k8s:
-    definition:
-      apiVersion: rbac.authorization.k8s.io/v1
-      kind: RoleBinding
-      metadata:
-        name: alertmanager-stf
-        namespace: '{{ ansible_operator_meta.namespace }}'
-      roleRef:
-        apiGroup: rbac.authorization.k8s.io
-        kind: ClusterRole
-        name: alertmanager-stf
-        namespace: '{{ ansible_operator_meta.namespace }}'
-      subjects:
-      - kind: ServiceAccount
-        name: alertmanager-stf
-        namespace: '{{ ansible_operator_meta.namespace }}'
+- name: Setup RoleBinding for Alertmanager
+  block:
+    - name: Define RoleBinding/alertmanager-stf
+      set_fact:
+        def_alertmanager_stf_rb: |
+          apiVersion: rbac.authorization.k8s.io/v1
+          kind: RoleBinding
+          metadata:
+            name: alertmanager-stf
+            namespace: '{{ ansible_operator_meta.namespace }}'
+          roleRef:
+            apiGroup: rbac.authorization.k8s.io
+            kind: Role
+            name: alertmanager-stf
+            namespace: '{{ ansible_operator_meta.namespace }}'
+          subjects:
+          - kind: ServiceAccount
+            name: alertmanager-stf
+            namespace: '{{ ansible_operator_meta.namespace }}'
+
+    - name: Create RoleBinding/alertmanager-stf
+      k8s:
+        definition:
+          "{{ def_alertmanager_stf_rb }}"
+  rescue:
+    - name: Remove RoleBinding/alertmanager-stf when fail to update
+      k8s:
+        state: absent
+        definition:
+          apiVersion: rbac.authorization.k8s.io/v1
+          kind: RoleBinding
+          metadata:
+            name: alertmanager-stf
+            namespace: '{{ ansible_operator_meta.namespace }}'
+
+    - name: Create RoleBinding/alertmanager-stf
+      k8s:
+        definition:
+          "{{ def_alertmanager_stf_rb }}"
 
 - name: Set default alertmanager service template
   set_fact:
--- a/roles/servicetelemetry/tasks/component_alertmanager.yml
+++ b/roles/servicetelemetry/tasks/component_alertmanager.yml
@@ -77,14 +77,13 @@
         annotations:
           serviceaccounts.openshift.io/oauth-redirectreference.alertmanager: '{{ alertmanager_oauth_redir_ref | to_json }}'
 
-- name: Create Role/alertmanager-stf
+- name: Create ClusterRole/alertmanager-stf
   k8s:
     definition:
       apiVersion: rbac.authorization.k8s.io/v1
-      kind: Role
+      kind: ClusterRole
       metadata:
         name: alertmanager-stf
-        namespace: '{{ ansible_operator_meta.namespace }}'
       rules:
         - apiGroups:
           - authentication.k8s.io
@@ -117,7 +116,7 @@
         namespace: '{{ ansible_operator_meta.namespace }}'
       roleRef:
         apiGroup: rbac.authorization.k8s.io
-        kind: Role
+        kind: ClusterRole
         name: alertmanager-stf
         namespace: '{{ ansible_operator_meta.namespace }}'
       subjects:
--- a/roles/servicetelemetry/tasks/component_alertmanager.yml
+++ b/roles/servicetelemetry/tasks/component_alertmanager.yml
@@ -77,6 +77,35 @@
         annotations:
           serviceaccounts.openshift.io/oauth-redirectreference.alertmanager: '{{ alertmanager_oauth_redir_ref | to_json }}'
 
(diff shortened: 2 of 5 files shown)


leifmadsen (approved):

leifmadsen:
test

elfiesmelfie:
The Jenkins job timed out :disappointed:

leifmadsen:
> The Jenkins job timed out 😞

There is a bug. It times out on the validation because all components are not operational. Here is the first error I found in the STO logs:

```
 TASK [Bind the local prometheus SA to prometheus cluster role (for oauth perms)] ******************************** 
fatal: [localhost]: FAILED! => {"changed": false, "error": 422, "msg": "Failed to patch object: b'{\"kind\":\"Status\",\"apiVersion\":\"v1\",\"metadata\":{},\"status\":\"Failure\",\"message\":\"ClusterRoleBinding.rbac.authorization.k8s.io \\\\\"prometheus-k8s-service-telemetry\\\\\" is invalid: roleRef: Invalid value: rbac.RoleRef{APIGroup:\\\\\"rbac.authorization.k8s.io\\\\\", Kind:\\\\\"ClusterRole\\\\\", Name:\\\\\"prometheus-stf\\\\\"}: cannot change roleRef\",\"reason\":\"Invalid\",\"details\":{\"name\":\"prometheus-k8s-service-telemetry\",\"group\":\"rbac.authorization.k8s.io\",\"kind\":\"ClusterRoleBinding\",\"causes\":[{\"reason\":\"FieldValueInvalid\",\"message\":\"Invalid value: rbac.RoleRef{APIGroup:\\\\\"rbac.authorization.k8s.io\\\\\", Kind:\\\\\"ClusterRole\\\\\", Name:\\\\\"prometheus-stf\\\\\"}: cannot change roleRef\",\"field\":\"roleRef\"}]},\"code\":422}\\n'", "reason": "Unprocessable Entity", "status": 422}
```

roles/servicetelemetry/tasks/component_alertmanager.yml
```diff
@@ -88,7 +117,7 @@
       roleRef:
         apiGroup: rbac.authorization.k8s.io
         kind: ClusterRole
-        name: alertmanager-main
+        name: alertmanager-stf
```
leifmadsen, line 120: Kubernetes won't let you change an existing roleRef in a ClusterRoleBinding, so we'll actually need to remove this ClusterRoleBinding and re-create it with a new roleRef. I'm going to implement a potential fix and test it locally.

roles/servicetelemetry/tasks/component_prometheus.yml
```diff
@@ -87,7 +132,7 @@
       roleRef:
         apiGroup: rbac.authorization.k8s.io
         kind: ClusterRole
-        name: prometheus-k8s
+        name: prometheus-stf
```
leifmadsen, line 135: Same issue as above where roleRef's cannot be changed after the fact.

This isn't found in upstream CI on CRC because the clusters are (presumably) created fresh every time, but our existing Jenkins-based deployments finds it since I guess we're not (properly?) cleaning out all the ClusterRoleBindings before doing a fresh deployment?

leifmadsen: Fixed via #467

roles/servicetelemetry/tasks/component_prometheus.yml
```diff
@@ -7,90 +7,133 @@
...
+        - watch
+      - apiGroups:
+        - extensions
+        - networking.k8s.io
+        resources:
+        - ingresses
+        verbs:
+        - get
+        - list
+        - watch
```
leifmadsen, line 110: The `prometheus-k8s` role doesn't include these, so not sure if we're using this because we permit UI access via oauth-proxy perhaps?

CC: @csibbitt

deploy/role.yaml
```diff
@@ -22,6 +22,7 @@ rules:
   - watch
   - update
   - patch
+  - delete
```
csibbitt, line 25: Noting that this should be transitional only and could be removed in a later version.

roles/servicetelemetry/tasks/component_prometheus.yml
```diff
@@ -7,90 +7,133 @@
...
+        resources:
+        - securitycontextconstraints
+        verbs:
+        - use
+
+- name: Create ClusterRoleBinding/prometheus-stf
   k8s:
     definition:
       apiVersion: rbac.authorization.k8s.io/v1
       kind: ClusterRoleBinding
```
csibbitt, line 71: I believe this should change to a "RoleBinding" if possible.

From https://kubernetes.io/docs/reference/access-authn-authz/rbac/#kubectl-create-rolebinding

`kubectl create rolebinding` Grants a Role or ClusterRole within a specific namespace.
`kubectl create clusterrolebinding` Grants a ClusterRole across the entire cluster (all namespaces).

roles/servicetelemetry/tasks/component_alertmanager.yml
```diff
@@ -77,18 +77,49 @@
         annotations:
           serviceaccounts.openshift.io/oauth-redirectreference.alertmanager: '{{ alertmanager_oauth_redir_ref | to_json }}'
 
+- name: Create Role/alertmanager-stf
+  k8s:
+    definition:
+      apiVersion: rbac.authorization.k8s.io/v1
+      kind: Role
```
csibbitt, line 84: :+1:

roles/servicetelemetry/tasks/component_alertmanager.yml
```diff
@@ -77,18 +77,49 @@
...
+          - securitycontextconstraints
+          verbs:
+          - use
+
 - name: Bind role
   k8s:
     definition:
       apiVersion: rbac.authorization.k8s.io/v1
-      kind: ClusterRoleBinding
+      kind: RoleBinding
```
csibbitt, line 114: :+1:

csibbitt (commented):
Since we're mucking about in here, I'd like to take the opportunity to try to narrow this scope as much as possible. It looks pretty good right now, but I think there are still a few things than can move from the cluster scope to the namespaced scope.

roles/servicetelemetry/tasks/component_prometheus.yml
```diff
@@ -7,90 +7,133 @@
...
+      metadata:
+        name: prometheus-stf
+      rules:
+      - apiGroups:
+        - ""
+        resources:
+        - nodes/metrics
+        verbs:
+        - get
+      - nonResourceURLs:
```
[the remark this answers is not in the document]
leifmadsen: I tried moving these things to the Role, but every time I removed something here, Prometheus wouldn't have access to scrap the sg-core endpoints. It's not really clear to me why that is the case, because I started with a minimal file, and slowly added things until they worked again. Unfortunately it seems this was necessary to get sg-core to not show up with a 403 in the Prometheus Targets screen.

roles/servicetelemetry/tasks/component_prometheus.yml
leifmadsen: Ahhhh ok, let me try that as well. I didn't test using a RoleBinding against a ClusterRole.

roles/servicetelemetry/tasks/component_prometheus.yml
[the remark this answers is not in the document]
csibbitt: Okay, if that's true then I'm satisfied. Maybe it's because the subjectaccessreviews in question (for NonResourceURLs) and tokenreviews (in general) are inherently non-namespaced operations. I don't have the foggiest idea why the securitycontextconstraints/use permission would need to be in a ClusterRole though. I have some additional ideas here, but they are not for this PR.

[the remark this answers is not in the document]
leifmadsen: Also I think this is still a ClusterRole, because previously this would have been the prometheus-k8s ClusterRole rather than our own. For reference here is the `prometheus-k8s` ClusterRole:

```yaml
oc get clusterrole prometheus-k8s -oyaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  creationTimestamp: "2023-06-13T14:51:40Z"
  labels:
    app.kubernetes.io/component: prometheus
    app.kubernetes.io/instance: k8s
    app.kubernetes.io/name: prometheus
    app.kubernetes.io/part-of: openshift-monitoring
    app.kubernetes.io/version: 2.39.1
  name: prometheus-k8s
  resourceVersion: "24913"
  uid: 0dc0c1ba-af3b-4824-a9e7-c03523a074b2
rules:
- apiGroups:
  - ""
  resources:
  - nodes/metrics
  verbs:
  - get
- nonResourceURLs:
  - /metrics
  verbs:
  - get
- apiGroups:
  - authentication.k8s.io
  resources:
  - tokenreviews
  verbs:
  - create
- apiGroups:
  - authorization.k8s.io
  resources:
  - subjectaccessreviews
  verbs:
  - create
- apiGroups:
  - ""
  resources:
  - namespaces
  verbs:
  - get
- apiGroups:
  - security.openshift.io
  resourceNames:
  - nonroot
  resources:
  - securitycontextconstraints
  verbs:
  - use
```

roles/servicetelemetry/tasks/component_prometheus.yml
leifmadsen: Other prometheus-k8s objects are using this same Role configuration across the OCP cluster. I'm going to leave it for now until we can do further investigation about what CRDs etc are using this interface. If we're truly not using it and it won't cause issues, then I'll be more comfortable removing it at that point.

roles/servicetelemetry/tasks/component_prometheus.yml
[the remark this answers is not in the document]
leifmadsen: From what I've read and tested, it seems SCC _can_ move but that TokenReviews and SARs are cluster bound and can't be moved to a namespace. I could definitely be wrong, but at the very least it seems Prometheus needs them at the Cluster scope to allow Targets to work.

roles/servicetelemetry/tasks/component_prometheus.yml
leifmadsen: Tested. Unfortunately it fails and this will need to remain a ClusterRole.

leifmadsen:
Requesting final review prior to merge!

leifmadsen:
CI timed out again. I think the Jenkins-based CI systems are just overwhelmed.

deploy/role.yaml
leifmadsen: I was going to file an issue about this, but then I realized that if someone was somehow running STF 1.5.2 and moved to STF 1.5.4 they would need this.

Additionally, the latest changes I landed in https://github.com/infrawatch/service-telemetry-operator/pull/475/commits/5d0210d1416681a102cba3cc097b18e74b22b927 will continually require the delete RBAC.

We _could_ refine this though and break out the `clusterroles` and `clusterrolebindings` though to have the `delete` resource only on the `clusterrolebindings`. I'm also curious if we could further lock this down to just resources we manage and not _all_ ClusterRoles and ClusterRoleBindings.

leifmadsen:
I'm skipping out on approving this since I landed a fair number of changes here, and it'd effectively be self-approving.

Deferring to Chris and Victoria on this one, but I think we've all probably looked at this enough that if CI passes, landing it isn't going to be a problem for anyone. Thanks everyone for the patience and quick turn around on the reviews!

csibbitt (approved):

leifmadsen:
LET'S GO!!!!!!!

PR #467, opened 2023-09-15
Fix up the RBAC setup for prometheus-stf

leifmadsen:
Fix up the RBAC changes to fully get prometheus-stf working and
decoupled from prometheus-k8s. Changes to using a separate
prometheus-stf ClusterRole, ClusterRoleBinding, and ServiceAccount,
along with a Role and RoleBinding, all using prometheus-stf as the
ServiceAccount. Also updates the Alertmanager configuration to use
alertmanager-stf instead of alertmanager-main.

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -22,6 +22,7 @@ rules:
   - watch
   - update
   - patch
+  - delete
 - apiGroups:
   - authorization.k8s.io
   resources:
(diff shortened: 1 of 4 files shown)


leifmadsen:
I did some local testing too verifying alertmanager, alert delivery etc. I was able to load the existing alerts.yml from STO, waiting 10 mins for a pending alert to fire, and verified I could see it in alertmanager UI (which also verified I could login to it via oauth-proxy) and that alerts were delivered to prometheus-webhook-snmp receiver. I check the logs of prometheus-webhook-snmp receiver and it saw the alerts.

PR #468, opened 2023-09-15
Refactor smoketest script

leifmadsen:
Perform a bit of smoketest refactoring and fix up a few bugs.

* Update alert trigger to use startsAt in order to potentially speed up
  delivery of the alerts. Failures in the SNMP_WEBHOOK_STATUS seems to
  be primarily to delayed alert notification through
  prometheus-snmp-webhook.
* Add an alert clean up task as part of the clean up logic at the end.
* Update openssl x509 to not use the -in flag which seems unnecessary
  and on some systems causes a failure.
* Add new SMOKETEST_VERBOSE boolean so local testing can skip massive
  amounts of information dumped to stdout.
* Remove curl pod using label selector for slightly cleaner output.
* Update failure check to combine RET and SNMP_WEBHOOK_STATUS since
  testing seems to show changes are slightly more reliable.

### DIFF
--- a/tests/smoketest/smoketest.sh
+++ b/tests/smoketest/smoketest.sh
@@ -80,6 +80,7 @@ fi
 # create the alert using startsAt which in theory may cause trigger to be faster
 echo "*** [INFO] Create alert"
 oc delete pod -l run=curl ; oc run curl --wait --restart='Never' --image=quay.io/infrawatch/busyboxplus:curl -- sh -c "curl -v -k -H \"Content-Type: application/json\" -H \"Authorization: Bearer ${PROMETHEUS_K8S_TOKEN}\" -d '[{\"status\":\"firing\",\"labels\":{\"alertname\":\"smoketest\",\"severity\":\"warning\"},\"startsAt\":\"$(date --rfc-3339=seconds | sed 's/ /T/')\"}]' https://default-alertmanager-proxy:9095/api/v1/alerts"
+oc logs curl
 
 echo "*** [INFO] Waiting to see SNMP trap message in webhook pod"
 SNMP_WEBHOOK_POD=$(oc get pod -l "app=default-snmp-webhook" -ojsonpath='{.items[0].metadata.name}')
--- a/tests/smoketest/smoketest.sh
+++ b/tests/smoketest/smoketest.sh
@@ -28,6 +28,7 @@ if [ "${OC_CLIENT_VERSION_Y}" -lt "${OC_CLIENT_VERSION_Y_REQUIRED}" ] || [ "${OC
 fi
 
 CLEANUP=${CLEANUP:-true}
+SMOKETEST_VERBOSE=${SMOKETEST_VERBOSE:-true}
 
 for ((i=1; i<=NUMCLOUDS; i++)); do
   NAME="smoke${i}"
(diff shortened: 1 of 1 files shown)


tests/smoketest/smoketest.sh
```diff
@@ -112,70 +96,85 @@ while [ $SNMP_WEBHOOK_CHECK_COUNT -lt $SNMP_WEBHOOK_CHECK_MAX_TRIES ]; do
...
-
-echo "*** [INFO] Logs from smart gateway operator..."
-oc logs "$(oc get pod -l app=smart-gateway-operator -o jsonpath='{.items[0].metadata.name}')"
-echo
-
-echo "*** [INFO] Logs from prometheus..."
-oc logs "$(oc get pod -l prometheus=default -o jsonpath='{.items[0].metadata.name}')" -c prometheus
-echo
+echo "*** [INFO] Checking that the qdr certificate has a long expiry"
+EXPIRETIME=$(oc get secret default-interconnect-openstack-ca -o json | grep \"tls.crt\"\: | awk -F '": "' '{print $2}' | rev | cut -c3- | rev | base64 -d | openssl x509 -text | grep "Not After" | awk -F " : " '{print $2}')
```
leifmadsen, line 108: My machine fails with the following error when using the `-in -` flag with openssl. Found a post from 4 years ago that says it is unnecessary? Guess we'll find out if maybe it's a difference between older containers and my more modern desktop.

```
Could not open file or uri for loading certificate from -
002EB2F38A7F0000:error:16000069:STORE routines:ossl_store_get0_loader_int:unregistered scheme:crypto/store/store_register.c:237:scheme=file
002EB2F38A7F0000:error:80000002:system library:file_open:No such file or directory:providers/implementations/storemgmt/file_store.c:267:calling stat(-)
Unable to load certificate
```

tests/smoketest/smoketest.sh
```diff
@@ -112,70 +96,85 @@ while [ $SNMP_WEBHOOK_CHECK_COUNT -lt $SNMP_WEBHOOK_CHECK_MAX_TRIES ]; do
...
+EXPIRETIME_UNIX=$(date -d "${EXPIRETIME}" "+%s")
+TARGET_UNIX=$(date -d "now + 7 years" "+%s")
+if [ ${EXPIRETIME_UNIX} -lt ${TARGET_UNIX} ]; then
+    echo "[FAILURE] Certificate expire time (${EXPIRETIME}) less than 7 years from now"
```
leifmadsen, line 112: Should this actually cause a failure in our testing?

tests/smoketest/smoketest.sh
```diff
@@ -112,70 +96,85 @@ while [ $SNMP_WEBHOOK_CHECK_COUNT -lt $SNMP_WEBHOOK_CHECK_MAX_TRIES ]; do
...
-echo "*** [INFO] Logs from elasticsearch..."
-oc logs "$(oc get pod -l common.k8s.elastic.co/type=elasticsearch -o jsonpath='{.items[0].metadata.name}')"
+echo "*** [INFO] Showing oc get all..."
+oc get all
 echo
 
-echo "*** [INFO] Logs from snmp webhook..."
-oc logs "$(oc get pod -l app=default-snmp-webhook -o jsonpath='{.items[0].metadata.name}')"
+echo "*** [INFO] Showing servicemonitors..."
+oc get servicemonitors.monitoring.rhobs -o yaml
```
leifmadsen, line 120: Need to also remember to go through docs and make sure we call things out in expanded form because the defaults will look everything up via *.monitoring.coreos.com.

tests/smoketest/smoketest.sh
elfiesmelfie: I've seen this fail if there's sufficient time between deploying STF and running the smoketests.

I don't think that kind of test is really useful, especially if you rotate your certificates more often.
Maybe a test to make sure the certificate is valid and the value is not the default? i.e. confirm that we set the right vars to actually update the expiry time.

leifmadsen:
Actually I thought both OCP 4.12 and 4.10 failed, but it seems that the OCP 4.10 CI system failed again. I've done local testing and gotten multiple passes to happen, and it looks like OCP 4.12 has also passed. However I see this at the very end:

```bash
*** [FAILURE] Smoke test job still not succeeded after 300s
```

Then I can see in the logs output for the snmp-webhook receiver that it never got an alert:

```bash
*** [INFO] Logs from snmp webhook...
DEBUG:prometheus_webhook_snmp.utils:Configuration settings: {"debug": true, "snmp_host": "192.168.24.254", "snmp_port": 162, "snmp_community": "public", "snmp_retries": 5, "snmp_timeout": 1, "alert_oid_label": "oid", "trap_oid_prefix": "1.3.6.1.4.1.50495.15", "trap_default_oid": "1.3.6.1.4.1.50495.15.1.2.1", "trap_default_severity": "", "host": "0.0.0.0", "port": 9099, "metrics": false, "cert": "", "key": ""}
INFO:cherrypy.error:[15/Sep/2023:20:55:15] ENGINE Listening for SIGTERM.
INFO:cherrypy.error:[15/Sep/2023:20:55:15] ENGINE Listening for SIGHUP.
INFO:cherrypy.error:[15/Sep/2023:20:55:15] ENGINE Listening for SIGUSR1.
INFO:cherrypy.error:[15/Sep/2023:20:55:15] ENGINE Bus STARTING
INFO:cherrypy.error:[15/Sep/2023:20:55:15] ENGINE Serving on http://0.0.0.0:9099
INFO:cherrypy.error:[15/Sep/2023:20:55:15] ENGINE Bus STARTED
```

And I see near the top that the `curl` pod actually errored out (not sure why...):

```bash
*** [INFO] Creating smoketest jobs...
No resources found
job.batch/stf-smoketest-smoke1 created
*** [INFO] Triggering an alertmanager notification...
*** [INFO] Create alert
No resources found
pod/curl created
*** [INFO] Waiting to see SNMP trap message in webhook pod
*** [INFO] Waiting on job/stf-smoketest-smoke1...
job.batch/stf-smoketest-smoke1 condition met
*** [INFO] Checking that the qdr certificate has a long expiry
*** [INFO] Showing oc get all...
NAME                                                            READY   STATUS      RESTARTS   AGE
pod/alertmanager-default-0                                      3/3     Running     0          3m17s
pod/cert-manager-operator-controller-manager-7fccfddcbb-bnrt4   2/2     Running     0          14m
pod/curl                                                        0/1     Error       0          2m31s
```

Not sure why the smoketest stage actually passed then. Seems like I must have broken something there.

tests/smoketest/smoketest.sh
```diff
@@ -112,70 +96,85 @@ while [ $SNMP_WEBHOOK_CHECK_COUNT -lt $SNMP_WEBHOOK_CHECK_MAX_TRIES ]; do
...
-if [ $SNMP_WEBHOOK_STATUS -ne 0 ]; then
-    echo "*** [FAILURE] SNMP Webhook failed"
-    exit 1
-fi
-
-if [ $RET -eq 0 ]; then
+if [ $RET -eq 0 ] && [ $SNMP_WEBHOOK_STATUS -eq 0 ]; then
     echo "*** [SUCCESS] Smoke test job completed successfully"
 else
     echo "*** [FAILURE] Smoke test job still not succeeded after ${JOB_TIMEOUT}"
```
leifmadsen, line 180: Oh I see the issue here now. I should actually have added the `exit 1` here because the script will never fail if SNMP_WEBHOOK_STATUS fails now due to my combination of the checks here. Since the $RET value was 0 it passed the job.

Actually I kind of discussed in chat about maybe having the snmp-webhook test be a bit of an optional pass, in which case, maybe this is working as designed now? :)

tests/smoketest/smoketest.sh
csibbitt: Yeah, this is a regression test; but we don't have a regression test suite anywhere, so it ended up in the smoke test

csibbitt (approved):

leifmadsen:
OK I'm merging this down. It's landing in issue/306 so we still have an opportunity to adjust if we want, but I'm looking to get everything merged down. Passes 1 of 2 CI systems, so good enough for merging to not-master :)

tests/smoketest/smoketest.sh
vkmc: I would vote for keep SNMP as experimental (meaning, don't fail the CI is SNMP tests don't pass) if it doesn't behave in a reliable way

leifmadsen: I've run into some issues where the SNMP check is actually catching a real issue :) I'm going to end up putting it back in as a required test to pass.

PR #463, opened 2023-09-14
generate_bundle: add logging

elfiesmelfie:
* use absolute dir for REL
* specify default value for OPERATOR_SDK
* pass a logfile
* use -x for more debug info

PR #471, opened 2023-09-18
Improve logging

elfiesmelfie:
Add in better logs to aid in debugging on the CI and when running locally.

### DIFF
--- a/build/stf-run-ci/tasks/setup_stf_local_build.yml
+++ b/build/stf-run-ci/tasks/setup_stf_local_build.yml
@@ -20,10 +20,11 @@
         ./generate_bundle.sh
     register: generate_bundle_sgo
   always:
+    # "|| true" is needed until https://github.com/infrawatch/smart-gateway-operator/pull/143 is merged
   - name: Show generate bundle log
     ansible.builtin.shell:
       cmd: |
-        cat {{ logfile_dir }}/sgo_gen_bundle.log
+        cat {{ logfile_dir }}/sgo_gen_bundle.log || true
 
 - name: Replace namespace in SGO role binding
   ansible.builtin.replace:


build/stf-run-ci/tasks/create_catalog.yml
```diff
@@ -5,10 +5,12 @@
     state: directory
     mode: '0755'
 
+# Updating to use stdout_lines[-1] so that any additional info that gets added to generate_bundles (e.g. for debug) doesn't break this task
+# Adding from_json so that the JSON output is parsed into a dictionary
 - name: Create info variables from bundle generation output
   ansible.builtin.set_fact:
-    sto_bundle_info: "{{ generate_bundle_sto.stdout }}"
-    sgo_bundle_info: "{{ generate_bundle_sgo.stdout }}"
+    sto_bundle_info: "{{ generate_bundle_sto.stdout_lines[-1] | from_json }}"
```
csibbitt, line 12: :+1:

csibbitt (approved):

build/generate_bundle.sh
```diff
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
-set -e
-REL=$(dirname "$0")
+set -x
```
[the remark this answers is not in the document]
elfiesmelfie: No, It might have been lost when resolving a merge conflict

elfiesmelfie, line 2:
```suggestion
set -e
set -x
```

build/generate_bundle.sh
```diff
@@ -1,6 +1,10 @@
...
+
+LOGFILE=${LOGFILE:-/dev/null}
+truncate --size=0 $LOGFILE
```
elfiesmelfie, line 5:
```suggestion
# If LOGFILE is /dev/null, this command fails, so ignore that error
truncate --size=0 ${LOGFILE} || true
```

build/stf-run-ci/tasks/create_builds.yml
```diff
@@ -23,11 +23,25 @@
 
   - name: Kill first build since it will always fail (triggered on BuildConfig creation)
     ansible.builtin.shell: sleep 10 ; oc delete build {{ artifact.name }}-1 -n "{{ namespace }}"
+    ignore_errors: true
+    retries: 3
+    delay: 10
+    register: kill_build
+    until: kill_build.rc == 0
```
elfiesmelfie, line 30: This is added to cover the case where there's an error sending the request e.g. a timeout, which I saw a few times when testing this in zuul

build/stf-run-ci/tasks/setup_stf_local_build.yml
```diff
@@ -3,25 +3,27 @@
...
+        RELATED_IMAGE_BRIDGE_SMARTGATEWAY_TAG={{ sg_bridge_image_path | parse_tag | quote }} \
+        OPERATOR_IMAGE={{ sgo_image_path | parse_image | quote }} \
+        OPERATOR_TAG={{ sgo_image_path | parse_tag | quote }} \
+        ./generate_bundle.sh
+    register: generate_bundle_sgo
+  always:
+  - name: Show generate bundle log
+    ansible.builtin.shell:
+      cmd: |
+        cat {{ logfile_dir }}/sgo_gen_bundle.log || true
```
elfiesmelfie, line 26: the `|| true`  may not be needed here. I added it in when I was getting errors because the logfile wasn't being created

build/stf-run-ci/tasks/setup_stf_local_build.yml
```diff
@@ -55,7 +57,7 @@
   ansible.builtin.shell:
     chdir: "{{ base_dir }}"
     cmd: |
-      LOGFILE="{{ ansible_env.HOME }}/sto_gen_bundle.log" \
```
elfiesmelfie, line 58: Since the ansible_env is being removed, it's likely that the "gather_facts:" in build/run-ci.yaml can also be removed.

build/stf-run-ci/tasks/main.yml
```diff
@@ -79,6 +79,15 @@
...
     creates: "{{ base_dir }}/working/operator-sdk-{{ operator_sdk_v1 }}"
 
+- name: Set logfile_dir
+  when: not (logfile_dir is defined)
+  ansible.builtin.set_fact:
+    logfile_dir: "{{ base_dir }}/working/logs"
+
+- name: Make sure the logging dir exists
+  ansible.builtin.shell:
+    cmd: mkdir -p {{ logfile_dir }}
```
leifmadsen, line 89: This could also add a `creates` which will help in the future if we enable other lint testing which will complain here about idempotency. Also what about just setting a default path instead of setting it inline?

build/stf-run-ci/tasks/create_builds.yml
```diff
@@ -37,10 +51,12 @@
   ansible.builtin.set_fact:
     build_describe: "{{ build_describe_results.stdout | from_json }}"
 
-- ansible.builtin.debug:
+- name: Show the ``oc get build ...`` results
```
leifmadsen, line 54: I know it's minor and doesn't really matter, but since this isn't RST I think it'd be better just to say "Get the Build results"

leifmadsen (approved):
Couple minor things but overall looks good

vkmc (approved):
Nothing much to add to previous reviews, overall looks good

build/generate_bundle.sh
```diff
@@ -30,7 +32,7 @@ generate_bundle() {
     REPLACE_REGEX="s#<<CREATED_DATE>>#${CREATED_DATE}#g;s#<<OPERATOR_IMAGE>>#${OPERATOR_IMAGE}#g;s#<<OPERATOR_TAG>>#${OPERATOR_TAG}#g;s#<<RELATED_IMAGE_PROMETHEUS_WEBHOOK_SNMP>>#${RELATED_IMAGE_PROMETHEUS_WEBHOOK_SNMP}#g;s#<<RELATED_IMAGE_PROMETHEUS_WEBHOOK_SNMP_TAG>>#${RELATED_IMAGE_PROMETHEUS_WEBHOOK_SNMP_TAG}#g;s#<<RELATED_IMAGE_OAUTH_PROXY>>#${RELATED_IMAGE_OAUTH_PROXY}#g;s#<<RELATED_IMAGE_OAUTH_PROXY_TAG>>#${RELATED_IMAGE_OAUTH_PROXY_TAG}#g;s#<<OPERATOR_BUNDLE_VERSION>>#${OPERATOR_BUNDLE_VERSION}#g;s#1.99.0#${OPERATOR_BUNDLE_VERSION}#g;s#<<OPERATOR_DOCUMENTATION_URL>>#${OPERATOR_DOCUMENTATION_URL}#g;s#<<BUNDLE_OLM_SKIP_RANGE_LOWER_BOUND>>#${BUNDLE_OLM_SKIP_RANGE_LOWER_BOUND}#g"
 
     pushd "${REL}/../" > /dev/null 2>&1
-    ${OPERATOR_SDK} generate bundle --verbose --channels ${BUNDLE_CHANNELS} --default-channel ${BUNDLE_DEFAULT_CHANNEL} --manifests --metadata --version "${OPERATOR_BUNDLE_VERSION}" --output-dir "${WORKING_DIR}" >> ${LOGFILE} 2>&1
```
vkmc, line 33: Do we want to drop --verbose?

build/stf-run-ci/tasks/main.yml
elfiesmelfie: We can't set a default in defaults/main, because the default value is to be relative to base_dir.

That's why there's a "set if not defined"

build/generate_bundle.sh
elfiesmelfie: That should still be there. I think this change was originally before [1], which introduced a change to the same line.
[1] https://github.com/infrawatch/service-telemetry-operator/pull/463

PR #472, opened 2023-09-19
Install ObO during index-based deployments

leifmadsen:
Because Observability Operator (ObO) is a cluster-scoped Operator, the OLM dependency management can't resolve the dependency for us. Update setup_base to pre-install ObO when observabilityStrategy is use_redhat or use_hybrid, even when when index-based deployment is enabled.

Resolves: STF-1483

### DIFF
--- a/build/stf-run-ci/tasks/setup_base.yml
+++ b/build/stf-run-ci/tasks/setup_base.yml
@@ -121,24 +121,24 @@
       when:
         - __service_telemetry_observability_strategy == "use_community"
 
-    - name: Subscribe to Red Hat Obervability Operator
-      kubernetes.core.k8s:
-        definition:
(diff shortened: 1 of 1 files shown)


csibbitt (approved):
Whitespace-only changes that have a functional impact (vs a broken syntax correction) kinda freak me out.

elfiesmelfie (approved):
It looks like it should work. 
Jenkins isn't going to confirm that it works, since we're looking at index-based deployments.

A Jenkins pass on this will confirm that other things aren't broken.

vkmc (approved):
Code looks good, I can test this in one of my environments

leifmadsen:
Awesome thanks @vkmc ! Let me know if you run into any issues.

PR #473, opened 2023-09-19
Convert alertmanager-stf Role to ClusterRole

leifmadsen:
Convert alertmanager-stf Role to ClusterRole as the tokenreviews and
subjectaccessreviews resources need to be accessable at the cluster
scope.


csibbitt (approved):

elfiesmelfie (approved):
This looks like it does what it says on the tin. It's being merged into a feature branch, so there's more time for adjustments

leifmadsen:
Heh I probably should have put this in draft mode and local testing since this doesn't work :) In the process of a new patch with local testing before posting.

PR #479, opened 2023-09-22
[zuul] Add base job for testing with zuul

elfiesmelfie:
This doesn't configure any additional jobs to run for the moment. It is the base of the rest of the jobs. 

* Update default crc_parameters, so that there's enough memory for cluster_monitoring and for schedulling all the pods [1]
* Add a 60 minute timeout to the job, since the default (~30 minutes) is not enough to run through all the job stages
* galaxy collection and pip package versions are pinned to known-working versions
* post task: Gathers some information from oc (builds, images, CSVs, etc) Copies logs from the job nodes so they can be viewed in the job buildresults

Depends-On: https://github.com/infrawatch/service-telemetry-operator/pull/484


elfiesmelfie:
This should break nothing, but will be required for the later job definitions to merge

ci/deploy_stf.yml
```diff
@@ -0,0 +1,25 @@
...
+        file: "vars-zuul-common.yml"
+
+    - name: "Get scenario-specific vars"
+      ansible.builtin.include_vars:
+        file: "vars-{{ scenario }}.yml"
+
+    - name: "Log into the cluster"
+      ansible.builtin.import_role:
+        name: rhol_crc
+        tasks_from: add_crc_creds.yml
```
csibbitt, line 21: In case anyone else is looking for this, it comes from here https://github.com/openstack-k8s-operators/ci-framework/blob/9f9d01c1611a9408903e2aab5f6264d406891936/ci_framework/roles/rhol_crc/tasks/add_crc_creds.yml  which is checked out by the zuul job above

ci/post-collect_logs.yml
```diff
@@ -0,0 +1,166 @@
...
+      ansible.builtin.shell:
+        cmd: |
+          oc get subscriptions > {{ logfile_dir }}/post_oc_get_subscriptions.log 2>&1
+          oc describe subscription service-telemetry-operator >> {{ logfile_dir }}/post_oc_get_subscriptions.log 2>&1
+          cat {{ logfile_dir}}/post_oc_get_subscriptions.log
+
+    - name: "Get smoketest logs"
+      ansible.builtin.shell:
+        cmd: |
+          echo "*** [INFO] Showing oc get all..." > {{ logfile_dir }}/oc_log.log 2>&1
```
csibbitt, line 62: It looks like this is basically a copy/paste from https://github.com/infrawatch/service-telemetry-operator/blob/master/tests/smoketest/smoketest.sh#L117 It seems a shame to maintain this meatloaf in two places. 

All of these are going to the same "oc_log.log" file, but it looks like we are already capturing the same output in `{{ logfile_dir }}/smoketest.log` below [1]. Am I missing a reason that we'd need to capture all of this twice?


[1] https://github.com/infrawatch/service-telemetry-operator/pull/479/files#diff-84232940de50ce920ee1af83cd272927075b521e821c56a4fd1933f93dd5eb24R26

ci/prepare.yml
```diff
@@ -0,0 +1,44 @@
...
+      ansible.builtin.pip:
+        name: pip
+        state: latest
+        extra_args: "-U"
+
+    - name: "Install pre-reqs from pip"
+      ansible.builtin.pip:
+        name:
+            # https://stackoverflow.com/questions/64073422/importerror-cannot-import-name-oauth1session-from-requests-oauthlib
+          - requests==2.27.1
```
csibbitt, line 15: These seem like something that should maybe live in `roles/stf-run-ci/requirements.txt`?

csibbitt (approved):
This LGTM in general and I think it's mergeable, but have added a couple suggestions to try to stave off future duplication errors.

ci/post-collect_logs.yml
vkmc: +1

elfiesmelfie: Missed in refactoring, thanks for pointing it out. This re-generation of logs was done before I added the updates to improve logging

ci/prepare.yml
elfiesmelfie: Yes, they probably should. It's a minor improvement that would help with other systems using this role

elfiesmelfie: I added a requirements.txt here: https://github.com/infrawatch/service-telemetry-operator/pull/484 

The Depends-On in the PR description should check out the dependant PR and let both be checked together.
I'll push the update for this file shortly

csibbitt (approved):

ci/prepare.yml
```diff
@@ -0,0 +1,37 @@
...
+        chdir: "{{ playbook_dir }}"
+        state: present
+
+    - name: "Install ansible collections"
+      community.general.ansible_galaxy_install:
+        type: collection
+        name: "{{ item }}"
+      with_items:
+        - "kubernetes.core:2.3.2"
+        - "community.general:6.2.0"
```
leifmadsen, line 23: Is there a better place to control these versions than hard coding them into the play?

ci/prepare.yml
```diff
@@ -0,0 +1,37 @@
...
+    - name: "Log into the cluster"
+      ansible.builtin.import_role:
+        name: rhol_crc
+        tasks_from: add_crc_creds.yml
+
+    - name: "Create the service-telemetry project"
+      kubernetes.core.k8s:
+        api_version: v1
+        kind: Namespace
+        name: "service-telemetry"
```
leifmadsen, line 34: Our current upstream ends up creating different namespaces base on PR name etc which at least helps validate that everything can work outside of a statically named project. I think it would be useful to keep that going because I've seen it catch at least one issue in the past.

leifmadsen: Seems there is a `{{ namespace }}` variable that could be used here at the very least

ci/vars-zuul-common.yml
```diff
@@ -0,0 +1,6 @@
+---
+namespace: "service-telemetry"
```
leifmadsen, line 2: Is it possible to have this dynamically defined based on PR etc?

ci/deploy_stf.yml
```diff
@@ -0,0 +1,25 @@
...
+
+    - name: "Deploy STF using stf-run-ci"
+      ansible.builtin.import_role:
+        name: '../build/stf-run-ci'
```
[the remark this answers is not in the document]
elfiesmelfie: {{ sto_dir }}/ci/

ci/prepare.yml
elfiesmelfie: It could be put into a requirements.yml file under {{ sto_dir }}/build/stf-run-ci. 

The format of that MAY depend on the version of ansible/galaxy

ci/deploy_stf.yml
```diff
@@ -0,0 +1,25 @@
...
+        sto_dir: '{{ ansible_env.HOME }}/{{ zuul.project.src_dir }}'
+      when: sto_dir | default('') | length == 0
+
+    - name: "Get vars common to all jobs"
+      ansible.builtin.include_vars:
+        file: "vars-zuul-common.yml"
+
+    - name: "Get scenario-specific vars"
+      ansible.builtin.include_vars:
+        file: "vars-{{ scenario }}.yml"
```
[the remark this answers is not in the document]
elfiesmelfie: It is always set, the in a concrete job.

The first one is added in:
https://github.com/infrawatch/service-telemetry-operator/pull/480/files#diff-108978819c05ae183d88ec87959c2341a94cfc3f9465e3aeee82d554217b4f58R37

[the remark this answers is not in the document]
elfiesmelfie: If there's a default, we may end up in a case where the var is not set explicitly and we don't notice because the it defaults.

IMHO, not having a default means that we avoid the wrong value being set, but also get an early fail if the var is omitted. The error for an undefined var is pretty explicit

ci/vars-zuul-common.yml
elfiesmelfie: With zuul, a new env is created each time.

However, if you want to use these playbooks locally (e.g. for recreating issues), it would be useful to have that separation.

leifmadsen: I'm more worried about implementing changes and only testing with `service-telemetry` and then finding that if someone uses an alternative namespace, that everything gets broken. I've definitely implemented something in the past that was caught by using different namespace values in the CI, and if possible I'd like to keep that test functionality.

It could be as simple as using `set_fact` and some `random()` function or whatever, and then using that as the namespace value across the test?

.zuul.yaml
```diff
@@ -1,4 +1,33 @@
...
+      - ci/deploy_stf.yml
+      - ci/test_stf.yml
+    post-run:
+      - ci/post-collect_logs.yml
+    nodeset: centos-9-crc-xxl
+    # The default (~30 minutes) is not enough to run through all the job stages
+    timeout: 3600
+    vars:
+      # vars for base-simple-crc
+      crc_config_cluster_monitoring: true
```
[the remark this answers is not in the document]
elfiesmelfie: Excellent.

ci/vars-zuul-common.yml
elfiesmelfie: Zuul definitely has some var that would be suitable. Maybe not PR number, but a build number or something similar

elfiesmelfie: FYI, you can find all the zuul vars in the logs tab in a build, under zuul-info/inventory.yaml
zuul.change is the PR number, and zuul.build is the build number (same as the URL below), I think they'll do nicely

e.g. https://review.rdoproject.org/zuul/build/8edc7f4a83a6463b9c40f9263af82e84/log/zuul-info/inventory.yaml

ci/prepare.yml
leifmadsen: Won't block approving/merging this without the change, but I always get itchy with hard coded versions especially as we have dependabot which will propose updating library versions.

ci/vars-zuul-common.yml
```diff
@@ -0,0 +1,6 @@
+---
+namespace: "service-telemetry-PR#{{ zuul.change }}-{{ zuul.build }}"
```
leifmadsen, line 2: 👍🏻 thank you!

leifmadsen: I'll be curious if things fail due to namespace being too long ever :)

leifmadsen (approved):

PR #482, opened 2023-09-26
Add local generation difference check

leifmadsen:
Check if generating a bundle locally would result in a git diff. If so,
then we should fail since that means changes are missed in the resulting
bundle.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -38,6 +38,31 @@ jobs:
         run: ${HOME}/.local/bin/ansible-lint .
         working-directory: ./build/stf-run-ci
 
+  generate-bundle-diff-check:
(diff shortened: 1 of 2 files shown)


leifmadsen:
Initial commit here is an intentional fail (ehrmegherd is this TDD?)

leifmadsen:
> I like this because I often forget to regenerate some of the OLM artifacts, but thinking about it makes me wonder if these things need to be checked into the repo at all, or should we just generate them as part of the build process? My thinking hasn't gone any deeper than that, though.

In theory we could really just have GHA do the local generation and then commit changes automatically, but this was something quick to do which seemed to at least say "hey you forgot this!".

I agree though, there really doesn't need to be a need to self-generate these when CI could really just clean it all up for you.

PR #483, opened 2023-09-26
Remove logs API interface from STF

leifmadsen:
The logs interface in STF has never been supported or productized. A
recent bug was found where Loki can be interfered with when deployed on
the same system as STF. Since this has never been documented or
productized, it is being removed, as OpenStack logging should be sent
off-cluster from rsyslog directly to Elasticsearch.

Closes STF-1504


leifmadsen:
Moving out of draft now that I know it'll deploy without error. Going to do some more visual validations today, but I think this is pretty much ready to go pending review.

leifmadsen:
No longer have logs related collectors or backends.

image

deploy/olm-catalog/service-telemetry-operator/manifests/service-telemetry-operator.clusterserviceversion.yaml
```diff
@@ -253,6 +242,7 @@ spec:
           - watch
           - update
           - patch
+          - delete
```
leifmadsen, line 245: This will go away in this patch once https://github.com/infrawatch/service-telemetry-operator/pull/482 lands and is merged here.

elfiesmelfie (approved):
It looks reasonable

vyzigold:
Can't find any relevant mention of logs or loki in the code anymore, so LGTM.
Bye logs API :wave:

vyzigold (approved):

PR #490, opened 2023-09-27
Increase sensubility smart-gateway ring buffer

leifmadsen:
Increase the ring buffer size of the sensubility Smart Gateway as
messages often exceed the default of 16384 bytes, resulting in no API
healt check data from controllers arriving.

Closes: rhbz#2241033


csibbitt (approved):
Seems a bit weird to go from (2^14) to (2^16 - 1), but okay. :)

leifmadsen:
> Seems a bit weird to go from (2^14) to (2^16 - 1), but okay. :)

Yea the jump is quite big, but the error message I was getting was that the message was 40000, larger than 16384, so I picked the next exponential that I could remember off the top of my head :)

PR #491, opened 2023-09-28
Disable event pipeline management

leifmadsen:
Disable the event pipeline management in the enable-stf.yaml file for the RHOSP 17.1 helper script to align to STF 1.5.3 disabling eventing by default, matching our documentation changes.

Related STF-1498


leifmadsen:
This aligns the infrared helper script in 17.1 to match STF 1.5.3. Our CI will _not_ change and will continue to deploy eventing by default to verify that it continues to work. The helper scripts should align against our default deployment and  documentation methods though so that we can verify and test things based on what users would deploy if following docs.

csibbitt (approved):

PR #493, opened 2023-09-28
Move the metrics_result retrieve

leifmadsen:
Move the metrics_result value retrieval to above an echo command so that
the result is of the curl command and not the echo command. Also change
the scripts to set +e so that the scripts don't exist right away since
there are checks at the end that set an appropriate exit code (and gives
more information since the script will complete vs exit immediately).

Reported by Chris in Slack

### DIFF
--- a/tests/smoketest/smoketest_ceilometer_entrypoint.sh
+++ b/tests/smoketest/smoketest_ceilometer_entrypoint.sh
@@ -29,7 +29,6 @@ echo "*** [INFO] Checking for recent image metrics..."
 echo "[DEBUG] Running the curl command to return a query"
 curl -k -u "internal:${PROMETHEUS_AUTH_PASS}" -g "${PROMETHEUS}/api/v1/query?" --data-urlencode 'query=ceilometer_image_size' 2>&1 | grep '"result":\[{"metric":{"__name__":"ceilometer_image_size"'
 metrics_result=$?
-echo "[DEBUG] Query returned"
 echo "[DEBUG] Set metrics_result to $metrics_result"
 
 if [ "$OBSERVABILITY_STRATEGY" != "use_redhat" ]; then
--- a/tests/smoketest/smoketest_ceilometer_entrypoint.sh
+++ b/tests/smoketest/smoketest_ceilometer_entrypoint.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-set -e
+set +e
 
 # Executes inside the test harness container to start collectd and look for resulting metrics in prometheus
 PROMETHEUS=${PROMETHEUS:-"https://default-prometheus-proxy:9092"}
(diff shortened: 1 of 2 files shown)


leifmadsen:
I tested with this diff to make sure an invalid metric find would result in a failed test, which it did.

```diff
diff --git a/tests/smoketest/smoketest_ceilometer_entrypoint.sh b/tests/smoketest/smoketest_ceilometer_entrypoint.sh
index 674a6e2..b4c67c4 100644
--- a/tests/smoketest/smoketest_ceilometer_entrypoint.sh
+++ b/tests/smoketest/smoketest_ceilometer_entrypoint.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-set -e
+set +e
 
 # Executes inside the test harness container to start collectd and look for resulting metrics in prometheus
 PROMETHEUS=${PROMETHEUS:-"https://default-prometheus-proxy:9092"}
@@ -27,9 +27,9 @@ echo; echo
 echo "*** [INFO] Checking for recent image metrics..."
 
 echo "[DEBUG] Running the curl command to return a query"
-curl -k -u "internal:${PROMETHEUS_AUTH_PASS}" -g "${PROMETHEUS}/api/v1/query?" --data-urlencode 'query=ceilometer_image_size' 2>&1 | grep '"result":\[{"metric":{"__name__":"ceilometer_image_size"'
-echo "[DEBUG] Query returned"
+curl -k -u "internal:${PROMETHEUS_AUTH_PASS}" -g "${PROMETHEUS}/api/v1/query?" --data-urlencode 'query=ceilometer_image_size' 2>&1 | grep '"result":\[{"metric":{"__name__":"ceilometer_imge_size"'
 metrics_result=$?
+echo "[DEBUG] Query returned"
 echo "[DEBUG] Set metrics_result to $metrics_result"
 
 if [ "$OBSERVABILITY_STRATEGY" != "use_redhat" ]; then
diff --git a/tests/smoketest/smoketest_collectd_entrypoint.sh b/tests/smoketest/smoketest_collectd_entrypoint.sh
index a8ce110..d7f5132 100755
--- a/tests/smoketest/smoketest_collectd_entrypoint.sh
+++ b/tests/smoketest/smoketest_collectd_entrypoint.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-set -e
+set +e
 
 # Executes inside the test harness container to start collectd and look for resulting metrics in prometheus
 PROMETHEUS=${PROMETHEUS:-"https://default-prometheus-proxy:9092"}
```

tests/smoketest/smoketest_collectd_entrypoint.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
-set -e
```
csibbitt, line 2: Ah yes, I noticed the early exit sometimes today. I usually like `-e` for test scripts like this, but at some point I guess it was no longer correct and we're just noticing now. :)

tests/smoketest/smoketest_ceilometer_entrypoint.sh
```diff
@@ -28,8 +28,8 @@ echo "*** [INFO] Checking for recent image metrics..."
 
 echo "[DEBUG] Running the curl command to return a query"
 curl -k -u "internal:${PROMETHEUS_AUTH_PASS}" -g "${PROMETHEUS}/api/v1/query?" --data-urlencode 'query=ceilometer_image_size' 2>&1 | grep '"result":\[{"metric":{"__name__":"ceilometer_image_size"'
-echo "[DEBUG] Query returned"
 metrics_result=$?
+echo "[DEBUG] Query returned"
```
csibbitt, line 32:
```suggestion
```

csibbitt (approved):
Sure. I'd be happy to just drop the echo since I'm not sure it adds anything

tests/smoketest/smoketest_collectd_entrypoint.sh
leifmadsen: Yea this is kind of an embedded meta-script and since it has control logic for evaluating an exit code of 0 or 1, I figured we should just let them fall through.

PR #494, opened 2023-09-28
[zuul] Add job to deploy from nightly bundles

elfiesmelfie:
This job doesn't build STF, but deploys from the pre-built and published bundles.
This is useful to be able to do periodically to make sure our latest bundles are deploying, and no dependencies are out-of-date, for example

Depends-On: https://github.com/infrawatch/service-telemetry-operator/pull/496
Depends-On: https://github.com/infrawatch/service-telemetry-operator/pull/497


ci/vars-nightly_bundles.yml
```diff
@@ -0,0 +1,7 @@
+---
+# from: https://github.com/infrawatch/service-telemetry-operator/pull/437
+# ansible-playbook -e __service_telemetry_storage_ephemeral_enabled=true -e __local_build_enabled=false -e __deploy_from_bundles_enabled=true -e __service_telemetry_bundle_image_path=quay.io/infrawatch-operators/service-telemetry-operator-bundle:nightly-head -e __smart_gateway_bundle_image_path=quay.io/infrawatch-operators/smart-gateway-operator-bundle:nightly-head --skip-tags bundle_registry_tls_ca --skip-tags bundle_registry_auth build/run-ci.yaml
+
+__local_build_enabled: false
+__deploy_from_bundles_enabled: true
+__service_telemetry_storage_ephemeral_enabled: true
```
leifmadsen, line 7: Does CRC provide pre-configured PVs? I'd actually prefer if we could stop using ephemeral storage in our testing scenarios if possible. It's probably not a huge deal, but since our documentation uses persistent storage, it'd be ideal to test that part of the code path.

ci/vars-nightly_bundles.yml
```diff
@@ -0,0 +1,7 @@
+---
+# from: https://github.com/infrawatch/service-telemetry-operator/pull/437
+# ansible-playbook -e __service_telemetry_storage_ephemeral_enabled=true -e __local_build_enabled=false -e __deploy_from_bundles_enabled=true -e __service_telemetry_bundle_image_path=quay.io/infrawatch-operators/service-telemetry-operator-bundle:nightly-head -e __smart_gateway_bundle_image_path=quay.io/infrawatch-operators/smart-gateway-operator-bundle:nightly-head --skip-tags bundle_registry_tls_ca --skip-tags bundle_registry_auth build/run-ci.yaml
```
leifmadsen, line 3: We might have to start considering tests that span `stable-1.5` and `master` since the nightly builds are results from different branches and thus different code bases. As we move towards STF 2.0 (and even STF 1.5.4) we'll have different code bases that could really both use separate test paths.

leifmadsen: (not for now, I'm happy to just see this start landing against `master` for now, but this code will eventually make its way into a stable-1.5 branch at some point, and we'll need to make some adjustments when that happens in a week or two)

elfiesmelfie: This nightly check might well just run periodically, once a week or so.
To get tests spanning multiple branches, we need to add the job to that branch.

I think we also need a few more updates to get it working, but I'm focussing on getting the jobs on master first

leifmadsen: Yep makes sense. Just noting that once we merge this to stable-1.5 as part of STF 1.5.3 then these will technically be the wrong tags.

ci/vars-nightly_bundles.yml
leifmadsen: Note that ephemeral storage was only used historically because there is no PV in quicklab, where our tests were originally running. We should be trying to use persistent storage whenever possible.

csibbitt: Agree, if CRC will provide PVs, that would be ideal.

csibbitt (approved):

ci/vars-nightly_bundles.yml
elfiesmelfie: That's something to try later, I think. This set of jobs (local_build, local_build-index_deploy, nightly-bundles) need to get merged first so we have testing on PRs

elfiesmelfie: I've added a spike to the next sprint to try answer these questions, and scope adding PVs. It might be trivial, not possible or require changes in rdo-jobs

elfiesmelfie:
recheck

elfiesmelfie:
recheck

I changed the base of the PR

PR #497, opened 2023-09-29
[zuul] use 'service-telemetry' as namespace

elfiesmelfie:
The format was incorrect, and cannot include '#' or capital letters.

https://review.rdoproject.org/zuul/build/7b4f5f48368a41d2be73044bbd00abed

### DIFF
--- a/ci/vars-zuul-common.yml
+++ b/ci/vars-zuul-common.yml
@@ -1,5 +1,5 @@
 ---
-namespace: "service-telemetry-PR#{{ zuul.change }}-{{ zuul.build }}"
+namespace: "service-telemetry"
 setup_bundle_registry_tls_ca: false
 setup_bundle_registry_auth: false
 base_dir: "{{ sto_dir }}/build"


ci/vars-zuul-common.yml
```diff
@@ -1,5 +1,5 @@
 ---
-namespace: "service-telemetry-PR#{{ zuul.change }}-{{ zuul.build }}"
+namespace: "service-telemetry"
```
leifmadsen, line 2: Can you not just remove the # part and add a `| lower` function here? I'd still rather not lose the non-default namespace test.

leifmadsen: I'll reluctantly approve this to get things working, but can you file a follow up task to come back and make sure we can test with dynamic namespaces?

leifmadsen (approved):

ci/vars-zuul-common.yml
elfiesmelfie: I've created an issue to track this internally

PR #499, opened 2023-10-05
Add requires infrastructure annotations

leifmadsen:
Add required infrastructure annotations for the bundle. Implementation
is done in generate_bundle.sh because the annotations.yaml file in the
deploy/olm-catalog/ directory is not read by operator-sdk-0.19.4. Append
required additional feature annotations to the generated
annotations.yaml by operator-sdk generate bundle.

Related STF-1530

### DIFF
--- a/build/generate_bundle.sh
+++ b/build/generate_bundle.sh
@@ -36,18 +36,6 @@ generate_bundle() {
     popd > /dev/null 2>&1
 
     sed -i -E "${REPLACE_REGEX}" "${WORKING_DIR}/manifests/${OPERATOR_NAME}.clusterserviceversion.yaml"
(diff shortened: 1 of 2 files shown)


leifmadsen:
I don't think these are supposed to land in the annotations.yaml so I'm likely moving these to the CSV directly. More testing in progress.

leifmadsen:
I was waiting on some additional feedback because best on my testing and such I think I approached this wrong.

The feedback received is that I am indeed doing it right, and that the annotations should be added to the CSV directly, as the extra items in the annotations.yaml don't seem to get parsed back into the CSV like some of the existing items.

Changes coming shortly as I craft the updated CSV and remove the generate_bundle.sh changes.

vkmc (approved):
LGTM

PR #504, opened 2023-10-19
Cluster Observability Operator dependency

leifmadsen:
Add cluster observability operator as the preferred dependency (bottom
of list is highest priority) when installing Service Telemetry Operator.
The cluster-observability-operator is the name of the downstream
(product) bundle in the Red Hat Operators CatalogSource.

If installing for upstream, preferred operator will be
observability-operator (when the Red Hat Operators CatalogSource is not
available or enabled). And then as a fall-back method when neither
Observability Operator or Cluster Observability Operator is not
available, allow for Prometheus Operator from the Community Operators to
satisfy for the Prometheus storage backend.

### DIFF
--- a/deploy/olm-catalog/service-telemetry-operator/metadata/properties.yaml
+++ b/deploy/olm-catalog/service-telemetry-operator/metadata/properties.yaml
@@ -39,3 +39,6 @@ properties:
           - package:
               packageName: observability-operator
               versionRange: '>=0.0.1'
+          - package:
+              packageName: cluster-observability-operator
+              versionRange: '>=0.0.1'


leifmadsen:
recheck

deploy/olm-catalog/service-telemetry-operator/metadata/properties.yaml
```diff
@@ -39,3 +39,6 @@ properties:
           - package:
               packageName: observability-operator
               versionRange: '>=0.0.1'
+          - package:
+              packageName: cluster-observability-operator
```
csibbitt, line 43: This is the same thing, right? `observability-operator` is renaming to `cluster-observability-operator`? Do we have any idea what happens if you've already installed it under the old name?

leifmadsen: So basically the downstream (Red Hat Operators CatalogSource) will ship observability-operator as cluster-observability-operator. If you didn't have Red Hat Operators (and only Community Operators) then the observability-operator package we be available. Then finally if you don't have either of those for some reason, then prometheus-operator is the one it'd use.

Basically, any of these being installed will satisfy, but it will try to install in the order (bottom to top) as the preference.

So basically, in a new installation, cluster-observability-operator would be preferred over observability-operator. If prometheus-operator were installed, then that would satisfy for the dependencies, and would then pass the dependency checks.

Basically, if you already have observability-operator installed, the dependency check will pass, and OLM will move onto the next set of package manifest checks.

csibbitt: Aha! I didn't realize the old naming would remain valid in the community operators catalog; thanks.

PR #511, opened 2023-10-26
[stf-run-ci] Fix check to include bool filter

elfiesmelfie:
Update the check to use bool filter instead of a bar var. By default, ansible parses vars as strings, and without the | bool filter, this check is invalid, as it will always resolve to true, since it is a non-empty string. Other instances of the same check did this, but this one was missed.

### DIFF
--- a/build/stf-run-ci/tasks/setup_stf_from_bundles.yml
+++ b/build/stf-run-ci/tasks/setup_stf_from_bundles.yml
@@ -64,7 +64,7 @@
       data:
         cert.pem: "{{ lookup('file', 'CA.pem') | b64encode }}"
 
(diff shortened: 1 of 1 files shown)


leifmadsen (approved):

leifmadsen:
Now that PRs are opened to merge master into stable-1.5, we're going to have to be aware of a requirement to cherry pick these back to stable-1.5.

I would block on merging anything into master until the STF 1.5.3 work is landed and tagged.

elfiesmelfie:
This was missed in: https://github.com/infrawatch/service-telemetry-operator/pull/474/files

vkmc (approved):
LGTM

csibbitt (approved):

PR #512, opened 2023-10-26
[allow_skip_clone] Allow skipping of the cloning stages

elfiesmelfie:
Use <repo>_dir instead of hardcoding all directories relative to base_dir

This will allow configuration of the repo clone destination, so we can use pre-cloned dirs instead of explicitly cloning the dirs each time.

This is essential for CI systems like zuul, that set-up the repos with particular versions/branches prior to running the test scripts.

### DIFF
--- a/ci/vars-zuul-common.yml
+++ b/ci/vars-zuul-common.yml
@@ -5,3 +5,4 @@ setup_bundle_registry_auth: false
 __service_telemetry_transports_qdr_auth: none
 base_dir: "{{ sto_dir }}/build"
 logfile_dir: "{{ ansible_user_dir }}/zuul-output/logs/controller"
+clone_repos: false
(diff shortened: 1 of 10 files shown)


elfiesmelfie:
I forgot to actually set the values for *_dir

ci/vars-zuul-common.yml
```diff
@@ -5,3 +5,8 @@ setup_bundle_registry_auth: false
 __service_telemetry_transports_qdr_auth: none
 base_dir: "{{ sto_dir }}/build"
```
vkmc, line 6: Where is sto_dir defined for Zuul?

csibbitt: When zuul runs, the entrypoint is deploy_stf.yml, which sets sto_dir right away[1] before importing the variables in this file.

[1] https://github.com/infrawatch/service-telemetry-operator/pull/512/files#diff-2873fa0b1fe3441a27c4eead4f6fb7a81095410bbfdfdb7f011526e9f555fa9cR7

csibbitt (approved):

vkmc (approved):

PR #514, opened 2023-11-02
[zuul] Define a project template for stf-crc-jobs

elfiesmelfie:
Instead of listing all the jobs for each project in-repo, and needing to update the list every time that a new job is added, the project template can be updated and the changes propagated to the other infrawatch projects

Depends-On: https://review.rdoproject.org/r/c/config/+/50707


leifmadsen (approved):
Looks safe enough :)

How does the project template get used? Seems like the `project` definition under the `project-template` would get updated to just refer to the template instead of defining the `github-check` and `jobs`?

leifmadsen:
Ah looks like you tried to use the template but it's not defined. Does this need to land first so that it can find it?

elfiesmelfie:
> Ah looks like you tried to use the template but it's not defined. Does this need to land first so that it can find it?

I think I need to make a change to rdo zuul so that project templates can be imported from projects.
https://review.rdoproject.org/cgit/config/tree/zuul/rdo.yaml#n1243

I'll keep this as a draft for now, until I can get that change done in rdo.

elfiesmelfie:
recheck

csibbitt (approved):

PR #518, opened 2023-11-03
Fix qdr auth one_time_upgrade label check

csibbitt:
Bug is here: https://github.com/infrawatch/service-telemetry-operator/blob/master/roles/servicetelemetry/tasks/component_qdr.yml#L152

We need to upgrade the default password that is generated by the Interconnect Operator because it is extremely weak. The logic on L177 is supposed to pair with the label on L173 to make this a "one time" operation. However, you can see on L158 it's querying and registering the Interconnect object itself to _qdr_basicauth_object, not the Secret, so `_qdr_basicauth_object[0].metadata.labels.stf_one_time_upgrade` will never be defined (also it should be `_qdr_basicauth_object.resources[0]`, so a double-bug)

### DIFF
--- a/roles/servicetelemetry/tasks/component_qdr.yml
+++ b/roles/servicetelemetry/tasks/component_qdr.yml
@@ -172,7 +172,7 @@
             labels:
               stf_one_time_upgrade: "{{ lookup('pipe', 'date +%s') }}"
           stringData:
-            guest: "{{ lookup('password', '/dev/null') }}"
+            guest: "{{ lookup('password', '/dev/null chars=ascii_letters,digits length=32') }}"
       when:
         - _qdr_basicauth_object.resources[0] is defined and _qdr_basicauth_object.resources[0].metadata.labels.stf_one_time_upgrade is not defined
 
--- a/roles/servicetelemetry/tasks/component_qdr.yml
+++ b/roles/servicetelemetry/tasks/component_qdr.yml
@@ -154,9 +154,9 @@
   block:
     - name: Get QDR BasicAuth secret
       k8s_info:
-        api_version: interconnectedcloud.github.io/v1alpha1
-        kind: Interconnect
-        name: "{{ ansible_operator_meta.name }}-interconnect"
+        api_version: v1
+        kind: Secret
+        name: "{{ ansible_operator_meta.name }}-interconnect-users"
         namespace: "{{ ansible_operator_meta.namespace }}"
       register: _qdr_basicauth_object
 
--- a/roles/servicetelemetry/tasks/component_qdr.yml
+++ b/roles/servicetelemetry/tasks/component_qdr.yml
@@ -174,7 +174,7 @@
           stringData:
             guest: "{{ lookup('password', '/dev/null') }}"
       when:
-        - _qdr_basicauth_object.resources[0] is defined and _qdr_basicauth_object[0].metadata.labels.stf_one_time_upgrade is not defined
+        - _qdr_basicauth_object.resources[0] is defined and _qdr_basicauth_object.resources[0].metadata.labels.stf_one_time_upgrade is not defined
 
 - name: Set default Interconnect manifest
   set_fact:


csibbitt:
I'm in the middle of testing this right now.

csibbitt:
# Testing

I noticed this while working on adding auth into the infrared and smoketest scripts. The password changed when I redeployed the operator pod for unrelated reasons.

## Before

With the current 1.5.3 code, the original password was in place and unchanged overnight in a quiesced environment; but changes when I patch the STF object.

```
$ oc get secret default-interconnect-users -o json | jq -r .data.guest | base64 -d; echo
NCewFQQFHWl_osYfL87U

$ oc patch stf default --patch '{"spec":{"alerting":{"alertmanager":{"receivers":{"snmpTraps":{"community": "test1"}}}}}}' --type=merge; sleep 60

$ oc get secret default-interconnect-users -o json | jq -r .data.guest | base64 -d; echo
SzRlEeD6Ap,t:UyGgmLU
```

With the fixed code, the password remains unchanged in the face of both multiple STF changes and the passage of time.

```
$ git log --decorate=short | head -1
commit 5798487aff329b460125877b90f8fd33128f2f9d (HEAD -> csibbitt/fix-qdr-auth-creds-upgrade, origin/csibbitt/fix-qdr-auth-creds-upgrade)

$ oc delete project service-telemetry
project.project.openshift.io "service-telemetry" deleted

$ oc new-project service-telemetry
[...]

$ ansible-playbook --extra-vars __local_build_enabled=true --extra-vars working_branch="$(git rev-parse --abbrev-ref HEAD)" --extra-vars __service_telemetry_snmptraps_enabled=true --extra-vars __service_telemetry_storage_ephemeral_enabled=true --extra-vars __service_telemetry_observability_strategy=use_redhat ./run-ci.yaml
[...]

$ ./validate_deployment.sh
[...]
* [info] CI Build complete. You can now run tests.

$ oc get secret default-interconnect-users -o json | jq -r .data.guest | base64 -d; echo
548ymk-p,NT0Fd:Mhv-w

$ oc patch stf default --patch '{"spec":{"alerting":{"alertmanager":{"receivers":{"snmpTraps":{"community": "test1"}}}}}}' --type=merge; sleep 60
servicetelemetry.infra.watch/default patched

$ oc get secret default-interconnect-users -o json | jq -r .data.guest | base64 -d; echo
548ymk-p,NT0Fd:Mhv-w

$ oc patch stf default --patch '{"spec":{"alerting":{"alertmanager":{"receivers":{"snmpTraps":{"community": "test2"}}}}}}' --type=merge; sleep 600
servicetelemetry.infra.watch/default patched

$ oc get secret default-interconnect-users -o json | jq -r .data.guest | base64 -d; echo
548ymk-p,NT0Fd:Mhv-w
```

vkmc (approved):
Thanks for the detailed explanation of the issue Chris, the fix looks good

leifmadsen:
Just a heads up that based on some additional testing and discussion (only related here to help with verification) that we found the parsing for the plaintext password in the qdrouterd.conf can result in certain generated passwords not working. I ran into this when connecting my OSP to an STF for other reasons, but using the authentication method, still failing.

Our hunch was the `:` in the password causing issues. I changed the password to be significantly simpler, and that resulted in connections being authenticated and setup without issue.

Chris did some investigation (and continues to investigate as of the time I posted this comment) about a comment in the qpid-dispatch code in qd_dispatch_validate_config() `// TODO Add real validation code`.

Our expectation is that we'll need to make sure the passwords being generated don't use any special characters, which is unfortunate, but likely required for this iteration.

roles/servicetelemetry/tasks/component_qdr.yml
```diff
@@ -174,7 +174,7 @@
           stringData:
             guest: "{{ lookup('password', '/dev/null') }}"
       when:
-        - _qdr_basicauth_object.resources[0] is defined and _qdr_basicauth_object[0].metadata.labels.stf_one_time_upgrade is not defined
+        - _qdr_basicauth_object.resources[0] is defined and _qdr_basicauth_object.resources[0].metadata.labels.stf_one_time_upgrade is not defined
```
leifmadsen, line 177: Path looks good here.

```yaml
oc get secret default-interconnect-users -oyaml
apiVersion: v1
data:
  guest: XzlEbExJdlNldU9KbVVwMkdDX2o=
kind: Secret
metadata:
  creationTimestamp: "2023-10-24T14:56:47Z"
  labels:
    application: default-interconnect
    interconnect_cr: default-interconnect
    stf_one_time_upgrade: "1699037433"
  name: default-interconnect-users
  namespace: service-telemetry
```

leifmadsen (approved):

PR #523, opened 2023-11-07
[zuul] Add ocp 4.14  jobs

elfiesmelfie:


.zuul.yaml
```diff
@@ -2,7 +2,8 @@
 - job:
     name: stf-base
     # defined in: https://review.rdoproject.org/cgit/config/tree/zuul.d/_jobs-crc.yaml
-    parent: base-simple-crc
+    #parent: base-simple-crc
+    parent: base-crc
```
elfiesmelfie, line 6: This doesn't work because the custom bundle url is not parsed by the crc/start.yml playbook used by this job

elfiesmelfie:
This is no longer needed, and the OSP 4.14 jobs will be added as part of http://github.com/iunfrawatch/service-telemetry-operator/pull/523

PR #524, opened 2023-11-07
[zuul] Add STO to required repos

elfiesmelfie:
It appears that STO is not included explictly when running jobs from SGO [1]. This will be the case in all the other repos. This change explicitly add it, in case it's not already included by zuul.

[1] https://review.rdoproject.org/zuul/build/edd8f17bfdac4360a94186b46c4cea3f

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -12,6 +12,7 @@
     required-projects:
       - name: openstack-k8s-operators/ci-framework
         override-checkout: main
+      - name: github.com/infrawatch/service-telemetry-operator
       - name: github.com/infrawatch/smart-gateway-operator
       - name: github.com/infrawatch/sg-bridge
       - name: github.com/infrawatch/sg-core


csibbitt (approved):

vyzigold (approved):
lgtm

leifmadsen:
My only concern here (and I'm not sure exactly how the tooling works here, or if this matters for "same-named" branches etc), but the reason it wasn't included was because the expectation was the deployment would be running out of STO repo from the currently checked out branch/code base.

I just want to note it in case this matters, since I would expect it will clone out the `master`(?) branch, which would be behind any working branch. This is fine on anything other than STO, but if changes are landing here in a working branch, will this skip testing anything?

elfiesmelfie:
> My only concern here (and I'm not sure exactly how the tooling works here, or if this matters for "same-named" branches etc), but the reason it wasn't included was because the expectation was the deployment would be running out of STO repo from the currently checked out branch/code base.
> 
> I just want to note it in case this matters, since I would expect it will clone out the `master`(?) branch, which would be behind any working branch. This is fine on anything other than STO, but if changes are landing here in a working branch, will this skip testing anything?

Zuul needs to be told explicitly what to clone (besides the repo being tested).
I tested this on STO, so that was clones because it was the repo under test.
When the jobs run anywhere else (e.g. SGO), the STO repo is not there, because I never told zuul to include it.
Usually, this is hidden in a common base job somewhere.

As for the branches, by default, the branch with the same name as the target (usually master, sometimes stable*) is checked out.
When you use Depends-On, it checks out those changes in the relevant repo.

leifmadsen (approved):

elfiesmelfie:
> OK so just to clarify, basically you tell it the list of repos you need, but the "current" repo would satisfy for that, meaning it'll not checkout a different copy/branch of the same repo, and rather use the local contents for the currently tested branch?
Yes

 
> I noticed that I could use Depends-On across PRs in the same repo, but I don't think I was able to make it work across repositories. Is there a syntax that can be used for that to allow PR dependency resolution across repositories?

Depends-On works across repos e.g. when Jaromir used Depends-On for  http://github.com/infrawatch/sg-core/pulls/123

Note: Depends-On needs to be in the PR description

leifmadsen:
> > OK so just to clarify, basically you tell it the list of repos you need, but the "current" repo would satisfy for that, meaning it'll not checkout a different copy/branch of the same repo, and rather use the local contents for the currently tested branch?
> > Yes
> 
> > I noticed that I could use Depends-On across PRs in the same repo, but I don't think I was able to make it work across repositories. Is there a syntax that can be used for that to allow PR dependency resolution across repositories?
> 
> Depends-On works across repos e.g. when Jaromir used Depends-On for [http://github.com/infrawatch/sg-core/pulls/123](https://github.com/infrawatch/sg-core/pulls/123)
> 
> Note: Depends-On needs to be in the PR description
> 
> Did you mean across GH organisations? You can use Depends-On to reference a change in any repo that zuul knows about (i.e. included in project.yaml in RDO in this case). It **might** be possible to also reference other projects/changes.

OK great! Maybe I was remembering something else that didn't work across repos. I did mean in the same organization at least though :)

PR #525, opened 2023-11-07
QDR Auth in smoketest

csibbitt:
* Added qdr-test as a mock of the OSP-side QDR
* Connection from qdr-test -> default-interconnect is TLS+Auth
* Collectors point at qdr-test instead of default-interconnect directly
* Much more realistic than the existing setup
* Eliminated a substitution in sensubility config

```
OLD: [collectors ===clear+anon===> default-interconnect ===> SG]
NEW: [collectors ===clear+anon===> qdr-test ===TLS+auth===> default-interconnect ===> SG] 
```

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -36,7 +36,7 @@ spec:
           strategy: ephemeral
   transports:
     qdr:
-      auth: none
+      auth: basic
       enabled: true
       deploymentSize: 1
       web:
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -36,7 +36,6 @@ spec:
           strategy: ephemeral
   transports:
     qdr:
-      auth: basic
       enabled: true
       deploymentSize: 1
       web:
--- a/tests/smoketest/collectd-sensubility.conf
+++ b/tests/smoketest/collectd-sensubility.conf
@@ -10,7 +10,7 @@ worker_count=2
 checks={"check-container-health":{"command":"cat /healthcheck.log","handlers":[],"interval":3,"occurrences":3,"refresh":90,"standalone":true}}
 
 [amqp1]
-connection=amqp://default-interconnect.<<NAMESPACE>>.svc:5671
+connection=amqp://qdr-test:5672
 results_channel=sensubility/cloud1-telemetry
 client_name=smoketest.redhat.com
 results_format=smartgateway
--- a/tests/smoketest/minimal-collectd.conf.template
+++ b/tests/smoketest/minimal-collectd.conf.template
@@ -11,8 +11,8 @@ LoadPlugin cpu
 LoadPlugin amqp1
 <Plugin "amqp1">
   <Transport "name">
-    Host "default-interconnect"
-    Port "5671"
+    Host "qdr-test"
+    Port "5672"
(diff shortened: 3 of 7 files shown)


csibbitt:
test

Jenkinsfile
```diff
@@ -36,7 +36,6 @@ spec:
           strategy: ephemeral
   transports:
     qdr:
-      auth: none
```
leifmadsen, line 39: 🥳

leifmadsen (approved):
I saw a failure, so I guess we'll see if that was an anomaly or something to fix, but looking through this it all looks good to me.

Only question is why not just convert to `default-interconnect` with the names here vs changing to qdr-test. The name really doesn't matter, it just struck me as a bit odd.

csibbitt:
Tests are failing with 
```
sed: can't read ./qdr-test.conf.yaml.template: No such file or directory
error: no objects passed to create
error: the path "qdr-test.yaml" does not exist
```

Looks like a simple path problem, I was running locally from within the smoketest dir but I guess the test harness does not. I'll fix it up.

> Only question is why not just convert to `default-interconnect` with the names here vs changing to qdr-test. The name really doesn't matter, it just struck me as a bit odd.

I'm not sure what you're asking here. What, specifically would I "convert" to "default-interconnect"? I'm not renaming anything, I'm adding the (previously missing) OSP-side QDR into the test scenario and connecting the OSP collectors to it.

leifmadsen:
> > Only question is why not just convert to `default-interconnect` with the names here vs changing to qdr-test. The name really doesn't matter, it just struck me as a bit odd.
> 
> I'm not sure what you're asking here. What, specifically would I "convert" to "default-interconnect"? I'm not renaming anything, I'm adding the (previously missing) OSP-side QDR into the test scenario and connecting the OSP collectors to it.

I was just looking at the path changes in the .conf files for sensubility and collectd, but I get it now.

leifmadsen:
> ```
> OLD: [collectors ===clear+anon===> default-interconnect ===> SG]
> NEW: [collectors ===clear+anon===> qdr-test ===TLS+auth===> default-interconnect ===> SG] 
> ```

Thanks, that makes a lot more sense :)

Jenkinsfile
csibbitt: There is an equivalent change pending in stf-v-c MR#103

vkmc: 🥳++

vkmc (approved):
LGTM, thanks Chris!

mgirgisf (approved):
I tried to check the syntax and every thing looks good , but for the template configuration I will trust you

csibbitt:
> I tried to check the syntax and every thing looks good , but for the template configuration I will trust you

For reference, the qdr config template here is a mock of the file produced by the puppet automation[1] and is ~largely the same. For comparison, here is a full-rendered version straight from my OSP17.1 deployment[2].

[1] https://github.com/openstack/puppet-qdr/blob/stable/wallaby/templates/qdrouterd.conf.erb
[2] https://paste.opendev.org/show/bhjp9Cnn8JVvtCfXhLv6/

PR #526, opened 2023-11-08
Mgirgisf/stf 1580/fix log commands

mgirgisf:


mgirgisf:
test

build/stf-run-ci/tasks/main.yml
```diff
@@ -221,6 +221,5 @@
     register: validate_deployment
 
   - name: Show the result of the validate_deployment script
-    ansible.builtin.shell:
-      cmd: |
-        cat {{ logfile_dir }}/validate_deployment.log
+    ansible.builtin.debug:
+      msg: "{{lookup('ansible.builtin.file', '{{ logfile_dir }}/validate_deployment.log') }}"
```
elfiesmelfie, line 225: This is causing the failures in the zuul jobs. I'll explain why and propose an alternative.

Why?
The lookup plugins are run on the same host that you run ansible-playbook from; i.e. your local machine, not the target host. This is a very common mistake: often ``localhost`` is the target host, so using lookup plugins has the intended behaviour.
In this case, ``lookup('file', ...)`` is trying to find a the logfile locally, rather than on the host it was generated on.

What's the alternative?
I have three suggestions:
* Use the slurp module.
    https://docs.ansible.com/ansible/latest/collections/ansible/builtin/slurp_module.html
    This involves two tasks to replace the existing shell task (L224, original code)   
  ```
    # note: you need to fix the style and the var names, etc; this is a shorthand representation of the suggested changes
    - ansible.builtin.slurp:
        ## parameters here to get the contents of the file
       register: log_output
        
     - ansible.builtin.debug:
           var: log_output.contents | b64decode
   ```

* Continue to use cat. Optionally, with a debug task following to show the output.
    This would be the easiest option:
        * Add ``register: <var>`` to the shell task (L224, original code)
        * Add a debug task that shows ``<var>.stdout``
        
* Use ``-vvv`` when running the ``ansible-playbook ....`` command
    I suspect that your reason for changing this task is because the output of the shell task isn't shown with the default verbosity level

build/stf-run-ci/tasks/preflight_checks.yml
```diff
@@ -19,9 +19,13 @@
       ansible.builtin.command:
         cmd: |
           oc describe csv $(oc get csv | grep "service-telemetry-operator" | awk '{print $1}') > {{ logfile_dir }}/oc_get_csv_sto.log 2>&1
-          cat {{ logfile_dir }}
-
+          cat {{ logfile_dir }}/oc_get_csv_sto.log
```
elfiesmelfie, line 22: I made a big oopsie here. Thanks for fixing it

build/stf-run-ci/tasks/preflight_checks.yml
```diff
@@ -19,9 +19,13 @@
...
+          
     - name: "Show fail message if CSV isn't Succeeded after the alotted time"
       ansible.builtin.fail:
         msg: "Service Telemetry Operator CSV not Succeeded after 10 minutes. Check {{ logfile_dir }}/oc_get_csv_sto.log for more information"
       when: output.rc != 0
+
+    - name: "Show service-telemetry-operator CSV information"
```
elfiesmelfie, line 29: Please refer to my earlier comment, which applies here too.

This task needs to be moved above the ``fail`` task, otherwise, it is unreachable.

This is mutually exclusive with L22;
Either fix the ``cat`` in L22, or add the new task to show the contents.

build/run-ci.yaml
```diff
@@ -7,9 +7,14 @@
   - name: Run the STF CI system
     import_role:
       name: stf-run-ci
-          
+
+  - name: Create Log directory
+    file:
+      path: build/working/logs
```
elfiesmelfie, line 13: Please replace this with an absolute path.

I would suggest "{{ playbook_dir }}/working/logs", which is "path/to/STO/build/working/logs"

build/run-ci.yaml
```diff
@@ -7,9 +7,14 @@
...
+      state: directory
+
   - name: Collect the logs
     import_role:
       name: stf-collect-logs
     vars:
-      logfile_dir: "{{ playbook_dir }}/"
+      logfile_dir: "build/working/logs/"
```
elfiesmelfie, line 20: Same as above re: absolute dir

build/stf-collect-logs/tasks/main.yml
```diff
@@ -1,35 +1,23 @@
...
+- name: "Get resources logs [ Builds, Subscriptions, Image, Imagestreams, Pods ]"
+  ansible.builtin.shell: |
+    for resource in {{ resource_types|join(' ') }}; do
+      log_file="{{ logfile_dir }}/post_oc_get_$resource.log"
+      echo "*** [INFO] Showing oc get '$resource'" > "$log_file" 2>&1
+      oc -n {{ namespace }} get "$resource" >> "$log_file" 2>&1
+      echo "[INFO] oc get '$resource' -oyaml" >> "$log_file" 2>&1
+      oc -n {{ namespace }} get "$resource" -oyaml >> "$log_file" 2>&1
+    done
+  delay: 10
```
elfiesmelfie, line 11: This is not required unless ``retries`` is also set.

https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_loops.html#retrying-a-task-until-a-condition-is-met

build/stf-collect-logs/tasks/main.yml
```diff
@@ -1,35 +1,23 @@
...
   ignore_errors: true
   changed_when: false
 
-- name: "Get subscription details"
+- name: "Get Additional Information  details"
   ansible.builtin.shell:
     cmd: |
-      oc -n {{ namespace }} get subscriptions > {{ logfile_dir }}/post_oc_get_subscriptions.log 2>&1
       oc -n {{ namespace }} describe subscription service-telemetry-operator >> {{ logfile_dir }}/post_oc_get_subscriptions.log 2>&1
```
elfiesmelfie, line 18: While you're here....
The contents of this file would be clearer if it was called "post_oc_describe_subscription_STO.log", which would match the format of the names of the other log files.

It is currently called ``post_oc_get_subscriptions.log`` because the first command was ``oc get subscriptions``

elfiesmelfie (changes requested):

mgirgisf:
test

mgirgisf:
test

mgirgisf:
test

build/stf-run-ci/tasks/preflight_checks.yml
```diff
@@ -19,9 +19,16 @@
...
+      ansible.builtin.debug:
+        var: csv_sto.stdout
+            
     - name: "Show fail message if CSV isn't Succeeded after the alotted time"
       ansible.builtin.fail:
         msg: "Service Telemetry Operator CSV not Succeeded after 10 minutes. Check {{ logfile_dir }}/oc_get_csv_sto.log for more information"
       when: output.rc != 0
+
+
```
elfiesmelfie, line 34:
```suggestion
```

elfiesmelfie (approved):
Looks good; approved!

I'll sync with master and make sure Jenkins runs, but otherwise, this looks good to go!

build/stf-collect-logs/tasks/main.yml
```diff
@@ -50,14 +39,6 @@
   retries: 3
```
elfiesmelfie, line 39:
```suggestion
  retries: 3
  ignore_errors: true
```

elfiesmelfie:
I updated the target branch for this PR, so that it picks up a change I made that is blocking it. Once PR#529 merges, this can also merge.

PR #527, opened 2023-11-08
QDR Auth in smoketest (#525)

csibbitt:
* QDR Auth in smoketest

* Added qdr-test as a mock of the OSP-side QDR
* Connection from qdr-test -> default-interconnect is TLS+Auth
* Collectors point at qdr-test instead of default-interconnect directly
* Much more realistic than the existing setup
* Eliminated a substitution in sensubility config
* Used default QDR basic auth in Jenkinsfile

(cherry picked from commit 37b6f035d6ff44a39598aacf812b3b893bafda7e)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -36,7 +36,6 @@ spec:
           strategy: ephemeral
   transports:
     qdr:
-      auth: none
       enabled: true
       deploymentSize: 1
       web:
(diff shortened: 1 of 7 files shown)


csibbitt:
@leifmadsen Just an FYI that this need to merge here, but no need to re-tag anything for future 1.5.3 builds since this doesn't affect deliverables.

leifmadsen (approved):

PR #529, opened 2023-11-08
[stf-collect-logs] Add ignore_errors to task

elfiesmelfie:
The "Question the deployment" task didn't have
ignore_errors: true set, so when the task fails, the play is finished. This means that we don't get to the
"copy logs" task and can't see the job logs in zuul.

ignore_errors is set to true to be consistent with other tasks


leifmadsen (approved):

elfiesmelfie:
recheck

csibbitt (approved):

elfiesmelfie:
recheck

The failure in the 4.12 local_build job looks random, since the same task ran successfully in the 4.12-nightly_bundles job

elfiesmelfie:
retry

elfiesmelfie:
recheck

PR #537, opened 2023-11-17
[zuul] Add a new job that will compare crc/simple-start with crc/start

elfiesmelfie:


elfiesmelfie:
This is irrelevant since the jobs are moving to extracted crc images in [http://github.com/infrawatch/service-telemetry-operator/pull/523](https://github.com/iunfrawatch/service-telemetry-operator/pull/531)

PR #538, opened 2023-11-17
Adjust Operator dependency version requirements

leifmadsen:
Adjust the operator package dependency requirements to align to known
required versions. Primarily reduce the version of
openshift-cert-manager from 1.10 to 1.7 in order to support the
tech-preview channel which was previously used.

Lowering the version requirement allows for the
openshift-cert-manager-operator installed previously to be used during
the STF 1.5.2 to 1.5.3 update, removing the update from being blocked.

Related: STF-1636

### DIFF
--- a/deploy/olm-catalog/service-telemetry-operator/metadata/properties.yaml
+++ b/deploy/olm-catalog/service-telemetry-operator/metadata/properties.yaml
@@ -27,7 +27,7 @@ properties:
           - failureMessage: Package openshift-cert-manager-operator is needed for AMQ Interconnect setup
             package:
               packageName: openshift-cert-manager-operator
-              versionRange: '>=1.10.0'
+              versionRange: '>=1.7.0'
   - type: olm.constraint
     value:
       failureMessage: Require Prometheus backend for data storage of metrics for Service Telemetry Framework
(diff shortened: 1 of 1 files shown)


deploy/olm-catalog/service-telemetry-operator/metadata/properties.yaml
```diff
@@ -38,7 +38,7 @@ properties:
               versionRange: '>=0.56.0'
           - package:
               packageName: observability-operator
-              versionRange: '>=0.0.1'
+              versionRange: '>=0.0.25'
           - package:
               packageName: cluster-observability-operator
-              versionRange: '>=0.0.1'
+              versionRange: '>=0.1.0'
```
leifmadsen, line 44: I also aligned these versions to known released versions of the operators.

vkmc (approved):
Code looks good. The failure in the OCP 4.13 CI doesn't seem to be related to this change, so I think it's good to go. Thanks Leif!

vkmc:
recheck

leifmadsen:
I'm going to merge this for easier testing and because there is enough CI to show it doesn't fundamentally break anything. If we find this is _really_ wrong we'll revert it prior to backporting.

PR #539, opened 2023-11-20
Clean up stf-run-ci for OCP 4.12 minimum version

leifmadsen:
Update the stf-run-ci base setup to no longer need testing against OCP
4.10 and earlier, meaning we can rely on a single workflow for
installation. Also update the deployment to use
cluster-observability-operator via the redhat-operators CatalogSource
for installation via use_redhat and use_hybrid strategies.

### DIFF
--- a/build/stf-run-ci/tasks/setup_base.yml
+++ b/build/stf-run-ci/tasks/setup_base.yml
@@ -28,47 +28,7 @@
         targetNamespaces:
         - "{{ namespace }}"
 
-# deploy cert-manager from tech-preview when using versions of OCP < 4.12
-- when: not __deploy_from_index_enabled | bool and ocp_ver.stdout is version ('4.12', '<')
-  block:
-    - name: Create openshift-cert-manager-operator namespace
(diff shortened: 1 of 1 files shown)


build/stf-run-ci/tasks/setup_base.yml
```diff
@@ -129,13 +87,13 @@
       metadata:
         labels:
           operators.coreos.com/observability-operator.openshift-operators: ""
-        name: observability-operator
+        name: cluster-observability-operator
         namespace: openshift-operators
       spec:
-        channel: stable
+        channel: development
```
csibbitt, line 93: Is this right? If COO is released now, shouldn't this say stable? (I haven't installed it yet)

leifmadsen: Heh oddly enough, the COO channel is released under `development`. I advocated for the channel to be `tech-preview` but it was too late to make that kind of change.

csibbitt (approved):

PR #540, opened 2023-11-21
create pvs in crc

mgirgisf:
Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/642

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -12,6 +12,8 @@
     required-projects:
       - name: openstack-k8s-operators/ci-framework
         override-checkout: main
+      - name: github.com/openstack-k8s-operators/install_yamls
+        override-checkout: main
       - name: github.com/infrawatch/service-telemetry-operator
       - name: github.com/infrawatch/smart-gateway-operator
       - name: github.com/infrawatch/sg-bridge
@@ -99,7 +101,7 @@
       jobs:
         - pvs_crc
         #- stf-crc-ocp_412-nightly_bundles
-        #- stf-crc-ocp_412-local_build
+        - stf-crc-ocp_412-local_build
         #- stf-crc-ocp_413-nightly_bundles
         #- stf-crc-ocp_413-local_build
 
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -98,10 +98,10 @@
     github-check:
       jobs:
         - pvs_crc
-        - stf-crc-ocp_412-nightly_bundles
-        - stf-crc-ocp_412-local_build
-        - stf-crc-ocp_413-nightly_bundles
-        - stf-crc-ocp_413-local_build
+        #- stf-crc-ocp_412-nightly_bundles
+        #- stf-crc-ocp_412-local_build
+        #- stf-crc-ocp_413-nightly_bundles
+        #- stf-crc-ocp_413-local_build
 
 - project:
     name: infrawatch/service-telemetry-operator
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -98,6 +98,7 @@
       STF CRC jobs that build and deploy STF
     github-check:
       jobs:
+        - pvs_crc
         - stf-crc-ocp_412-nightly_bundles
         - stf-crc-ocp_412-local_build
         - stf-crc-ocp_413-nightly_bundles
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -34,7 +34,6 @@
 - job:
     name: pvs_crc
     parent: base-simple-crc
-    abstract: true
     description: |
       Create Pvs in crc Cluster
     run:
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -32,6 +32,17 @@
       crc_parameters: '--memory 16000 --disk-size 80 --cpus 6'  # Increase from 14336
 
 - job:
+    name: pvs_crc
+    parent: base-simple-crc
+    abstract: true
+    description: |
+      Create Pvs in crc Cluster
+    run:
+      - ci/define_pvs.yml
+    vars:
+      scenario: "nightly_bundles"
+
+- job:
     name: stf-crc-nightly_bundles
     parent: stf-base
     abstract: true
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -33,7 +33,7 @@
 
 - job:
     name: pvs_crc
-    parent: base-simple-crc
+    parent: stf-base
     description: |
       Create Pvs in crc Cluster
     run:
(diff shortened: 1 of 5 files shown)


.zuul.yaml
```diff
@@ -31,6 +31,16 @@
...
 
+- job:
+    name: pvs_crc
+    parent: base-simple-crc
+    description: |
+      Create Pvs in crc Cluster
+    run:
+      - ci/define_pvs.yml
+    vars:
+      scenario: "nightly_bundles"
```
elfiesmelfie, line 42: FYI, this won't do anything here, since ``scenario`` is a parameter for the ci/prepare.yml, ci/deploy_stf.yml, ci/test_stf.yml and ci/post-collect_logs.yml playbooks.

You will need these lines if you use stf-base as the parent job.

.zuul.yaml
```diff
@@ -87,6 +97,7 @@
       STF CRC jobs that build and deploy STF
     github-check:
       jobs:
+        - pvs_crc
         - stf-crc-ocp_412-nightly_bundles
         - stf-crc-ocp_412-local_build
         - stf-crc-ocp_413-nightly_bundles
```
elfiesmelfie, line 103: These jobs take quite a while to run.

For testing, I recommend that you comment them out so that you can get results faster from the ``pvs_crc`` job.

```suggestion
        #- stf-crc-ocp_412-nightly_bundles
        #- stf-crc-ocp_412-local_build
        #- stf-crc-ocp_413-nightly_bundles
```

ci/define_pvs.yml
```diff
@@ -0,0 +1,49 @@
...
+    - name: Check if make command exists
+      ansible.builtin.command: "command -v make"
+      register: make_check_result
+      ignore_errors: true
+
+    - name: Install make if not present
+      ansible.builtin.package:
+        name: make
+        state: present
+      when: make_check_result.rc != 0
```
elfiesmelfie, line 32: The package module should only install make if it doesn't already exist.
So this check and when are not needed

```suggestion
    - name: Install make if not present
      ansible.builtin.package:
        name: make
        state: present
```

ci/define_pvs.yml
```diff
@@ -0,0 +1,49 @@
...
+    - name: Print make result
+      ansible.builtin.debug:
+        var: make_output
+
+    - name: Confirm  Pvs exist
+      ansible.builtin.shell: 
+        cmd: |
+          oc get pv | grep "storage..-crc"| wc -l
+      register: cmd_output
+      failed_when: "cmd_output.stdout|int !=12"
```
elfiesmelfie, line 48: I'd like to see some output here as well e.g. a debug.
You can use the `ansible.builtin.fail` module to check `cmd_output.stdout_lines | length`

Since we don't control the creation of the PVs, I would revise this test, since the number of lines may change without changing the PVs, for example.

elfiesmelfie (changes requested):

ci/define_pvs.yml
elfiesmelfie: IINM, we want to be able to use the PVs in the ``{{ namespace }}`` namespace/project [1], so there should be a check for that.

[1] https://github.com/infrawatch/service-telemetry-operator/blob/77dea87dec0656c48d0c7e0b9a09d0fe826425aa/ci/vars-zuul-common.yml#L2

ci/define_pvs.yml
mgirgisf: I forgot that, thanks emma 😅

ci/define_pvs.yml
mgirgisf: yes , this is not the final test , I believe I still need to create pvc to be used directly in the pods this create only the pvs , but wanted to check how the zuul will run it , so I created it as a draft pull request but this is not done yet

ci/define_pvs.yml
```diff
@@ -0,0 +1,54 @@
...
+  vars:
+   base_dir: "{{ ansible_user_dir }}"
+   namespace: "service-telemetry" 
+  tasks:
+    - name: "Log into the cluster"
+      ansible.builtin.import_role:
+        name: rhol_crc
+        tasks_from: add_crc_creds.yml
+
+    - name: "Check if the install_yamls already exists"
```
elfiesmelfie, line 13: Best practice with zuul is to never clone anything in your jobs.

You can configure zuul to clone the required repos as part of the job setup.
This is done in .zuul.yaml in the `required-projects` section of the job (in this case, the parent `stf-base`)
Once a project is cloned, it is available at `{{ ansible_env.HOME }}/{{ zuul.projects[<fully qualified project name>].src_dir }}`
You can see examples of this in `ci/vars-zuul-common.yml`.

The projects you can include are listed in rdo/config repo [2].
In this case, the relevant section is in [3] and the fully-qualified project name is `github.com/openstack-k8s-operators/install_yamls`, which will also be used in `.zuul.yaml` to tell zuul to include that project.

[1] https://review.rdoproject.org/zuul/build/62079c915ab846b99f1cffc83a4b97f9/log/zuul-info/inventory.yaml
[2] https://review.rdoproject.org/cgit/config/tree/zuul/rdo.yaml
[3] https://review.rdoproject.org/cgit/config/tree/zuul/rdo.yaml#n1231

elfiesmelfie: I can't comment on an unchanged line, the required project should be added after line 14 in `.zuul.yaml`.

By default, zuul will try to check out the same branch name (i.e. master), but install_yamls does not have this branch, so you'll also need the `override-checkout` as well as the `name` for `install_yamls`.

https://zuul-ci.org/docs/zuul/latest/config/job.html#attr-job.required-projects

ci/vars-zuul-common.yml
```diff
@@ -9,3 +9,5 @@ sgo_dir: "{{ ansible_env.HOME }}/{{ zuul.projects['github.com/infrawatch/smart-g
 sg_core_dir: "{{ ansible_env.HOME }}/{{ zuul.projects['github.com/infrawatch/sg-core'].src_dir }}"
 sg_bridge_dir: "{{ ansible_env.HOME }}/{{ zuul.projects['github.com/infrawatch/sg-bridge'].src_dir }}"
 prometheus_webhook_snmp_dir: "{{ ansible_env.HOME }}/{{ zuul.projects['github.com/infrawatch/prometheus-webhook-snmp'].src_dir }}"
+install_yamls_dir: "{{ ansible_env.HOME }}/{{ zuul.projects['github.com/openstack-k8s-operators/install_yamls'].src_dir }}"
+__service_telemetry_storage_persistent_storage_class: "crc-storage"
```
elfiesmelfie, line 13: Typo. In vars-local_build, you used an underscore.

Both should be the same.

elfiesmelfie (changes requested):

PR #541, opened 2023-11-22
Only require Interconnect and Smart Gateway

leifmadsen:
Update the dependency management within Service Telemetry Operator to
only require AMQ Interconnect and Smart Gateway Operator, which is
enough to deploy STF with observabilityStrategy: none. Other Operators
can be installed in order to satisfy data storage of telemetry and
events.

Installation of cert-manager is also required, but needs to be
pre-installed similar to Cluster Observability Operator, either as a
cluster-scoped operator with the tech-preview channel, or a single time
on the cluster as a namespace scoped operator, which is how the
stable-v1 channel installs.

Documentation will be updated to adjust for this change.

Related: STF-1636

### DIFF
--- a/build/stf-run-ci/tasks/create_catalog.yml
+++ b/build/stf-run-ci/tasks/create_catalog.yml
@@ -170,4 +170,4 @@
           securityContextConfig: legacy
         updateStrategy:
           registryPoll:
-            interval: 1m
+            interval: 5m
--- a/build/stf-run-ci/tasks/pre-clean.yml
+++ b/build/stf-run-ci/tasks/pre-clean.yml
@@ -134,6 +134,7 @@
   loop:
     - openshift-cert-manager-operator
     - cert-manager-operator
+    - cert-manager
 
 - name: Remove Elasticsearch
   ignore_errors: true
--- a/build/stf-run-ci/tasks/pre-clean.yml
+++ b/build/stf-run-ci/tasks/pre-clean.yml
@@ -122,7 +122,6 @@
         name: smart-gateway-operator-catalog
         namespace: "{{ namespace }}"
 
-# Remove the cert manager since we install it as part of the CI/documented pre-install process
 - name: Remove openshift-cert-manager-operator namespace
   kubernetes.core.k8s:
     state: absent
@@ -131,7 +130,10 @@
       apiVersion: project.openshift.io/v1
       kind: Project
       metadata:
-        name: openshift-cert-manager-operator
+        name: "{{ item }}"
+  loop:
+    - openshift-cert-manager-operator
+    - cert-manager-operator
 
 - name: Remove Elasticsearch
   ignore_errors: true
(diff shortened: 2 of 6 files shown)


csibbitt (approved):
This is needed for 1.5.3 as an outcome of the testing? We're pretty darn sure this is necessary? I approve only hesitantly, because this is not super cool. We've been trying to get OLM dependencies to work properly since before STF 1.0 and I've really been rooting for this to be the time!

vkmc (approved):
Not a fan of this change either, but I believe is the right approach. We are still reducing the number of steps the end user needs to follow (by adding some dependencies with OLM), so I think it's a partial win. We can re-iterate on this for the next release, when most end users are consuming newer versions of dependencies. Thanks Leif!

leifmadsen:
I'm probably the most disappointed since I really thought it was going to work this time :) At least the documentation wasn't too crazy to update, and might have almost made things a bit simpler than where it looked like it was going.

I suspect this will be the best we get in terms of dependency management, but there are probably some checks we can add to the Operator itself to make things a bit nicer. At least things are starting to converge on the supported stack now.

vkmc:
Manually tested this change performing an update from STF 1.5.2 to STF 1.5.3. Everything is working as expected.

leifmadsen:
> While this is passing CI, I'm going to do another check of the stf-run-ci workflow to make sure there aren't any other changes I should be making. At the very least, I think I'm going to change cert-manager to running in the default namespace recommended by their documentation, which would also align it to how we're going to install from our own documentation. It appears cert-manager doesn't need to be in the same namespace as STF, so I'll just leave it over where they recommend.

I performed an stf-run-ci walkthrough based on these changes and documentation changes. They are targeting _this_ branch and are available here: https://github.com/infrawatch/service-telemetry-operator/pull/542

Once reviewed and tested, the CI changes will land here, and after a file round of verified testing, everything will land into master in preparation for backport.

elfiesmelfie (approved):
It looks reasonable to me. It appears you're removing the automatic installation of the (former) reqs and then adding some tasks into ci to ensure the required set-up is done.

leifmadsen:
> It looks reasonable to me. It appears you're removing the automatic installation of the (former) reqs and then adding some tasks into ci to ensure the required set-up is done.

Yep you got it. Basically now that cert-manager needs to be pre-installed, I put that back in to the CI, and made sure everything aligned to what was documented (namespaces, etc).

leifmadsen:
recheck

leifmadsen:
recheck

leifmadsen:
At this point, I think I'm just going to merge this. It seems all the problems in the Zuul testing seem to be intermittent problems with CRC deployments. My local test passed without issue, the Jenkins deployment on 4.14 passed without issue, and running the Zuul checks 3 times now results in things getting closer to passing without any changes.

I have one more extra change to make here and will push it, in the hopes I'll hit the magic pass window.

PR #495, opened 2023-09-28
[zuul] Add job to build locally and do an index-based deployment

elfiesmelfie:
Depends-On: https://github.com/infrawatch/service-telemetry-operator/pull/497
Depends-On: https://github.com/infrawatch/service-telemetry-operator/pull/488

PR #542, opened 2023-11-23
Perform CI update to match docs install changes

leifmadsen:
Update the stf-run-ci scripting to match the documented installation
procedures which landed in
https://github.com/infrawatch/documentation/pull/513. These changes are
also reflected in #541.

### DIFF
--- a/build/stf-run-ci/tasks/create_catalog.yml
+++ b/build/stf-run-ci/tasks/create_catalog.yml
@@ -170,4 +170,4 @@
           securityContextConfig: legacy
         updateStrategy:
           registryPoll:
-            interval: 1m
+            interval: 5m
--- a/build/stf-run-ci/tasks/pre-clean.yml
+++ b/build/stf-run-ci/tasks/pre-clean.yml
@@ -122,7 +122,6 @@
         name: smart-gateway-operator-catalog
         namespace: "{{ namespace }}"
 
-# Remove the cert manager since we install it as part of the CI/documented pre-install process
 - name: Remove openshift-cert-manager-operator namespace
   kubernetes.core.k8s:
     state: absent
@@ -131,7 +130,10 @@
       apiVersion: project.openshift.io/v1
       kind: Project
       metadata:
-        name: openshift-cert-manager-operator
+        name: "{{ item }}"
+  loop:
+    - openshift-cert-manager-operator
+    - cert-manager-operator
 
 - name: Remove Elasticsearch
   ignore_errors: true
--- a/build/stf-run-ci/tasks/setup_base.yml
+++ b/build/stf-run-ci/tasks/setup_base.yml
@@ -82,15 +82,16 @@
 
   - name: Create OperatorGroup for cert-manager for Red hat OpenShift
     kubernetes.core.k8s:
-      apiVersion: operators.coreos.com/v1
-      kind: OperatorGroup
-      metadata:
-        name: cert-manager-operator
-        namespace: cert-manager-operator
-      spec:
-        targetNamespaces:
-        - cert-manager-operator
-        upgradeStrategy: Default
+      definition:
+        apiVersion: operators.coreos.com/v1
+        kind: OperatorGroup
+        metadata:
+          name: cert-manager-operator
+          namespace: cert-manager-operator
+        spec:
+          targetNamespaces:
+          - cert-manager-operator
+          upgradeStrategy: Default
 
   - name: Subscribe to cert-manager for Red Hat OpenShift Operator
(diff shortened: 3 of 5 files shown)


build/stf-run-ci/tasks/create_catalog.yml
```diff
@@ -170,4 +170,4 @@
           securityContextConfig: legacy
         updateStrategy:
           registryPoll:
-            interval: 1m
+            interval: 5m
```
leifmadsen, line 173: I find that the index image can get into a loop with a really short interval. In testing I found setting this to 5 minutes tended to avoid that constant update loop.

build/stf-run-ci/tasks/pre-clean.yml
```diff
@@ -131,7 +130,10 @@
       apiVersion: project.openshift.io/v1
       kind: Project
       metadata:
-        name: openshift-cert-manager-operator
+        name: "{{ item }}"
+  loop:
+    - openshift-cert-manager-operator
+    - cert-manager-operator
```
leifmadsen, line 136: tech-preview channel vs stable-v1 channel uses different namespaces, so this covers both instances.

build/stf-run-ci/tasks/setup_stf.yml
```diff
@@ -9,7 +9,7 @@
         namespace: openshift-marketplace
       spec:
         displayName: InfraWatch Operators
-        image: quay.io/infrawatch-operators/infrawatch-catalog:unstable
+        image: quay.io/infrawatch-operators/infrawatch-catalog:nightly
```
leifmadsen, line 12: Unstable is not a thing, so I fixed this to match a tag that gets updated regularly. Not sure how often this path is used, but it would not have worked previously.

build/stf-run-ci/tasks/setup_stf.yml
```diff
@@ -49,6 +49,14 @@
         sourceNamespace: openshift-marketplace
   when: service_telemetry_operator_subscription_manifest is not defined
 
+# enable catalogsource
```
leifmadsen, line 52: Seemed weird to subscribe before creating the catalogsource, so I just moved this up in the task list.

build/stf-run-ci/tasks/setup_stf.yml
elfiesmelfie: I had no idea that this was here. 

Is this related to index deployment?

If so, it doesn't currently run in CI but the CI job is a WIP at #495

build/stf-run-ci/tasks/setup_base.yml
```diff
@@ -98,27 +49,107 @@
...
+        apiVersion: project.openshift.io/v1
+        kind: Project
+        metadata:
+          name: cert-manager-operator
+        spec:
+          finalizers:
+          - kubernetes
+
+  - name: Create OperatorGroup for cert-manager for Red hat OpenShift
+    kubernetes.core.k8s:
```
elfiesmelfie, line 84: You need to update this one to use the `definition` parameter (see the other k8s tasks)

build/stf-run-ci/tasks/setup_base.yml
```diff
@@ -98,27 +49,107 @@
...
+      apiVersion: operators.coreos.com/v1
+      kind: OperatorGroup
+      metadata:
+        name: cert-manager-operator
+        namespace: cert-manager-operator
+      spec:
+        targetNamespaces:
+        - cert-manager-operator
+        upgradeStrategy: Default
```
elfiesmelfie, line 93:
```suggestion
    kubernetes.core.k8s:
      definition:
        apiVersion: operators.coreos.com/v1
        kind: OperatorGroup
        metadata:
          name: cert-manager-operator
          namespace: cert-manager-operator
        spec:
          targetNamespaces:
          - cert-manager-operator
          upgradeStrategy: Default
```

elfiesmelfie (changes requested):

build/stf-run-ci/tasks/setup_stf.yml
leifmadsen: I'm just cleaning it up because I was here, but I think it's a legacy interface and I wouldn't necessarily expect CI to test for it.

build/stf-run-ci/tasks/setup_base.yml
leifmadsen: Absolutely right. Thanks for catching that!

leifmadsen:
Yay it passes! :)

PR #545, opened 2023-11-30
[stf-run-ci] Explicitly check the validate_daployment was successful

elfiesmelfie:
In [1], the validate_deployment step is successful, despite the deployment not being successful.
This causes the job to timeout because the following steps continue to run despite an invalid state.

To get the expected behaviour, the output should be checked for a string indicating success.
i.e. * [info] CI Build complete. You can now run tests. [2] shows the output for a successful run.

[1] https://review.rdoproject.org/zuul/build/245ae63e41884dc09353d938ec9058d7/console#5/0/144/controller
[2] https://review.rdoproject.org/zuul/build/802432b23da24649b818985b7b1633bb/console#5/0/82/controller

### DIFF
--- a/build/stf-run-ci/tasks/main.yml
+++ b/build/stf-run-ci/tasks/main.yml
@@ -220,7 +220,8 @@
     args:
       executable: /bin/bash
     register: validate_deployment
+    failed_when: validate_deployment.stdout_lines[-1] != "* [info] CI Build complete. You can now run tests."
 
   - name: Show the result of the validate_deployment script
     ansible.builtin.debug:
-      var: validate_deployment.stdout
+      var: validate_deployment.stdout_lines[-1]


csibbitt (approved):
It think it might be better to have validate_deployment.sh exit non-zero if the deployment does not validate so that we don't depend on that specific string, but either solutions is better than the status quo.

elfiesmelfie:
> It think it might be better to have validate_deployment.sh exit non-zero if the deployment does not validate so that we don't depend on that specific string, but either solutions is better than the status quo.

The execution of the script depends on timeouts, there's no explicit check-and-emit-error piece.

The only other option I can think of immediately is something like:
```
# validate_deployment.sh
export validation_result=FAILED
<do the test steps>
export validation_result=PASSED
```

```
- shell:
    cmd: |
      ./validate_deployment.sh
      echo $validation_result
    failed_when: output != PASSED
```

Alternatively, check whether $validation_result is set or not
```
# validate_deployment.sh
unset validation_passed
<do test>
export validation_passed
```

``` 
shell: ./validate_deployment.sh; echo $validation_passed
register: output    
failed_when: output.stdout | length == 0

**Comment by leifmadsen** [comment]
I did a deployment from this branch, and it didn't fail at least :)

```json
TASK [stf-run-ci : Validate system is operational] ************************************************************************************************************************************************************
changed: [localhost]

TASK [stf-run-ci : Show the result of the validate_deployment script] *****************************************************************************************************************************************
ok: [localhost] => {
    "validate_deployment.stdout_lines[-1]": "* [info] CI Build complete. You can now run tests."
}

```

leifmadsen (approved):

elfiesmelfie:
recheck

elfiesmelfie:
recheck

PR #549, opened 2023-12-04
Auth to prometheus using token instead of basicauth

csibbitt:


roles/servicetelemetry/tasks/component_prometheus_reader.yml
```diff
@@ -0,0 +1,54 @@
+- name: Create ServiceAccount/stf-prometheus-reader
+  k8s:
+    definition:
```
leifmadsen, line 3: You could add the following line to these `k8s` module calls to allow STO to also remove the extra ServiceAccount, Role, and RoleBinding if the metrics storage is ever disabled (could happen in the future?):

```yaml
state: '{{ "present" if servicetelemetry_vars.backends.metrics.prometheus.enabled else "absent" }}'
```

roles/servicetelemetry/templates/manifest_prometheus.j2
```diff
@@ -42,10 +41,11 @@ spec:
     - -tls-cert=/etc/tls/private/tls.crt
     - -tls-key=/etc/tls/private/tls.key
     - -upstream=http://localhost:9090/
-    - -htpasswd-file=/etc/proxy/htpasswd/auth
     - -cookie-secret-file=/etc/proxy/secrets/session_secret
     - -openshift-service-account=prometheus-stf
-    - '-openshift-sar={"resource": "namespaces", "verb": "get"}'
+    - '-openshift-sar={"namespace":"{{ ansible_operator_meta.namespace }}","resource": "prometheus", "group":"{{ prometheus_operator_api_string | replace("/v1","") }}", "verb":"get"}'
```
csibbitt, line 46: I'd like to change the other SARs in the system to be more specific like this, too. The existing one matched what we found in openshift-monitoring when we were headed in that direction, and is a rough approximation of "admin user" without referencing a specific admin group. Using more specific check like this makes more sense to me. This is probably release-note worthy.

csibbitt:
@elfiesmelfie - AFAICT the failures in zuul are because these smoketest updates are not compatible with the existing nightly builds. If I understand correctly, I don't think it's possible to pass this gate if we are making changes that require modifications to the testing like this. Any advice or insight to add here?

leifmadsen:
> @elfiesmelfie - AFAICT the failures in zuul are because these smoketest updates are not compatible with the existing nightly builds. If I understand correctly, I don't think it's possible to pass this gate if we are making changes that require modifications to the testing like this. Any advice or insight to add here?

Oh that makes some sense, since obviously the changes need to land in order for them to arrive in the nightly builds.

I know we had some checks for Jenkins that would skip certain bits of testing when changes were only for a certain directory (which wouldn't have testing coverage, like a docs update). Do we have that kind of control with zuul-based testing? (or an auto-exit of the job with a positive return code?)

csibbitt:
More testing awkwardness: https://github.com/infrawatch/functional-tests/pull/62#issuecomment-1841116958

elfiesmelfie:
The stf-crc-ocp_413-local_build failure is due to this error: error: can't lookup images: the server is currently unable to handle the request (post imagestreamimports.image.openshift.io)

Not related to your change

leifmadsen:
Started looking at this, but really not feeling great. Going to review this later or tomorrow with a fresh brain.

elfiesmelfie:
> I know we had some checks for Jenkins that would skip certain bits of testing when changes were only for a certain directory (which wouldn't have testing coverage, like a docs update). Do we have that kind of control with zuul-based testing? (or an auto-exit of the job with a positive return code?)

Yes, we can add the files or the irrelevant-files attributes to the job.

https://zuul-ci.org/docs/zuul/latest/config/job.html#attr-job.files
https://zuul-ci.org/docs/zuul/latest/config/job.html#attr-job.irrelevant-files

csibbitt:
recheck

csibbitt:
recheck

roles/servicetelemetry/templates/manifest_prometheus.j2
csibbitt: https://github.com/infrawatch/service-telemetry-operator/pull/553
https://github.com/infrawatch/smart-gateway-operator/pull/154

roles/servicetelemetry/templates/manifest_grafana_ds.j2
```diff
@@ -12,14 +12,13 @@ spec:
...
+        httpHeaderName1: 'Authorization'
       name: STFPrometheus
       type: prometheus
       url: 'https://{{ ansible_operator_meta.name }}-prometheus-proxy.{{ ansible_operator_meta.namespace }}.svc:9092'
       version: 1
-      basicAuth: true
-      basicAuthUser: internal
       secureJsonData:
-        basicAuthPassword: '{{ prom_basicauth_passwd }}'
+        httpHeaderValue1: 'Bearer {{prometheus_reader_token}}'
```
leifmadsen, line 21: Too bad this didn't get the extra spaces here for consistency :) Nothing to block on certainly though.

leifmadsen (approved):

PR #550, opened 2023-12-05
Remove basic-auth method from grafana

csibbitt:
* Only openshift auth will be allowed

### DIFF
--- a/roles/servicetelemetry/templates/manifest_grafana.j2
+++ b/roles/servicetelemetry/templates/manifest_grafana.j2
@@ -7,9 +7,6 @@ spec:
   serviceAccount:
     annotations:
       serviceaccounts.openshift.io/oauth-redirectreference.primary: '{{ grafana_oauth_redir_ref | to_json }}'
-  deployment:
-    annotations:
-      hash-of-creds-to-force-restart-if-changed: {{ grafana_htpasswd_auth_string | b64encode }}
   baseImage: {{ servicetelemetry_vars.graphing.grafana.base_image }}
   ingress:
     enabled: {{ servicetelemetry_vars.graphing.grafana.ingress_enabled }}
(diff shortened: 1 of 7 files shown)


roles/servicetelemetry/templates/manifest_grafana.j2
```diff
@@ -40,13 +37,12 @@ spec:
...
     - -pass-basic-auth=false
     - -https-address=:3002
-    - -htpasswd-file=/etc/proxy/htpasswd/auth
     - -tls-cert=/etc/tls/private/tls.crt
     - -tls-key=/etc/tls/private/tls.key
     - -upstream=http://localhost:3000
     - -cookie-secret-file=/etc/proxy/secrets/session_secret
     - -openshift-service-account=grafana-serviceaccount
-    - '-openshift-sar={"resource": "namespaces", "verb": "get"}'
+    - '-openshift-sar={"namespace":"{{ ansible_operator_meta.namespace }}","resource": "grafana", "group":"integreatly.org", "verb":"get"}'
```
csibbitt, line 45: Another reminder for myself that this sort of change is release-notes worthy.  Basically I'm proposing that instead of requiring accounts with cluster-wide admin-like permissions (the ability to see all namespaces) we allow any accounts with access to read grafana objects in our namespace.

csibbitt: FWIW, making this SAR the same as the one we use on prometheus (permission to read prometheus objects in our namespace) would also make sense to me.

leifmadsen: I had to read through https://github.com/openshift/oauth-proxy#limiting-access-to-users to get a better understanding about what this was doing, but now I get it :)

leifmadsen (approved):

leifmadsen:
Is this all standalone code, or does it depend on any of the other work you've been doing around authentication access?

csibbitt:
> Is this all standalone code, or does it depend on any of the other work you've been doing around authentication access?

Standalone with no other dependencies. The basic auth only existed to provide backwards-compatibility to what we had before we implemented oauth (https://github.com/infrawatch/service-telemetry-operator/pull/281/files#diff-ae71801975adb4f8dd4aa5479a66ad46e46f17de40f9d147b2e09e13ce26633eL19) . Now that OAuth has been in place for several releases (since 1.4.0), and our threat model is suggesting removing all basic-auth and passwords in CRs, we can just turn this all off.

leifmadsen:
> > Is this all standalone code, or does it depend on any of the other work you've been doing around authentication access?
> 
> Standalone with no other dependencies. The basic auth only existed to provide backwards-compatibility to what we had before we implemented oauth (https://github.com/infrawatch/service-telemetry-operator/pull/281/files#diff-ae71801975adb4f8dd4aa5479a66ad46e46f17de40f9d147b2e09e13ce26633eL19) . Now that OAuth has been in place for several releases (since 1.4.0), and our threat model is suggesting removing all basic-auth and passwords in CRs, we can just turn this all off.

Awesome! Then approved and ready for merge at your discretion :)

csibbitt:
recheck

PR #552, opened 2023-12-07
Adjust Alertmanager SAR to be more specific

csibbitt:
* This matches recent changes in prometheus[1] and grafana[2]

[1] https://github.com/infrawatch/service-telemetry-operator/pull/549/files#diff-2cf84bcf66f12393c86949ec0d3f16c473a650173d55549bb02556d23aa22bd2R46
[2] https://github.com/infrawatch/service-telemetry-operator/pull/550/files#diff-ae71801975adb4f8dd4aa5479a66ad46e46f17de40f9d147b2e09e13ce26633eR45

### DIFF
--- a/roles/servicetelemetry/templates/manifest_alertmanager.j2
+++ b/roles/servicetelemetry/templates/manifest_alertmanager.j2
@@ -26,8 +26,8 @@ spec:
     - -upstream=http://localhost:9093/
     - -cookie-secret-file=/etc/proxy/secrets/session_secret
     - -openshift-service-account=alertmanager-stf
-    - '-openshift-sar={"resource": "namespaces", "verb": "get"}'
-    - '-openshift-delegate-urls={"/": {"resource": "namespaces", "verb": "get"}}'
+    - '-openshift-sar={"namespace":"{{ ansible_operator_meta.namespace }}","resource": "alertmanager", "group":"monitoring.rhobs", "verb":"get"}'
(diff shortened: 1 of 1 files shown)


roles/servicetelemetry/templates/manifest_alertmanager.j2
```diff
@@ -26,8 +26,8 @@ spec:
     - -upstream=http://localhost:9093/
     - -cookie-secret-file=/etc/proxy/secrets/session_secret
     - -openshift-service-account=alertmanager-stf
-    - '-openshift-sar={"resource": "namespaces", "verb": "get"}'
-    - '-openshift-delegate-urls={"/": {"resource": "namespaces", "verb": "get"}}'
+    - '-openshift-sar={"namespace":"{{ ansible_operator_meta.namespace }}","resource": "alertmanager", "group":"monitoring.rhobs", "verb":"get"}'
```
leifmadsen, line 29: One issue here is that this is going to break customers using `observabilityStrategy: use_community` right?

csibbitt: Yes, good catch thanks!

csibbitt:
I'm abandoning this and will return with changes for both Alertmanager and SGs as well as adjustmens to the permissions of the prometheus-stf account to match.

leifmadsen:
@csibbitt can we delete the branch associated with this PR or do you plan to base off that branch for a future PR?

PR #531, opened 2023-11-13
[zuul] Use extracted CRC nodes in stf-base

elfiesmelfie:
This updates the stf-base job to use a 2-node/extracted CRC setup instead of a single node/nested CRC set-up.

Based on the example here: https://ci-framework.readthedocs.io/en/latest/cookbooks/zuul-job-nodeset.html

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/837

PR #561, opened 2024-01-09
Implement support for Grafana Operator v5

leifmadsen:
Implement changes to support Grafana Operator v5 when the new
grafana.integreatly.org CRD is available. Use the new CRDs as default
when they are available. Fallover to deploying with Grafana Operator v4
when the Grafana Operator v5 CRDs are not available, thereby providing
backwards compatibility to allow administrators time to migrate.

Additionally, the polystat plugin has been removed from the rhos-cloud
dashboard due to compatibility issues with grafana-cli usage when
dynamically loading plugins. Usage of Grafana Operator v5 is also a
target for disconnected support, and dynamically loading plugins in
these environments is expected to be a problem.

Related: OSPRH-2577
Closes: STF-1667


roles/servicetelemetry/tasks/component_grafana.yml
```diff
@@ -47,94 +26,247 @@
...
+    - when: has_grafana_integreatly_api
+      name: Deploying with Grafana Operator v5
+      block:
+        - name: Construct oauth redirect reference
+          set_fact:
+            grafana_oauth_redir_ref:
+              kind: OAuthRedirectReference
+              apiVersion: v1
+              reference:
+                kind: Route
```
leifmadsen, line 39: The route name is now slightly dynamic and will change from the previously static `grafana-route` to now include the name of the STF deployment name (`default` by.... default). Tracking documentation updates at https://issues.redhat.com/browse/STF-1680

roles/servicetelemetry/tasks/component_grafana.yml
```diff
@@ -47,94 +26,247 @@
...
+
+        - name: Create an instance of Grafana (Grafana Operator v5)
+          k8s:
+            state: '{{ "present" if servicetelemetry_vars.graphing.enabled else "absent" }}'
+            definition:
+              '{{ grafana_manifest }}'
+
+        # NOTE: we only provide events forwarding with STF. We don't use events
+        # in dashboards, so there is no need to create an Elasticsearch
+        # datasource.
```
leifmadsen, line 59: I took some liberties here and only implemented Prometheus as a datasource for Grafana Operator v5. In Grafana Operator v4 the datasources were in a single datasource configuration as a list, but that's not the case here.

I was tempted to also remove the Elasticsearch datasource from the Grafana Operator v4 implementation, but thought that to be out of scope. Could certainly look into do that after this though.

roles/servicetelemetry/tasks/component_grafana.yml
```diff
@@ -47,94 +26,247 @@
...
+              metadata:
+                labels:
+                  app: grafana
+                  stf_owner: "{{ ansible_operator_meta.name }}"
+                name: rhos-cloud-dashboard-1
+                namespace: "{{ ansible_operator_meta.namespace }}"
+              spec:
+                instanceSelector:
+                  matchLabels:
+                    dashboards: "stf"
```
leifmadsen, line 89: You'll see the usage of `instanceSelector` everywhere because that's how association works with Grafana Operator v5.

The Grafana instance will have a traditional k8s label assigned to it as `dashboards: stf` and then the instance selectors on the GrafanaDashboard and GrafanaDatasource objects get bound to instances of Grafana with matching instance selector values.

roles/servicetelemetry/templates/manifest_grafana_ds_prometheus.j2
```diff
@@ -0,0 +1,24 @@
+apiVersion: grafana.integreatly.org/v1beta1
+kind: GrafanaDatasource
+metadata:
+  name: {{ ansible_operator_meta.name }}-ds-stf-prometheus
+  namespace: {{ ansible_operator_meta.namespace }}
+spec:
+  instanceSelector:
+    matchLabels:
+      dashboards: "stf"
```
leifmadsen, line 9: In theory we could also make this dynamic and have the dashboards aligned to the ServiceTelemetry name (`default`) but we don't really have an expectation (or testing, or documentation) around running multiple STF instances in the same namespace, although historically it has seemed to mostly work. (I'd bet there are at least a few gaps now though that doesn't permit that to happen.)

roles/servicetelemetry/templates/manifest_grafana_ds_prometheus.j2
```diff
@@ -0,0 +1,24 @@
...
+  datasource:
+    name: STFPrometheus
```
leifmadsen, line 11: Also have a hard coded datasource here. We could probably go through and make this slightly more dynamic going forward now that we're managing everything in STO now.

I also found dashboard-linter for Grafana which I'd like to implement at some point here, but the dashboards would need to be updated to have a templated datasource, which is the biggest complaint of the linter.

roles/servicetelemetry/templates/manifest_grafana_v5.j2
```diff
@@ -0,0 +1,95 @@
+apiVersion: grafana.integreatly.org/v1beta1
+kind: Grafana
+metadata:
+  name: {{ ansible_operator_meta.name }}-grafana
```
leifmadsen, line 4: The route name comes from the name of the Grafana instance due to how the manifest is configured now. I couldn't find a great way to override this without having to add more RBAC capabilities so the Operator could read and parse the cluster configuration to get the external hostname for routes so I could construct the entire `host` parameter. I didn't feel it was worth that complexity, so we'll have a slightly different route name in our documentation now.

roles/servicetelemetry/templates/manifest_grafana_v5.j2
```diff
@@ -0,0 +1,95 @@
...
+  namespace: {{ ansible_operator_meta.namespace }}
+  labels:
+    dashboards: "stf"
+spec:
+  serviceAccount:
+    metadata:
+      annotations:
+        serviceaccounts.openshift.io/oauth-redirectreference.primary: '{{ grafana_oauth_redir_ref | to_json }}'
+{% if servicetelemetry_vars.graphing.grafana.ingress_enabled is defined and servicetelemetry_vars.graphing.grafana.ingress_enabled %}
```
leifmadsen, line 13: No more ingressEnabled parameter that makes this simpler :) With Grafana Operator v5 there is more of an override approach than providing a bunch of custom parameters in the CRD. Overall I like this approach better.

roles/servicetelemetry/templates/manifest_grafana_v5.j2
```diff
@@ -0,0 +1,95 @@
...
+      enabled: "True"
+    auth.proxy:
+      enabled: "True"
+      enable_login_token: "True"
+      header_property: "username"
+      header_name: "X-Forwarded-User"
+    log:
+      level: warn
+      mode: "console"
+  deployment:
```
leifmadsen, line 42: I leaned heavily on this documentation to get all of this working as intended in Grafana Operator v5: https://grafana.github.io/grafana-operator/docs/examples/oauth_proxy/readme/

There are some really good examples there about how to deploy the various components.

elfiesmelfie:
The CI failures look familiar, and seem to be consistent with intermittent failures due to performance bottlenecks, most likely due to the nested virt. When I've seen these issues in the past, it's been resolved by rechecking.

I expect more consistent results when #531 is merged

leifmadsen:
recheck

roles/servicetelemetry/templates/manifest_grafana_v5.j2
```diff
@@ -0,0 +1,95 @@
...
+{% endif %}
+  client:
+    preferIngress: false
+  config:
+    auth:
+      disable_signout_menu: "{{ servicetelemetry_vars.graphing.grafana.disable_signout_menu }}"
+      disable_login_form: "True"
+    auth.anonymous:
+      enabled: "True"
+    auth.proxy:
```
csibbitt, line 34: I think this auth.proxy config could go away. I experimented with it in the last round of auth changes but ultimately found it unhelpful. The existing Grafana v4 config does not depend on any headers from the proxy, instead treating any user who makes it through the auth proxy as an anonymous admin.

Have you found any benefits to this config other than it showing your username in grafana? Does it treat the named users as admin?

csibbitt (approved):
No real objections, but I am leaning towards dropping the proxy config if we don't need it, just to keep things as simple as possible for people who are debugging this unsupported component.

roles/servicetelemetry/templates/manifest_grafana_v5.j2
leifmadsen: Yes, the dashboard URL will change. Thanks for the reminder. I did think of this for the release notes, but I haven't written it down yet. Now that I have the PR up, I'm going to stage a release note in the main epic that covers this.

deploy/olm-catalog/service-telemetry-operator/manifests/service-telemetry-operator.clusterserviceversion.yaml
```diff
@@ -365,6 +365,7 @@ spec:
           - monitoring.coreos.com
           - monitoring.rhobs
           - elasticsearch.k8s.elastic.co
+          - grafana.integreatly.org
```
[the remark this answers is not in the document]
leifmadsen: Yes exactly. The `group` for v4 is integreatly.org and the `group` for v5 is `grafana.integreatly.org`. I used that difference to determine which Operator was installed.

roles/servicetelemetry/templates/manifest_grafana_v5.j2
leifmadsen: No benefit necessarily. As noted in an earlier comment, most of this is hear because I followed the example to get things working. I am going to attempt to drop this and make sure I can still get through to the Grafana instance.

If this works, in theory I could add a simple flag to use this or not, but without a strong need for it, that's probably just more overhead work that isn't necessary for an unsupported component.

I will drop this and if it works as it did before, I will call that a win and a completed experience.

leifmadsen:
> No real objections, but I am leaning towards dropping the proxy config if we don't need it, just to keep things as simple as possible for people who are debugging this unsupported component.

I ended up needing the proxy config to get dashboards and datasources to apply correctly. Without it, I wouldn't get dashboards, and wouldn't be logged in, which is how the previous deployments would have worked.

leifmadsen:
So I also want to change the default image to be rhel8/grafana:9 vs 7 on the version, since it looks like Grafana v7 has gone EOL, and even downstream builds are now targeting Grafana 9 which continues to receive upstream CVE support.

I've been using grafana:9 in testing and everything seems to work. I guess changing the version and deleting/recreating the Grafana workload in the next minor release of STF is a bit awkward, but probably the better default going forward?

leifmadsen:
> So I also want to change the default image to be rhel8/grafana:9 vs 7 on the version, since it looks like Grafana v7 has gone EOL, and even downstream builds are now targeting Grafana 9 which continues to receive upstream CVE support.
> 
> I've been using grafana:9 in testing and everything seems to work. I guess changing the version and deleting/recreating the Grafana workload in the next minor release of STF is a bit awkward, but probably the better default going forward?

I'm just going to leave everything default here, and if we find notification that Grafana:7 image from RHEL is no longer maintained downstream for CVEs, it should be easy enough to move to another image without any changes.

No changes for now other than what I've already implemented.

leifmadsen:
recheck

PR #563, opened 2024-01-15
Remove old vendored operator_sdk/util collection

leifmadsen:
Remove the old 0.1.0 vendored collection operator_sdk/util from the
upstream Dockerfile and repository. Instead use the default
operator_sdk/util in the base image which is a newer version of 0.4.0.

We only use the util collection for one call to k8s_status when
ephemeral storage is enabled. The newer collection also provides a
k8s_event module which could be useful in the future.

Closes: STF-1683


csibbitt (approved):
CI doesn't test ephemeral, but customers aren't supposed to use it so I'll trust your pinky-promise that it works. :)

PR #571, opened 2024-01-31
[stf-run-ci] Update validation check for bundle URLs

elfiesmelfie:
An empty string passed as the bundle URL will pass the existing test of "is defined" and "is not None" and still be invalid.

* add a check for "length != 0"
* assert requires all conditions to be met, the checks can all be listed separately to make the task easier to grok.

### DIFF
--- a/build/stf-run-ci/tasks/setup_stf_from_bundles.yml
+++ b/build/stf-run-ci/tasks/setup_stf_from_bundles.yml
@@ -90,12 +90,8 @@
 - name: "Ensure that the bundle paths are set."
   ansible.builtin.assert:
     that:
-      - '__smart_gateway_bundle_image_path is defined'
-      - '__smart_gateway_bundle_image_path != None'
-      - '__smart_gateway_bundle_image_path | length != 0'
-      - '__service_telemetry_bundle_image_path is defined'
-      - '__service_telemetry_bundle_image_path != None'
-      - '__service_telemetry_bindle_image_path | length != 0'
+      - '__smart_gateway_bundle_image_path | default("") | length > 0'
+      - '__service_telemetry_bindle_image_path | default("") | length > 0'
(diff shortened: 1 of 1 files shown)


ayefimov-1 (approved):
Changes look good.

elfiesmelfie:
You can run a test for this using: https://gist.github.com/elfiesmelfie/e6705ea2a2f70b04a3ad8787fa93bd5b

build/stf-run-ci/tasks/setup_stf_from_bundles.yml
```diff
@@ -90,10 +90,14 @@
 - name: "Ensure that the bundle paths are set."
   ansible.builtin.assert:
     that:
-      - '__smart_gateway_bundle_image_path is defined and __smart_gateway_bundle_image_path != None'
-      - '__service_telemetry_bundle_image_path is defined and __service_telemetry_bundle_image_path != None'
+      - '__smart_gateway_bundle_image_path is defined'
```
csibbitt, line 93: Instead of these three checks per variable, I think the idiomatic ansible thing is this:
```
'__smart_gateway_bundle_image_path | default("") | length > 0
```

csibbitt (approved):
Not going to block it, but I suggested a simplification.

build/stf-run-ci/tasks/setup_stf_from_bundles.yml
elfiesmelfie: That's fair.

If the var is undefined, it becomes "", and the check fails, because of length
If the var is None, there's an error because None does not have a length
If the var is an empty string, the check fails because of the length

Okay, I'm convinced. And there was a job failure so the CI would run again anyway

elfiesmelfie: I'm leaving the success message as-is

csibbitt (approved):

ayefimov-1 (approved):
looks correct

PR #572, opened 2024-02-01
Add optional spec.replaces field to CSV for update graph compliance

### LINKED ISSUE #559: Bundle generation doesn't allow setting spec.replaces
The way we generate our CSVs uses OLM's `skipRange` functionality. This is fine, but using only this leads to older versions becoming unavailable after the fact:

https://olm.operatorframework.io/docs/concepts/olm-architecture/operator-catalog/creating-an-update-graph/#skiprange
> **WARNING**: As a consequence of using skipRange the skipped operator versions will be pruned from OLMs update graph and therefore will not be installable anymore by users with the spec.startingCSV property of Subscriptions. Use with caution. If you want to have direct (one hop) updates to this version from multiple previous versions but keep those previous versions available to users for install, always use skipRange in conjunction with replaces pointing to the immediate previous version of the operator version in question.
>
> SkipRange by itself is useful for teams who are not interested in supporting directly installing versions within a given range or for whom consumers of the operator are always on the latest version.

This is likely fine for most upstream builds where we build ad hoc catalogs for testing new changes, but it would be useful to add a `spec.replaces` field that can be optionally set via an environment variable.

This will allow update testing, as well as actual production updates for downstream builds that leverage the new field.

compi-migui:
The way we generate our CSVs uses OLM's skipRange functionality. This is fine, but using only this leads to older versions becoming unavailable after the fact -- see the warning at [1].

By adding an optional spec.replaces to our CSV we allow update testing as well as actual production updates for downstream builds that leverage it.

Populating the field requires knowledge of the latest-released bundle, so we take it from an environment variable to be provided by the builder. If this is unset we don't include the spec.replaces field at all -- leaving previous behavior unchanged.

Resolves #559
Related: STF-1658

[1] https://olm.operatorframework.io/docs/concepts/olm-architecture/operator-catalog/creating-an-update-graph/#skiprange

(cherry picked from commit 99221fbffe2d658161c00fae5905aebe0d989ffa)

### DIFF
--- a/build/generate_bundle.sh
+++ b/build/generate_bundle.sh
@@ -35,6 +35,15 @@ generate_bundle() {
     ${OPERATOR_SDK} generate bundle --verbose --channels ${BUNDLE_CHANNELS} --default-channel ${BUNDLE_DEFAULT_CHANNEL} --manifests --metadata --version "${OPERATOR_BUNDLE_VERSION}" --output-dir "${WORKING_DIR}" >> ${LOGFILE} 2>&1
     popd > /dev/null 2>&1
 
+    # CSVs without a spec.replaces field are valid, so fall back to those if
+    # latest released version is unknown.
+    # Placeholder value is validated by operator-sdk during local bundle
+    # generation and so needs to conform to RFC1123.
+    if [[ -n "$BUNDLE_LATEST_RELEASED_VERSION" ]]; then
+        REPLACE_REGEX="$REPLACE_REGEX;s#---bundle-latest-released-version#${BUNDLE_LATEST_RELEASED_VERSION}#g"
+    else sed -i '/---bundle-latest-released-version/d' "${WORKING_DIR}/manifests/${OPERATOR_NAME}.clusterserviceversion.yaml"
+    fi
+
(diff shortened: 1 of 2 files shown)


compi-migui:
Cherry-pick from the master branch https://github.com/infrawatch/service-telemetry-operator/pull/560 to ensure we pick this up before any other upcoming changes.

elfiesmelfie:
Since the branch names match, a successful jenkins build here picks up this change and the sgo change

csibbitt (approved):
Same suggestions as https://github.com/infrawatch/smart-gateway-operator/pull/157#pullrequestreview-1857584666  but I see these are both cherry-picks, so that comment belonged on the review that brought these to master. Feel free to ignore for now.

compi-migui:
> Same suggestions as [infrawatch/smart-gateway-operator#157 (review)](https://github.com/infrawatch/smart-gateway-operator/pull/157#pullrequestreview-1857584666) but I see these are both cherry-picks, so that comment belonged on the review that brought these to master. Feel free to ignore for now.

Addressed in that other review, see https://github.com/infrawatch/smart-gateway-operator/pull/157#discussion_r1475703989

I agree completely but operator-sdk forces our hand unless we rework bundle generation in a major way.

elfiesmelfie (approved):

PR #574, opened 2024-02-08
STF 1.5.4 release ops

vkmc:
Depends-On: https://github.com/infrawatch/smart-gateway-operator/pull/158
Depends-On: https://github.com/infrawatch/sg-core/pull/129
Depends-On: https://github.com/infrawatch/sg-bridge/pull/35
Depends-On: https://github.com/infrawatch/prometheus-webhook-snmp/pull/20


.zuul.yaml
```diff
@@ -0,0 +1,216 @@
...
+      - zuul: github.com/openstack-k8s-operators/ci-framework
+    # These are the additional repos that zuul will clone
+    required-projects:
+      - name: openstack-k8s-operators/ci-framework
+        override-checkout: main
+      - name: github.com/infrawatch/service-telemetry-operator
+      - name: github.com/infrawatch/smart-gateway-operator
+      - name: github.com/infrawatch/sg-bridge
+      - name: github.com/infrawatch/sg-core
+      - name: github.com/infrawatch/prometheus-webhook-snmp
```
leifmadsen, line 94: @elfiesmelfie these don't need overrides for `stable-1.5` do they?

CC: @mgirgisf

deploy/crds/infra.watch_v1beta1_servicetelemetry_cr.yaml
```diff
@@ -79,10 +79,11 @@ spec:
   graphing:
     enabled: false
     grafana:
-      ingressEnabled: false
-      adminPassword: secret
-      adminUser: root
+      ingressEnabled: true
       disableSignoutMenu: false
+      baseImage: registry.redhat.io/rhel8/grafana:7
```
leifmadsen, line 84: Ah shoot, I think we wanted to change this to `grafana:9`. Might need to get a PR up for that quickly. I have one other thing to review from a docs PR I have open that Chris commented on too.

roles/servicetelemetry/defaults/main.yml
```diff
@@ -84,11 +84,11 @@ servicetelemetry_defaults:
   graphing:
     enabled: false
     grafana:
-      ingress_enabled: false
-      admin_password: secret
-      admin_user: root
+      ingress_enabled: true
       disable_signout_menu: false
-      base_image: docker.io/grafana/grafana:8.1.2
+      base_image: registry.redhat.io/rhel8/grafana:7
```
leifmadsen, line 89: Would like to change this to grafana:9

.zuul.yaml
mgirgisf: I think we need to add override-checkout to point at the proposed-release branches for the repos in infrawatch.

Note: except  - name: github.com/infrawatch/service-telemetry-operator this don't need to be override to be able to test the changes in this branch

roles/servicetelemetry/defaults/main.yml
vkmc: Addressed by https://github.com/infrawatch/service-telemetry-operator/pull/574/commits/28183df3ad5200351d495b7f823a6d0d15dd0897

deploy/crds/infra.watch_v1beta1_servicetelemetry_cr.yaml
vkmc: Addressed by https://github.com/infrawatch/service-telemetry-operator/pull/574/commits/28183df3ad5200351d495b7f823a6d0d15dd0897

.zuul.yaml
csibbitt, line 94:
```suggestion
      - name: github.com/infrawatch/smart-gateway-operator
        override-checkout: stable-1.5
      - name: github.com/infrawatch/sg-bridge
        override-checkout: stable-1.5
      - name: github.com/infrawatch/sg-core
        override-checkout: stable-1.5
      - name: github.com/infrawatch/prometheus-webhook-snmp
        override-checkout: stable-1.5
```

@elfiesmelfie can we get your :+1:  on this? Do we need these?

csibbitt (commented):
I proposed a change to the zuul config based on the conversation. Rest LGTM

.zuul.yaml
csibbitt: Reading zuul docs[1], I think these are not required

"When Zuul runs jobs for a proposed change, it normally checks out the branch associated with that change on every project present in the job"

In fact I think this would be harmful, for example it could fail to test our release-prep-1.5.4 branch because of dependent changes that are not on stable-1.5 yet

[1] https://zuul-ci.org/docs/zuul/latest/config/job.html#attr-job.required-projects.override-checkout

elfiesmelfie: IMHO, it's better to be explicit, since I'm not 100% sure which the base branch will be. 

There is a quick way to check which versions Zuul has checked out.
Zuul creates an inventory file that shows the vars set for the job, which is available in the logs as the zuul-info/inventory.yaml:

https://logserver.rdoproject.org/74/574/c7cfcd3e735e03a2ae3d872b337eaa9eb3fbfa0a/github-check/stf-crc-ocp_414-local_build-index_deploy/d39737d/zuul-info/inventory.yaml

I think these SHOULD be pointed at release-prep-1.5.4 right now.
In order to test it properly, I recommend having this pr use stable-1.5 (so that it is testing stable-1.5 AFTER it merges)
For testing the release-prep branch, we may need to open another PR with the override-checkout set to `release-prep-1.5.4`.
This is needed because we don't have any zuul job on the target (stable-1.5) branch.

I am going to check the SHAs for the commits and confirm whether this change is needed.

elfiesmelfie: I am not 100% sure. They may need an override for a while, until zuul is merged into stable-1.5.

elfiesmelfie: Okay, I'll update the branches.

It might be better to have a follow-up to checkout the stable-1.5 branch, and add release-prep into this one.
I'll update this comment accordingly.

elfiesmelfie: TL;DR: it was fine the way it was, not need to add the override-checkout.

elfiesmelfie: There are additonal comments on this conversation below on Chris's proposed patch (which I unnecessarily updated).

I was over thinking this earlier, sorry for all the confusion.

TL;DR: this can stay as-is, but we need to use Depends-On to point at the open PRs until they merge.

elfiesmelfie:
check-rdo

csibbitt (approved):

leifmadsen (approved):

elfiesmelfie:
The PRs to the other repos need to merge first. Zuul does not handle merging to this repo, so having the Depends-On PR merged is not enforced

PR #581, opened 2024-02-26
Fix env var naming issue

leifmadsen:
Fix an environment variable naming issue in the CSV for STO when
attempting to install alertmanager disconnected. The env var being
looked up should have a postfix of _IMAGE to match the other env vars.

Found in testing by vkmc.


vkmc (approved):
LGTM, thanks!

PR #583, opened 2024-02-29
Update smoketest.sh

ayefimov-1:
changed default-alertmanager URL to v2

PR #584, opened 2024-02-29
Update smoketest.sh (#583)

elfiesmelfie:
Backport of https://github.com/infrawatch/service-telemetry-operator/pull/583


leifmadsen:
Marking this as a do-not-merge as the current midstream will attempt to pull this in due to a lack of tagging requirements. I might propose a change to allow us to merge between releases, but we'll see. Either way, let's avoid merging this until after the current release ships.

leifmadsen:
Based on the discussion I've had with Miguel, we don't need to block on this merge.

leifmadsen (approved):

PR #591, opened 2024-04-17
Add __deploy_disconnected option

vkmc:
This new option if stf-run-ci allows the utilization of
the setup_base and setup_stf roles for the purpose of deploying
STF on OCP Disconnected

It is expected that this is run in an environment in which the
catalogsources have been defined and the precise catalogsource for
each of the dependency operators needs to be passed to the role.

For this, the catalogsource for each of the operators can be modified
with the new parameters redhat_operators, community_operators,
certified_operators and infrawatch_operators. We also need to specify
the channel, being the default "unstable" (used when building from index).

Default values for these are set to the previous values to keep
backwards compatibility

Depends-On: https://github.com/infrawatch/service-telemetry-operator/pull/593


build/stf-run-ci/tasks/setup_stf.yml
```diff
@@ -1,5 +1,6 @@
 ---
 - name: Set default InfraWatch OperatorSource manifest
+  when: __deploy_from_index_enabled | bool and infrawatch_catalog_source_manifest is not defined
```
elfiesmelfie, line 3: fyi: for AND conditions, you can provide a list

```
when:
  - condition1
  - condition2
```

elfiesmelfie:
recheck

vkmc:
recheck

csibbitt (approved):

PR #592, opened 2024-04-18
Fail earlier when registry creds are not set

vkmc:
Move the credential setup for the internal registry up in the execution and perform a simple check with the "oc image info" command to fail earlier in case the credentials haven't been set properly

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1403

### DIFF
--- a/build/stf-run-ci/tasks/main.yml
+++ b/build/stf-run-ci/tasks/main.yml
@@ -55,7 +55,7 @@
   tags:
     - pre-clean
 
-- name: Set up internal registry credentials (deploy from bundles)
+- name: Set up bundle registry credentials (deploy from bundles)
   ansible.builtin.include_tasks: setup_registry_auth.yml
   when:
   - __deploy_from_bundles_enabled | bool
--- a/build/stf-run-ci/tasks/main.yml
+++ b/build/stf-run-ci/tasks/main.yml
@@ -49,9 +49,7 @@
 
 - name: Set up bundle registry credentials (deploy from bundles)
   ansible.builtin.include_tasks: setup_registry_auth.yml
-  when:
-  - __deploy_from_bundles_enabled | bool
-  - setup_bundle_registry_auth | bool
+  when: __deploy_from_bundles_enabled | bool or setup_bundle_registry_auth | bool
 
 - name: Setup supporting Operator subscriptions
   ansible.builtin.include_tasks: setup_base.yml
--- a/build/stf-run-ci/tasks/main.yml
+++ b/build/stf-run-ci/tasks/main.yml
@@ -55,6 +55,12 @@
   tags:
     - pre-clean
 
+- name: Set up internal registry credentials (deploy from bundles)
+  ansible.builtin.include_tasks: setup_registry_auth.yml
+  when:
+  - __deploy_from_bundles_enabled | bool
+  - setup_bundle_registry_auth | bool
+
 - name: Setup supporting Operator subscriptions
   ansible.builtin.include_tasks: setup_base.yml
   tags:
--- a/build/stf-run-ci/tasks/setup_registry_auth.yml
+++ b/build/stf-run-ci/tasks/setup_registry_auth.yml
@@ -82,12 +82,20 @@
     fail_msg: "Bundle path(s) not set. __smart_gateway_bundle_image_path is '{{ __smart_gateway_bundle_image_path }}' and __service_telemetry_bundle_image_path is '{{ __service_telemetry_bundle_image_path }}'. Both values need to be set."
     success_msg: "Bundle paths are defined, are not None and have a non-zero-length"
 
-- name: Check that we can access to the STO bundle
+- name: Try to access to the STO bundle
   ansible.builtin.command: oc image info {{ __service_telemetry_bundle_image_path }}
   register: sto_bundle_info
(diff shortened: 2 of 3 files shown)


elfiesmelfie:
recheck

elfiesmelfie (changes requested):

build/stf-run-ci/tasks/setup_registry_auth.yml
```diff
@@ -82,12 +82,20 @@
...
   register: sgo_bundle_info
-  failed_when: sgo_bundle_info.rc != 0
+  ignore_errors: true
+
+- name: Check successful read access to STO and SGO bundles in the internal registry
+  ansible.builtin.assert:
+    that:
+      - sto_bundle_info.rc != 0
+      - sgo_bundle_info.rc != 0
+    fail_msg: "Bundles couldn't be retrieved. Check configuration for the internal registry and retry."
```
elfiesmelfie, line 100: The issue is that the credentials may not be correct too, right?

build/stf-run-ci/tasks/setup_registry_auth.yml
```diff
@@ -82,12 +82,20 @@
...
+    success_msg: "Bundles were correctly retrieved from the internal registry."
```
elfiesmelfie, line 101: Not necessarily internal registry. This code patch should be followed for pulling nightly bundles from quay, for example

build/stf-run-ci/tasks/setup_registry_auth.yml
vkmc: Actually, it could be that any of the params passed for the internal registry is wrong: pull_secret_registry, pull_secret_user and pull_secret_pass. Perhaps instead of "internal registry" we could use "bundle registry". Would that be clearer?

build/stf-run-ci/tasks/setup_registry_auth.yml
vkmc: Right, maybe "bundle registry" is more accurate

build/stf-run-ci/tasks/main.yml
```diff
@@ -47,6 +47,12 @@
   tags:
     - pre-clean
 
+- name: Set up bundle registry credentials (deploy from bundles)
+  ansible.builtin.include_tasks: setup_registry_auth.yml
+  when:
+  - __deploy_from_bundles_enabled | bool
+  - setup_bundle_registry_auth | bool
```
csibbitt, line 54: Previously, only the auth-related block was behind this flag (L1 in setup_stf_from_bundles.yml), and there was the ability to alter `setup_bundle_registry_auth` and `setup_bundle_registry_tls_ca` separately depending on your needs based on where the bundles are hosted. 

AFAICT, both auth and tls are behind this same flag now, and setup_bundle_registry_tls_ca no longer does anything. That's probably okay, at least I don't know of any use cases offhand where we install from bundles with TLS but no auth or vice-versa, but if you're removing this functionality I think you should also change the README to reflect:

https://github.com/infrawatch/service-telemetry-operator/blob/fc28f2f77192cb81cea8fd5aa95bf2d51edcb7e8/build/stf-run-ci/README.md?plain=1#L26

https://github.com/infrawatch/service-telemetry-operator/blob/fc28f2f77192cb81cea8fd5aa95bf2d51edcb7e8/build/stf-run-ci/README.md?plain=1#L122

build/stf-run-ci/tasks/setup_registry_auth.yml
```diff
@@ -0,0 +1,101 @@
+- name: Update Pull Secret with bundle registry credentials
+  block:
```
csibbitt, line 2: This doesn't need to be a block anymore because you removed the condition

build/stf-run-ci/tasks/setup_registry_auth.yml
```diff
@@ -0,0 +1,101 @@
...
+- name: Ensure that the bundle paths are set
+  ansible.builtin.assert:
+    that:
+      - '__smart_gateway_bundle_image_path | default("") | length > 0'
+      - '__service_telemetry_bundle_image_path | default("") | length > 0'
+    fail_msg: "Bundle path(s) not set. __smart_gateway_bundle_image_path is '{{ __smart_gateway_bundle_image_path }}' and __service_telemetry_bundle_image_path is '{{ __service_telemetry_bundle_image_path }}'. Both values need to be set."
+    success_msg: "Bundle paths are defined, are not None and have a non-zero-length."
+
+- name: Try to access to the STO bundle
+  ansible.builtin.command: oc image info {{ __service_telemetry_bundle_image_path }}
```
csibbitt, line 86: :+1:

csibbitt (approved):
Looks like it will work, but you should consider either preserving the relevance of the setup_bundle_registry_tls_ca flag (and re-separating it from the authentication; I guess in a different file now), or altering the README to reflect the removal of that functionality.

build/stf-run-ci/tasks/main.yml
vkmc: That's a good point, I don't think we should alter the behavior, even though we don't have proof of this feature being used somewhere else. I will fix this in a follow up patch.

csibbitt (approved):

PR #593, opened 2024-04-25
Set a default image for ElasticSearch to use

vkmc:
Currently we are not selecting the image for ElasticSearch that the ElasticSearch Operator pulls, we are allowing the default (in ElasticSearch registry) to be used

Add the option to select the ElasticSearch image to use

Also bumps the version to ElasticSearch 7.17.20, since 7.16.1 has several vulnerabilities

### DIFF
--- a/build/stf-run-ci/defaults/main.yml
+++ b/build/stf-run-ci/defaults/main.yml
@@ -37,7 +37,7 @@ default_operator_registry_image_base: registry.redhat.io/openshift4/ose-operator
 default_operator_registry_image_tag: v4.13
 
 elasticsearch_version: 7.17.20
-elasticsearch_image: "registry.redhat.io/elastic/elasticsearch:7.17.20"
+elasticsearch_image: registry.connect.redhat.com/elastic/elasticsearch
 
 sgo_image_tag: latest
 sto_image_tag: latest
--- a/build/stf-run-ci/defaults/main.yml
+++ b/build/stf-run-ci/defaults/main.yml
@@ -36,7 +36,8 @@ __smart_gateway_bundle_image_path: "quay.io/infrawatch-operators/smart-gateway-o
 default_operator_registry_image_base: registry.redhat.io/openshift4/ose-operator-registry
 default_operator_registry_image_tag: v4.13
 
-elasticsearch_version: 7.16.1
+elasticsearch_version: 7.17.20
+elasticsearch_image: "registry.redhat.io/elastic/elasticsearch:7.17.20"
 
 sgo_image_tag: latest
 sto_image_tag: latest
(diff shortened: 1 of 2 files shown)


build/stf-run-ci/defaults/main.yml
```diff
@@ -36,7 +36,8 @@ __smart_gateway_bundle_image_path: "quay.io/infrawatch-operators/smart-gateway-o
 default_operator_registry_image_base: registry.redhat.io/openshift4/ose-operator-registry
 default_operator_registry_image_tag: v4.13
 
-elasticsearch_version: 7.16.1
+elasticsearch_version: 7.17.20
```
csibbitt, line 39: It seems not ideal to pin a specific version that needs to be kept up to date, so I went looking for a tag like :7 or :latest, and I see that there isn't one (neither in our registry nor the official one), and it looks like the version specification on the CR is not optional, which I guess is why this is here. 

We have an end-user KB[1] about this and FWIW it shows 7.17.11 (was probably the latest at time of writing), and despite the caveat that it's an example only, we might want to keep them in sync for consistency and ensuring we've tested with what we recommend to install.

[1] https://access.redhat.com/articles/7031236

build/stf-run-ci/defaults/main.yml
```diff
@@ -36,7 +36,8 @@ __smart_gateway_bundle_image_path: "quay.io/infrawatch-operators/smart-gateway-o
...
+elasticsearch_image: "registry.redhat.io/elastic/elasticsearch:7.17.20"
```
csibbitt, line 40:
```suggestion
elasticsearch_image: "registry.redhat.io/elastic/elasticsearch"
```

build/stf-run-ci/templates/manifest_elasticsearch.j2
```diff
@@ -49,4 +49,5 @@ spec:
       certificateAuthorities: {}
   updateStrategy:
     changeBudget: {}
-  version: {{ elasticsearch_version }}
\ No newline at end of file
+  version: {{ elasticsearch_version }}
+  image: {{ elasticsearch_image }}
```
csibbitt, line 53:
```suggestion
  image: {{ elasticsearch_image }}:{{ elasticsearch_version }}
```

csibbitt (approved):
LGTM but suggested a minor DRY change

build/stf-run-ci/defaults/main.yml
vkmc: That's a good point. For some reason I see there is no latest tag, but I can retrieve an elasticsearch image with the latest tag. We can directly use that. I will perform some more tests and see if we can update.

vkmc: Following up on this. The ElasticSearch CRD here expects a version number (cannot be latest). Also, the version field is required. So unfortunately we need to explicitly set a desired version. I will set it to 7.17.20, which is the latest version available in the RH registry.

csibbitt (approved):

PR #594, opened 2024-04-25
Bump actions checkout to v4

vkmc:
Node.js 16 actions are deprecated. We need to update to Node.js 20, which is included in actions/checkout@v4.


elfiesmelfie (approved):
All the actions still work, so looks like the updates are fine

csibbitt (approved):

PR #595, opened 2024-05-08
Change STO/SGO bundles sanity check

vkmc:
stdout for "oc image info" should be checked instead of the rc to make sure that the correct bundle is being used

### DIFF
--- a/build/stf-run-ci/tasks/setup_registry_auth.yml
+++ b/build/stf-run-ci/tasks/setup_registry_auth.yml
@@ -95,7 +95,7 @@
     msg:
     - "{{ sto_bundle_info }}"
 
-- name: Fail is the expected STO bundle was not correctly fetched
+- name: Fail if the expected STO bundle was not correctly fetched
   ansible.builtin.fail:
     msg: "The requested STO bundle couldn't be retrieved from the bundle registry. Check configuration for the bundles registry and retry."
   when: "__service_telemetry_bundle_image_path not in sto_bundle_info.stdout"
--- a/build/stf-run-ci/tasks/setup_registry_auth.yml
+++ b/build/stf-run-ci/tasks/setup_registry_auth.yml
@@ -114,11 +114,3 @@
   ansible.builtin.fail:
     msg: "The requested SGO bundle couldn't be retrieved from the bundle registry. Check configuration for the bundles registry and retry."
   when: "__smart_gateway_bundle_image_path not in sgo_bundle_info.stdout"
-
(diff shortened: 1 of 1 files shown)


build/stf-run-ci/tasks/setup_registry_auth.yml
```diff
@@ -90,15 +90,35 @@
...
 
+- name: Print STO bundle info
+  ansible.builtin.debug:
+    msg:
+    - "{{ sto_bundle_info }}"
+
+- name: Fail is the expected STO bundle was not correctly fetched
+  ansible.builtin.fail:
+    msg: "The requested STO bundle couldn't be retrieved from the bundle registry. Check configuration for the bundles registry and retry."
+  when: "__service_telemetry_bundle_image_path not in sto_bundle_info.stdout"
```
elfiesmelfie, line 101: Regarding verbosity, this check is the same as the assert below.

You could update the message for assert to list the failed bundle.
Something like:
```
fail_msg:
    "The requested bundle(s) ({% if __service_telemetry_bundle_image_path not in sto_bundle_info.stdout %}STO{%endif%}, {% if __smart_gateway_bundle_image_path not in sto_bundle_info.stdout %}SGO{%endif%}) couldn't be retrieved..."
```
i.e. use jinja2 templating to add in the name of the bundles(s) that weren't retrievable to the task in L98.
I'm assuming that you wanted to add these tests here so that there is a useful message to tell you which bundle was missing.

vkmc: Thanks for the review Emma. I opted for dropping the assertion altogether and just stick with the fail checks. Also, kept the debug, I think it's useful info.

elfiesmelfie: That's fair. looks good

elfiesmelfie (approved):

build/stf-run-ci/tasks/setup_registry_auth.yml
```diff
@@ -90,15 +90,27 @@
   register: sto_bundle_info
   ignore_errors: true
 
+- name: Print STO bundle info
+  ansible.builtin.debug:
+    msg:
+    - "{{ sto_bundle_info }}"
+
+- name: Fail is the expected STO bundle was not correctly fetched
```
csibbitt, line 98:
```suggestion
- name: Fail if the expected STO bundle was not correctly fetched
```

csibbitt (approved):

vkmc:
recheck

vkmc:
recheck

PR #597, opened 2024-05-14
Create index image from pre-built bundles

vkmc:
Deploy from bundles can now be used along with deploy from index enabled.

This way, an index image can be created with specific pre built bundles and deploy STF with those.


vkmc:
Still have to work on updating docs, further testing, and better handling of the bundles version (we should honor the sgo_bundle_tag and sto_bundle_tag we already have instead of manipulating the tags directly from the sto ang sto bundles path)

But... managed to get a working deployment :)

vkmc:
recheck

mrunge:
There was an earlier outage on RDO cloud, resulting in "could not download metadata" errors

vkmc:
recheck

build/stf-run-ci/tasks/create_catalog.yml
```diff
@@ -7,11 +7,57 @@
...
+        stf_channel: stable-1.5
+    - name: Set info variables from provided pre-built bundles
+      ansible.builtin.set_fact:
+        sto_bundle_info:
+          'bundle_default_channel': "{{ stf_channel }}"
+          'bundle_channels': "{{ stf_channel }}"
+          'operator_bundle_version': "{{ sto_bundle_image_tag }}"
+        sgo_bundle_info:
+          'bundle_default_channel': stable-1.5
+          'bundle_channels': stable-1.5
```
elfiesmelfie, line 30: Should this also be `"{{ stf_channel }}"`?

build/stf-run-ci/tasks/main.yml
```diff
@@ -159,7 +138,7 @@
           pod-security.kubernetes.io/audit: restricted
           pod-security.kubernetes.io/warn: restricted
 
-- when: __deploy_from_index_enabled | bool
+- when: __deploy_from_index_enabled | bool and __local_build_enabled | bool
```
elfiesmelfie, line 141: This looks like the behaviour for __local_build_enabled is being changed.

This is needed because `__deploy_from_index_enabled==true` previous implied `__local_build_enabled==true` previously.

build/stf-run-ci/tasks/main.yml
```diff
@@ -206,7 +187,7 @@
         name: service-telemetry-operator
         namespace: "{{ namespace }}"
       spec:
-        channel: unstable
+        channel: "{{ stf_channel }}"
```
elfiesmelfie, line 190: Perhaps I missed it, but where is this set to "unstable" for a local build?

elfiesmelfie:
The arbitrary CI failure that we were seeing last week should be addressed.
New issues are present due to CentOS 8 Stream going EOL on 31/05/24. These new issues will be resolved once this change merges : https://github.com/infrawatch/sg-core/pull/135.
You can use a Depends-On in the PR description to test until the PR merges.

build/stf-run-ci/tasks/main.yml
csibbitt: It's the default https://github.com/infrawatch/service-telemetry-operator/blob/b46f215d5a45a4b0b3703c3c7e54d2a4316ba02f/build/stf-run-ci/defaults/main.yml#L64

csibbitt (commented):
This LGTM (not having tested all, or any, the modes). Agree with Emma about the missing s/stable-1.5/{{stf_channel}}/

vkmc:
recheck

vkmc:
recheck

vkmc:
recheck

build/stf-run-ci/tasks/create_catalog.yml
```diff
@@ -7,11 +7,57 @@
...
+      ansible.builtin.command:
+        cmd: |
+          oc import-image -n {{ namespace }} service-telemetry-operator-bundle:{{ sto_bundle_image_tag }} --from={{ __service_telemetry_bundle_image_path }} --confirm --insecure
+      register: sto_is
+
+    - name: Create ImageStream for SGO
+      ansible.builtin.command:
+        cmd: |
+          oc import-image -n {{ namespace }} smart-gateway-operator-bundle:{{ sgo_bundle_image_tag }} --from={{ __smart_gateway_bundle_image_path }} --confirm --insecure
+      register: sgo_is
```
elfiesmelfie, line 59: These could all be combined into the same block.

The ``when:`` on L47 should  have the bool filter, otherwise there may be unexpected behaviour since the value is read as a string

build/stf-run-ci/tasks/create_catalog.yml
```diff
@@ -7,11 +7,57 @@
...
+          'bundle_default_channel': "{{ stf_channel }}"
+          'bundle_channels': "{{ stf_channel }}"
+          'operator_bundle_version': "{{ sgo_bundle_image_tag }}"
+
+- name: Show STO and SGO bundle info that will used in the index image
+  when: __local_build_enabled | bool or __deploy_from_bundles_enabled | bool
+  ansible.builtin.debug:
+    msg:
+    - "{{ sto_bundle_info }}"
+    - "{{ sgo_bundle_info }}"
```
elfiesmelfie, line 38: Is deploy_from index AND not deploy_from_bundle AND not local_build a valid option?

IF it is not, then this should not need a `when: __local_build_enabled | bool or __deploy_from_bundles_enabled | bool`

build/stf-run-ci/tasks/create_catalog.yml
vkmc: Good catch!

build/stf-run-ci/tasks/create_catalog.yml
vkmc: Another good catch

vkmc:
Thanks for the review Emma, of course, I will propose a job to see it running

vkmc:
recheck

vkmc:
recheck

elfiesmelfie (approved):

PR #606, opened 2024-06-18
Fully disable qdr web by default

csibbitt:
The default config has the http management port listening even when it seemed to be "disabled" (option only affected the route).

We want it truly disabled by default since it doesn't have any oauth-proxy protection.


vkmc (approved):
Good catch, thanks

PR #607, opened 2024-06-19
Use scrapeconfig instead of servicemonitor

csibbitt:
This allows us to create a prometheus config where targets are specified by DNS name, rather than IP addresses.

Solves a problem discovered here: https://github.com/infrawatch/service-telemetry-operator/pull/599#issuecomment-2137386683

Risks:

- This may not work when using `observability_strategy: use_community` on default deployments of OCP <= 4.14 (at least) 
  - the prometheus operator from openshift-monitoring _looks_ new enough, but the scrapeconfig CRD was missing in my test environments
  - `observability_strategy: use_redhat` w/ obo-prometheus-operator from COO 0.2.0 works fine 
- I haven't done additional testing, for example ~to ensure metrics labeling compatibility with all the other parts of the system, or~ impact on the (deprecated) HA mode


deploy/olm-catalog/service-telemetry-operator/manifests/service-telemetry-operator.clusterserviceversion.yaml
```diff
@@ -203,6 +203,15 @@ spec:
       - kind: ServiceMonitors
         name: servicemonitors.monitoring.coreos.com
         version: v1
+      - kind: ScrapeConfigs
+        name: scrapeconfigs.monitoring.coreos.com
+        version: v1alpha1
+      - kind: ServiceMonitors
```
csibbitt, line 209: This was missing previously, and I'm not 100% sure the effects. My guess is it would affected garbage collection in the operator-sdk, preventing these servicemonitors from being cleaned up automatically when the servicetelemetry object is deleted.

csibbitt:
I believe the smoketest failures are related to label changes. I'll take a closer look at this locally and see how bad it is (I see a lot of labels in that part of the code)

csibbitt:
I adjusted the labels to exactly match what we were producing before and the smoketests passed locally.

csibbitt:
> Looks good Chris, thanks! Considering we are removing the service monitors, what happen in brownfield deployments? Do the existing STF continue to run without change and the new scrape config component is deployed with new STF deployments? Thanks!

No. The code as-is deploys the new ScrapeConfig objects, and then deletes the old ServiceMonitor objects[1]
If users have manually configured additional servicemonitors via the "servicemonitor_manifest" option[2], then they will be preserved.

Upgrade testing will be required, but I don't forsee any difficulties. The labels of the new metrics exactly match the old ones.

[1] https://github.com/infrawatch/service-telemetry-operator/pull/607/files#diff-6b852f8c4b44841b7ddd16194f6304203cf7c6b134b7a008a0fc4b3c043b4d1cR87
[2] https://github.com/infrawatch/service-telemetry-operator?tab=readme-ov-file#overriding-default-manifests

roles/servicetelemetry/tasks/component_scrapeconfig.yml
```diff
@@ -0,0 +1,93 @@
...
+
+- name: Look up auth secret to get token secret name
+  k8s_info:
+    api_version: v1
+    kind: Secret
+    namespace: '{{ ansible_operator_meta.namespace }}'
+    name: '{{ service_account.resources[0].secrets[0].name }}'
+  register: auth_secret
+
+- name: Create SG-specific Service Monitor manifest
```
vyzigold, line 17:
```suggestion
- name: Create SG-specific Scrape Config manifest
```

vyzigold (changes requested):
Otherwise LGTM

paramite (approved):
Didnt' test this, but it seems lgtm to me.

vkmc (approved):
LGTM

vkmc:
recheck

PR #613, opened 2024-07-17
Update nodesets label

danpawlik:
The CI jobs does not need to be force use one cloud provider in the CI. This commit set the label that is defined in multiple cloud provider in Zuul. Available labels you can find [1].

[1] https://review.rdoproject.org/zuul/labels

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -3,7 +3,7 @@
     name: stf-crc_extracted-ocp412
     nodes:
       - name: controller
-        label: cloud-centos-9-stream-tripleo-vexxhost
+        label: cloud-centos-9-stream-tripleo
       - name: crc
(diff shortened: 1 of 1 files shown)


elfiesmelfie:
@danpawlik We can co-ordinate to merge and test this. 
OR
Duplicate and rename the current nodeset so we have a temperory name on the nodeset.
Then We can test with the new -all-clouds nodes, make sure it works (I expect it will) and then update the existing nodesets to match -all-clouds..

The first one is definitely easier: I can bypass the CI, merge this change and add a test patch to verify that everything still works

elfiesmelfie:
I've merged this and will use this change to test it: https://github.com/infrawatch/service-telemetry-operator/pull/615

If there is some failure, then I can revert this PR

PR #614, opened 2024-07-17
Update nodesets label in stable-1.5

danpawlik:
The CI jobs does not need to be force use one cloud provider in the CI. This commit set the label that is defined in multiple cloud provider in Zuul. Available labels you can find [1].

[1] https://review.rdoproject.org/zuul/labels

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -3,7 +3,7 @@
     name: stf-crc_extracted-ocp412
     nodes:
       - name: controller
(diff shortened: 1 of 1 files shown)


elfiesmelfie:
The same comments apply here as in https://github.com/infrawatch/service-telemetry-operator/pull/613

I've merged this change, and can revert it if the test PR doesn't pass

PR #623, opened 2024-07-24
Create our own token for prometheus-stf SA

csibbitt:
* In OCP 4.16, these are no longer created by default


csibbitt:
recheck

vkmc (approved):
LGTM, thanks for the quick fix Chris!

elfiesmelfie:
There's a "needs-backport" label on this PR, however, the scrapeconfig was added in July in #607, which was not backported to stable-1.5.

It looks to me like this doesn't need a backport.

PR #626, opened 2024-08-22
Use bundle version instead of latest as tag

vkmc:


csibbitt:
"And use external registry instead of internal " - I'm not seeing this in the changes, what am I missing?

Change otherwise makes sense to me.

vkmc:
Sorry about that, I reverted the change and fixed the tag there as well. Now we point to the internal registry with the correct tag. So this works with either deploy from bundles + index and local build + index.

PR #628, opened 2024-08-26
STF 1.5.5 release ops

vkmc:
Depends-On: https://github.com/infrawatch/sg-core/pull/141


csibbitt (commented):
- I think the spurious extra commits that github shows are because the previous release prep branches were squashed before merging to stable-1.5.(For example https://github.com/infrawatch/service-telemetry-operator/commit/ad468f2b1d33186c998beecc65366270e7a4f2f8)
  - The *content* from those commits is already on the branch, but the commits themselves are missing from the history, so github lists them.
  - **DO NOT SQUASH THIS PR** when merging to stable-1.5 and this problem should go away for the future
- Did a sanity-read on all content looking for version numbers going the wrong way, changes I couldn't explain, and reviewing things that didn't make immediate sense. 
  - **Just one comment.**
- Verified that I can see expected changes pertaining to the major epics and bugfixes for 1.5.5
  - IPv6 related (change to qdr port 5671 listen address and migration to scrapeconfigs)
  - 4.16 support (new auth token, changes to zuul, Dockerfile(!) & Operator metadata + other misc changes)
  - RBAC fixes (plural resource names and s/group/resourceAPIGroup/ in SARs)


I would be happy to approve if we sort out the OCP version range in the Dockerfile - either confirm it's intentional, or change it to match what we set in SGO.

vkmc:
I think this was an intentional change at the moment of submitting it, since we had CI for OCP 4.12, OCP 4.14 and OCP 4.16. Now, we dropped OCP 4.12 CI. So it makes sense we also fix this. I will submit a fix for this.

vkmc:
recheck

vkmc:
recheck

elfiesmelfie (approved):

build/Dockerfile
elfiesmelfie, line 1: This might need updating

PR #629, opened 2024-08-26
Drop OCP 4.12 from openshift.versions annotation

vkmc:
OCP 4.12 maintenance support has finished, so we should remove OCP 4.12 from the supported versions


csibbitt (approved):

vkmc:
recheck

PR #630, opened 2024-08-26
Use openshift-ansible-operator 4.14 in STO

vkmc:


csibbitt (approved):
LGTM if we can get the 4.14 CI to pass (looks like a crc problem rather than a this problem)

vkmc:
recheck

vkmc:
recheck

PR #640, opened 2024-11-04
Get tag from the name instead of the metadata

vkmc:
Retrieve image tag from name instead of the image metadata when deploying from index with pre-built bundles.


vkmc:
The proposed change is expected to fail since in the upstream bundles we don't set the tag on the name, instead, we have the image tag on the metadata. We will need to consider that case separately (if Quay, do X).

PR #647, opened 2024-11-27
Remove htpasswd_sha1 filter definition

vkmc:
htpasswd_sha1 is not used anywhere else in STO, but we still carry over the filter definition.

Since SHA1 is deprecated, it raises warnings on our static code analysis tools.


elfiesmelfie (approved):

ayefimov-1 (commented):
looks good.

PR #653, opened 2025-02-13
Update GH actions to use ubuntu-latest

vkmc:
The Ubuntu 20.04 runner image will be fully unsupported by April 1, 2025.

Actions need to be updated to a newer version.

For some jobs we are already using ubuntu-latest, so use the same for the jobs using ubuntu20.04


ayefimov-1 (approved):
Looks okay.

.ansible-lint-ignore
```diff
@@ -0,0 +1,83 @@
+# This file contains ignores rule violations for ansible-lint
```
elfiesmelfie, line 1: I think you can update the ansible-lint config to skip linting the roles/servicetelemetry/ directoty

Same for the build/run-stf-ci/ directory., although I think that the effort to update those is safer.

.github/workflows/main.yml
```diff
@@ -4,22 +4,30 @@ on: push
...
       - name: Checkout code
         uses: actions/checkout@v4.1.3
 
+      # Python version being set because of issue with Ubuntu permissions and versions of Python < 3.11 
+      - name: Set up Python
+        # This is the version of the action for setting up Python, not the Python version.
+        uses: actions/setup-python@v5
+        with:
+          # Set exact version of a Python version or can use '3.x'.
+          python-version: '3.12.8'
```
elfiesmelfie, line 18: oooh, very nice

.github/workflows/main.yml
```diff
@@ -65,7 +80,7 @@ jobs:
 
   build-operator-check:
     name: Build Operator check
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
elfiesmelfie, line 83: Thanks for updating this!

elfiesmelfie (approved):

.ansible-lint-ignore
vkmc: Indeed, we can work on fixing the linting issues or configuring ansible-lint in a follow up patch.

PR #654, opened 2025-02-19
Update COO channel to stable

vkmc:
COO 1.0.0 now publishes their artifacts in the stable channel

The development channel that was in use is no longer available

Update stf-run-ci to reflect this

### DIFF
--- a/build/stf-run-ci/tasks/setup_base.yml
+++ b/build/stf-run-ci/tasks/setup_base.yml
@@ -42,7 +42,7 @@
         name: cluster-observability-operator
         namespace: openshift-operators
       spec:
-        channel: development
+        channel: stable
         installPlanApproval: Automatic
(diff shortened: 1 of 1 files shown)


elfiesmelfie (approved):

elfiesmelfie:
I'm willing to ignore the 4.14 failures since it'll probably take a long time to fix them, and technically, this is testing STF 1.5.6rc, which isn't supported on OCP 4.14.

 @vkmc Once there is a second approval, and the 4.16 jobs are still passing, I am okay to bypass the merging rules

vkmc:
The failures in OCP 4.14 jobs are not related to this change.

The work required to fix OCP 4.14 jobs is in progress but, as correctly assessed by @elfiesmelfie, it will take some more time.

I manually tested this change in a OCP 4.14 environment with a positive result.

PR #657, opened 2025-03-07
[ci] Make the OCP 4.14 jobs non-voting

elfiesmelfie:


vkmc (approved):
LGTM, thanks Emma!

vkmc:
Not a blocker, but, any reason to make them non-voting instead of removing them?

PR #659, opened 2025-03-14
Use openshift-ansible-operator 4.18 in STO

vkmc:


compi-migui:
Looks like the 4.18 operator moved to UBI 9, so the EPEL 8 repository we're using for our deps needs to be replaced with something else:
https://github.com/openshift/ansible-operator-plugins/blob/release-4.18/images/ansible-operator/Dockerfile#L5
https://github.com/infrawatch/service-telemetry-operator/blob/master/build/Dockerfile#L5-L7

compi-migui (approved):

vkmc:
recheck

PR #666, opened 2025-05-02
Use operator-sdk v1.39.2

compi-migui:
Using a newer operator-sdk will make dependency management a lot easier for downstream hermetic build systems.

PR #667, opened 2025-05-16
Use operator-sdk v1.39.2 (#666)

vkmc:
* Use operator-sdk v1.39.2

Using a newer operator-sdk will make dependency management a lot easier for downstream hermetic build systems, like the ones used downstream.

It also lets us do away with using different versions for bundle generation and deployment in our CI.

* Close stdin on workflow operator-sdk generate bundle calls

An open stdin makes operator-sdk assume it's supposed to take input from stdin, when we actually want it to look at the working directory. There is no way to avoid this using command-line flags, so we need to explicitly close the stdin pipe in all calls to it.

(cherry picked from commit 5c5c5637c3811394d9c085a422d2dac237239de2)


elfiesmelfie (approved):

elfiesmelfie (approved):

vkmc:
recheck

PR #668, opened 2025-06-02
Log in to the registry before fetching bundles

vkmc:


vyzigold:
heh, interesting error `'dict object' has no attribute 'stdout'.`, even though it shows "stdout" in the debug output

vkmc:
> heh, interesting error `'dict object' has no attribute 'stdout'.`, even though it shows "stdout" in the debug output

Mysterious indeed :)

PR #670, opened 2025-06-04
[stable-1.5] Bundle changes for Konflux support (#669)

compi-migui:
Backport of https://github.com/infrawatch/service-telemetry-operator/pull/669

* Use full pullspecs for related images

Renovate, used by Konflux for updating dependency references, works better with full container pullspecs rather than separate repository URLs and tags.

This lets us inject pullspecs pinned by digest, which don't quite follow the repo:tag syntax.

* Don't set update graph metadata in CSV

This was originally added for the benefit of downstream builds, but those are moving to file-based catalog fragments which don't need update graph metadata to be embedded in the bundle.

* Refactor pullspec building to make it more readable

(cherry picked from commit ce33e1dcd1733d303872f1a71f40d01b983f9477)


vkmc (approved):
LGTM, thanks Miguel

PR #671, opened 2025-06-05
Use /tmp instead of base_dir for fetching STO and SGO CSV

vkmc:
Use /tmp as the dest for the fetch STO and SGO step instead of the base_dir

By default, we set base_dir to /opt/ci and this is normally a non writable directory in the Ansible control node. Because of this, running stf-run-ci with the purpose of creating a development environment fails.

Considering we execute the fetch command in order to be able to do an include_vars afterwards, it is safe to use a /tmp location.


elfiesmelfie (approved):
Looks good, and the CI jobs look like they're going to pass (5/6 already done as I'm writing)

PR #679, opened 2025-08-05
bundle gen: allow passing in pre-generated bundle version (#678)

compi-migui:
This improves our support for generating bundles as part of multistage container builds, as a single pre-generated bundle version can be fed into the generated bundle and into the bundle container labels.

RELDEL-7554

(cherry picked from commit e882edbd3cb0eaa71760fa2cc23af3f4ca9f8e00)


vkmc:
The failures in OCP 4.16 jobs are not related to this change

PR #682, opened 2025-08-26
Add support for deploy from catalog

vkmc:
Add __deploy_from_catalog and __stf_catalog_index_image_path options to stf-run-ci.

This new option allows to create a CatalogSource using a pre-built index image and deploy STF from it.

This is particularly useful to test package metadata issues that would be invisible when testing bundles but very apparent when installing from the actual catalog.

Closes: OSPRH-19028


vkmc:
recheck

elfiesmelfie (commented):
Looks good.

Besides suggesting that the _deploy_from_index could additionally be automatically set when the stf_index_imasge var is set, I have no suggestions.  (this is optional, but would also change the established pattern of needing both the enable_* and the image URl)

elfiesmelfie (approved):
The previous review was supposed to be an approve, but I didn't click the radio button

vkmc:
Totally agree... I stick with having a flag and a configuration option so we keep consistency with other deployment options (e.g. deploy from bundles uses both the flag and the bundles info)

Thanks for the review!

PR #683, opened 2025-08-26
Update olm.maxOpenShiftVersion to 4.18

vkmc:
olm.maxOpenShiftVersion in STO should be 4.18 instead of 4.16

This wrongly set config is preventing STF users to easily perform upgrades to the latest OCP supported version

Closes-Bug: OSPRH-18670


vkmc:
Performed a local build of the artifacts and issued an "oc adm upgrade". Verified that this change fixes the issue.

PR #686, opened 2025-09-22
Fix stf-run-ci mutually exclusive deployment flags condition

vkmc:
Fail when two or more of the non compatible deployment options are set. To make the check simpler, convert them to integers (where true becomes 1 and false becomes 0) and check if their sum is greater than 1.

Closes-Bug: OSPRH-20165


vkmc:
Errors on the Linting CI are not related to this change

Seems there was a regression on the environment, since no other change introduced an update on ansible-core or ansible-lint versioning.

vkmc:
recheck

ayefimov-1 (approved):
Looks good.

PR #688, opened 2025-09-25
Update ansible-lint pinned version

vkmc:
The version of ansible-lint is not compatible with the version of ansible-core we get in ubuntu-latest.

Bump the version of ansible-lint to the next available version that fixes the issue. 

https://github.com/ansible/ansible-lint/releases/tag/v25.2.0 fixes the issue.

Partially Closes: OSPRH-20316

PR #691, opened 2025-09-29
Update ansible-lint pinned version (#688)

vkmc:
* Update ansible-lint pinned version

    Bump ansible-lint to 25.2.0

* Update ansible-lint-ignore with new linting issues

ansible-lint 25.2.0 introduces new linting rules that we are not following. Add them to the ignore list.

* Pin to ansible 12.0.0

(cherry picked from commit 9be283230c8041ea41367857f7d49a2bafa19bef)


vkmc:
We have different linting issues in master/stable-1.5. The ansible-lint-ignore file is different between branches. The option we have is to update ansible-lint-ignore here with the corresponding requirements for stable-1.5 and then use fix the differences when a new upstream release is made.

ayefimov-1 (approved):
Eveything looks good to me.

PR #697, opened 2026-01-21
STF 1.5.7 release ops

vkmc:


build/stf-run-ci/README.md
```diff
@@ -155,6 +157,19 @@ Since you will fetch the selected images from a bundle registry, it is required
 access credentials for the desired registry correctly configured. Check the "Deployment with pre-build bundles"
 docs above to get more information about this.
 
+## Deployment with pre-built index image
+
+Is it also possible to pass the path for a pre-build index image to create a CatalogSource and deploy STF from it.
```
vyzigold: This sentence is strange. It should probably be "It is also ..." (different word order). But it isn't anything to prevent a release.

vkmc: Yeah... you are right. Sounds like a question more like a fact. Considering it is a README for the CI,, I think we can accept it as is :)

PR #699, opened 2026-02-27
Update OCP 4.20 job to use nodeset 4-20-14

vkmc:
This change https://github.com/openstack-k8s-operators/ci-framework/pull/3724/changes updates the nodeset names and labels from 4-20-1 to 4-20-14.

Update STF OCP 4.20 jobs to use the updated nodeset name

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -13,8 +13,7 @@
       - name: controller
         label: cloud-centos-9-stream-tripleo-vexxhost
       - name: crc
-        label: crc-cloud-ocp-4-20-14-3x
-
+        label: crc-cloud-ocp-4-20-14-3xl
 - job:
     name: stf-base-2node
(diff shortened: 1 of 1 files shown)


vkmc:
Failure is expected, we need to cherry pick this change stable-1.5 to once it merges.

.zuul.yaml
```diff
@@ -13,7 +13,7 @@
       - name: controller
         label: cloud-centos-9-stream-tripleo-vexxhost
       - name: crc
-        label: crc-cloud-ocp-4-20-1-3xl
+        label: crc-cloud-ocp-4-20-14-3x
```
elfiesmelfie: Did the name really change from using `3xl` to `3x` as the suffix?

vkmc: Good catch, fixing it

PR #703, opened 2026-04-22
Use role instead of playbook - 01-bootstrap.yml and 02-infra.yml

danpawlik:
Playbook usage was deprecated for a while, so now after the playbooks have been removed from CI Framework project [1], the CI raise an error.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/3882

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/3193

### DIFF
--- a/ci/2node.yml
+++ b/ci/2node.yml
@@ -1,0 +1,50 @@
+---
+- name: "Do pre-work to get kubeconfig"
+  hosts: controller
+  gather_facts: true
+  vars:
+    ci_framework_dir: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/openstack-k8s-operators/ci-framework'].src_dir }}"
+    cifmw_use_opn: false
+    cifmw_use_devscripts: false
+    cifmw_openshift_setup_skip_internal_registry_tls_verify: true
+    cifmw_target_host: controller
+  environment:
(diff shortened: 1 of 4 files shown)


danpawlik:
recheck

vkmc:
recheck

danpawlik:
recheck

danpawlik:
recheck

danpawlik:
recheck

ci/2node.yml
elfiesmelfie, line 1: Why is this being moved from pre-2node? What was the error that you saw that indicated this move was needed?

elfiesmelfie (commented):
Looks okay. I'm curious as to why the 2node playbook is introduced.

vkmc (approved):
LGTM, thanks Daniel for helping us get our CI unblocked!

danpawlik:
@elfiesmelfie because in pre-2node we are doing symlink to cifmw modules. The 2node is a new playbook, where Zuul executor runs separate ansible-playbook command. It means, that the cifmw modules would be available. 
It is more workaround, because normally we should install ci-framework collection via Zuul CI definition, but it require newer Zuul. Here we introduce a trick, where we know that each Zuul executor uses booblewrap with own HOME dir, then we are doing a symlink to Ansible dir.

elfiesmelfie (approved):

vkmc:
recheck

PR #712, opened 2026-05-21
Wait a while for OpenShift API server to apply last changes

danpawlik:
In some corner cases, when infrastructure might be overloaded, the OpenShift API is not restarted immediately, but it is done after a while. From basic check on failing job, it happens from up to 90 seconds. Making condition that would watch on OpenShift API server that has been finally changed state to to terminating then back to active might be a good alternative here, but can also impact future job if such restart is not needed.

Also remove duplicated line from periodic job in zuul config.

(cherry picked from commit c65611428baae2496819a4ccd4e11a3f157db752)

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -242,5 +242,3 @@
         - stf-crc-ocp_418-local_build-index_deploy
         - stf-crc-ocp_420-local_build-index_deploy
         - stf-crc-ocp_418-nightly_bundles-index_deploy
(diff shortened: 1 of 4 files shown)


vyzigold (approved):

vkmc:
Seems we need a backport of https://github.com/infrawatch/service-telemetry-operator/pull/713

danpawlik:
done @vkmc https://github.com/infrawatch/service-telemetry-operator/pull/714

vkmc:
recheck

PR #717, opened 2026-07-16
Update pyasn1 to newer version

danpawlik:
The pyasn1 which is installed in base container image has installed vulnerable version of pyasn1 library.
Update the library to latest version [1]

Related-To: OSPRH-32895

[1] https://github.com/pyasn1/pyasn1/releases/tag/v0.6.4

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -12,6 +12,12 @@ RUN bash -c -- 'if [ "${NO_PROXY:-__ZZZZZ}" == "__ZZZZZ" ]; then echo "Applying 
 # -- cleans up the cached data from dnf to keep the image as small as possible
 RUN dnf update -y --exclude=ansible* && dnf install -y python3-passlib python3-bcrypt && dnf clean all && rm -rf /var/cache/dnf
 
+# NOTE(dpawlik): The base image contains vulnerable pyasn1 library installed,
+# that's why it needs to be updated to newer.
+# upgrade pyasn1 to fix CVE-2026-59885, CVE-2026-59884, CVE-2026-59886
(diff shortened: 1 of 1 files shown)


lhh:
This may not work outside of upstream builds or in places where building from source is a requirement. Also, would it make sense to nuke the old one too?

Something like: 
`RUN rpm -e --nodeps python3-pyasn1`
`RUN pip3 -vvv install --compile --no-binary :all: pyasn1>=0.6.4`

danpawlik:
recheck

danpawlik:
@lhh is it fine ?

elfiesmelfie (approved):

lhh:
It's fine; turns out the source image didn't have it, but this keeps the BOM clear if it ever did have it:
`
2026-07-20 14:20:01.375524 | controller | error: package python3-pyasn1 is not installed
`

Same caveats we discussed apply (fine for these upstream builds, not fine in any place things need to be built from source). I'd rather it wasn't needed; hopefully the image you are sourcing here publishes a proper update.

@compi-migui ?

danpawlik:
> It's fine; turns out the source image didn't have it, but this keeps the BOM clear if it ever did have it: `2026-07-20 14:20:01.375524 | controller | error: package python3-pyasn1 is not installed`
> 
> Same caveats we discussed apply (fine for these upstream builds, not fine in any place things need to be built from source). I'd rather it wasn't needed; hopefully the image you are sourcing here publishes a proper update.
> 
> @compi-migui ?

directly only https://github.com/infrawatch/prometheus-webhook-snmp uses that library. Here is just "fixing" base image installation, which would not be updated for this current version.

PR #719, opened 2026-08-06
Mark ServiceTelemetry CR spec values as unsafe for Ansible templating

danpawlik:
Set markUnsafe: true in watches.yaml so that CR spec fields are treated as literal strings during Ansible reconciliation, matching the recommended ansible-operator configuration for custom resources.

Related-To: OSPRH-33070


vkmc (approved):
LGTM

PR #721, opened 2026-08-10
Harden SNMP webhook and Prometheus workload security

danpawlik:
Add pod/container securityContext, ConfigMap-based resource limits, and health probes to the SNMP webhook Deployment template. Remove the empty securityContext: {} override from the Prometheus CR so the Prometheus operator's own defaults apply.

The resource limits follow the same ConfigMap pattern used in smart-gateway-operator (PR #217): defaults ship in defaults/main.yml, a ConfigMap is created on first reconcile, and users can tune values without rebuilding the operator or modifying the CRD.

Related-To: OSPRH-33076

Assisted-By: Claude Opus 4.6 <noreply@anthropic.com>


danpawlik:
recheck

vkmc (approved):
LGTM

PR #722, opened 2026-08-12
Use operator-controlled image for Grafana instead of CR spec field

danpawlik:
Remove the baseImage field from the ServiceTelemetry CRD spec and move
Grafana container image selection to the operator defaults. This ensures
the Grafana image is controlled by the operator rather than user-supplied
CR values.

- Remove grafana.baseImage from CRD schema and CR example
- Add RELATED_IMAGE_GRAFANA to bundle generation for image pinning
- Regenerate OLM catalog CRD to match deploy/crds/ source of truth
- Remove baseImage from CSV alm-examples annotation

Related-To: OSPRH-33072

Assisted-By: Claude Opus 4.6 <noreply@anthropic.com>


danpawlik:
recheck

danpawlik:
recheck

vkmc (changes requested):
This one is tricky. The code looks good, but I wouldn't merge this. Here is why: Grafana is not part of the supported stack for STF. STF users bring their own Grafana to use with STF. Yes, it adds a potential attack vector. But no, we cannot include the Grafana image as part of the STF bundle. More info about this here https://infrawatch.github.io/documentation/#setting-up-grafana-to-host-the-dashboard_assembly-advanced-features

danpawlik:
@vkmc so here we need to choose:
- update what is in the release to fix potential issue
- drop grafana, which would not be correct - it can be done between main release
- do nothing and close bug

Which would be the best?

vkmc:
I think the first two options are not possible. If we add the Grafana image to the bundle, then we need to support Grafana. That won't happen since it is a community based operator. If we drop Grafana, we leave our users without a dashboard.

The STF admin is also an OCP admin, so it is a trusted user. If a non authorized user can modify STF config, then they already have access to the OCP cluster.

Let's close this issue for now and reassess if this is considered an actual concern.

PR #723, opened 2026-08-13
Add SHA256 checksum verification for operator-sdk and Go downloads

danpawlik:
CI previously downloaded the operator-sdk binary from a GitHub release URL via plain curl and executed it without verifying integrity, letting a compromised or substituted release asset tamper with the generated bundle. Add a sha256sum check right after each download, using the checksum published in operator-sdk checksums.txt.

Related-To: OSPRH-33078


build/get_go.sh
```diff
@@ -9,8 +9,11 @@ BINARY_NAME="go"${VERSION}
 GO_DL_URL=https://golang.org/dl/${ARCHIVE_NAME}
 
 
+GO_SHA256="${2:-013a489ebb3e24ef3d915abe5b94c3286c070dfe0818d5bca8108f1d6e8440d2}"
```
vkmc: This is dead code, we should avoid updating it.

vkmc:
You need to pass the hash as a second param in https://github.com/infrawatch/service-telemetry-operator/blob/master/build/stf-run-ci/tasks/main.yml#L96-L101

The first param is picking up the value from the defaults config in https://github.com/infrawatch/service-telemetry-operator/blob/c2ab8595fec5b8f7d454c4ac42488eeee55ea775/build/stf-run-ci/defaults/main.yml#L54

Maybe we can do something similar instead of setting the env vars in the jobs

vkmc (changes requested):

build/get_go.sh
danpawlik: agree.
Just FTR: we were not aware that the file is used or not when the PR was done.
It seems it is not used anymore, so no additional changes should be done there.

danpawlik:
> You need to pass the hash as a second param in https://github.com/infrawatch/service-telemetry-operator/blob/master/build/stf-run-ci/tasks/main.yml#L96-L101
> 
> The first param is picking up the value from the defaults config in
> 
> https://github.com/infrawatch/service-telemetry-operator/blob/c2ab8595fec5b8f7d454c4ac42488eeee55ea775/build/stf-run-ci/defaults/main.yml#L54
> 
> Maybe we can do something similar instead of setting the env vars in the jobs

Sure I can, but is there any sense if the default value already provides the correct value?

danpawlik:
@vkmc updated
From the other side, we should replace build/get_operator_sdk.sh and use just ansible - https://github.com/openstack-k8s-operators/osp-director-dev-tools/blob/376370441da2b4e250e9a17ae6091bee261fff53/ansible/roles/download_tools/tasks/main.yaml#L50

sdatko (approved):
/lgtm /approve Looks good to me

vkmc (approved):
LGTM, thanks!

PR #729, opened 2026-09-08
Set alertmanager and prometheus spec.version

danpawlik:
Both CRs set `spec.image` but not `spec.version`, so prometheus-operator fell back to
its own defaults (Alertmanager v0.33.1, Prometheus v3.13.1) while the shipped images
are Alertmanager v0.26.0 and Prometheus v2.48.1.

Two breakages after COO v1.5.0 picked up prometheus-operator v0.93.x:

    - Alertmanager gets `--cluster.peer-name` (added for >= v0.30.0), doesn't recognise it,
      and CrashLoopBackOffs.
    - Prometheus loses `--storage.tsdb.retention.time` (only passed below v3.11.0 now), so
      the configured retention is silently ignored and it falls back to the 15d default.

Declaring the real versions fixes both. STF sets none of the other version-gated
fields, so nothing else changes.

The upstream default images are pinned to the same versions, so the declared version
stays true on the dev and CI path where `:latest` was used.

Also:
Pass OPERATOR_IMAGE_PULLSPEC when generating bundles locally
generate_bundle.sh pairs each image variable with its own _TAG, as in
RELATED_IMAGE_PROMETHEUS plus RELATED_IMAGE_PROMETHEUS_TAG, so the operator image
needs OPERATOR_IMAGE_TAG, while we pass OPERATOR_TAG. The pair never matched, so
the CSV kept the quay.io/infrawatch defaults and CI deployed the published operator
image instead of the one built from the change under test.

Related-to: OSPRH-36575


danpawlik:
recheck

danpawlik:
```
[rhos-ci@test~]$ oc get pods
NAME                                                              READY   STATUS      RESTARTS   AGE
94344b152e7752a60291fafb8a32610854303c171a45d32ed06e950737srsgw   0/1     Completed   0          11m
alertmanager-default-0                                            3/3     Running     0          2m19s
default-cloud1-ceil-meter-smartgateway-6bc9fc7f7b-8bgmg           3/3     Running     0          80s
default-cloud1-coll-meter-smartgateway-75647fbb5-smbj5            3/3     Running     0          100s
default-cloud1-sens-meter-smartgateway-dc55cf7bb-z45pl            3/3     Running     0          58s
default-interconnect-68864d46cb-j4frm                             1/1     Running     0          3m38s
elasticsearch-es-default-0                                        1/1     Running     0          79m
interconnect-operator-5bb49f789d-tncxc                            1/1     Running     0          11m
prometheus-default-0                                              3/3     Running     0          3m3s
service-telemetry-operator-84f85d77b4-wtx7c                       1/1     Running     0          11m
smart-gateway-operator-65f8f74885-2rdws                           1/1     Running     0          11m
uay-io-dpawlik-service-telemetry-operator-bundle-1-5-1789129770   1/1     Running     0          11m
```

and
```
oc get alertmanager.monitoring.rhobs default -n service-telemetry -o yaml | grep '^  version:'
  version: v0.26.0

oc get prometheus.monitoring.rhobs default -n service-telemetry -o yaml | grep '^  version:'
  version: v2.48.1
```

danpawlik:
recheck

danpawlik:
recheck

danpawlik:
recheck

danpawlik:
recheck

danpawlik:
recheck

danpawlik:
recheck

vyzigold (approved):

danpawlik:
We will go with another way. Upstream will stay as it is, where bundle will have another version set, so it would not pick "latest" image version but the version available in bundle (because previous image version defined in bundle disappeared)

